PR #3, opened 2021-12-08
Yaml2nxdl

aalbino2:


nexusparser/tools/yaml2nxdl/yaml2nxdl.py
```diff
@@ -1,88 +1,114 @@
...
-            del self.yml['doc']
+
+def pretty_print_xml_given_root(root, output_xml):
+    """
+    Print formatted xml file with built-in libraries
+    """
+    xml_string = xml.dom.minidom.parseString(ET.tostring(root)).toprettyxml()
+    xml_string = os.linesep.join([s for s in xml_string.splitlines() if s.strip()])
+    with open(output_xml, "w") as file_out:
+        file_out.write(xml_string)
```
markus1978, line 42: Too long names are better than too short one. But just for the "feeling", I would have gone for `pretty_print_xml`. If you later use good names for the root, you could end up with something like `pretty_print_xml(xml_root)` or something.

nexusparser/tools/yaml2nxdl/yaml2nxdl.py
```diff
@@ -1,88 +1,114 @@
...
+    help='The path to the input data file to read. (Repeat for more than one file.)'
+)
+def yaml2nxdl(input_file: str):
+
+    # step1
+    yml_appdef = read_application_definition(input_file)
+
+    # step2a
+    attr_qname = ET.QName("http://www.w3.org/2001/XMLSchema-instance", "schemaLocation")
+    rt = ET.Element(
```
markus1978, line 57: Here, I would use `xml_root` or at least `root`.

nexusparser/tools/yaml2nxdl/yaml2nxdl.py
```diff
@@ -1,88 +1,114 @@
...
+        'definition', {attr_qname: 'http://definition.nexusformat.org/nxdl/nxdl.xsd'},
+        nsmap={None: 'http://definition.nexusformat.org/nxdl/3.1',
+               'xsi': 'http://www.w3.org/2001/XMLSchema-instance'}
+    )
+    # step2b
+    assert 'name' in yml_appdef, 'keyword not specified'
+
+    # pi = ET.ProcessingInstruction(
+    #    "xml-stylesheet", text='type="text/xsl" href="nxdlformat.xsl"')
+    # rt.addprevious(pi)
```
markus1978, line 67: I am not a 100%. I assume that this are XML transformation instructions that allows to print the XML in a certain format or style-sheet. I doubt that the bare-bones Python ET supports this.

aalbino2: I just left here this part to recall that yes, how you say, ET is not supporting that function and I have to find soon a proper replacement

rettigl:
Not sure this this is the right place for this comment.
I understood that if no name is given for a Nexus object, then either the name should be given as the class name in upper case, without the "NX", or simply be omitted. Right now, the name is just set to the class name:

```
            if kName != '':  # use the custom name for the group
                grp.set('name', kName)
            else:
                if kType != '':
                    grp.set('name', kType)
```

An example in the official nexus doc can e.g. be found in NXbeam:
https://manual.nexusformat.org/classes/base_classes/NXbeam.html#nxbeam
Here, DATA is an unnamed element. The corresponding nxdl file
https://github.com/nexusformat/definitions/blob/main/base_classes/NXbeam.nxdl.xml
does not specify a "name" tag for this field.

rettigl:
This part also seems to be obsolete, no?

```
            else:
                if keyword == 'type':
                    print(keyword + ' facing a type where we would not expect it!')
```
The types are given now by the bracket format. This generates an error, if you want to generate a field called "type" (which we do in one of our definition).

nexusparser/tools/yaml2nxdl/yaml2nxdl.py
```diff
@@ -1,88 +1,114 @@
...
+    # step1
+    yml_appdef = read_application_definition(input_file)
+
+    # step2a
+    attr_qname = ET.QName("http://www.w3.org/2001/XMLSchema-instance", "schemaLocation")
+    rt = ET.Element(
+        'definition', {attr_qname: 'http://definition.nexusformat.org/nxdl/nxdl.xsd'},
+        nsmap={None: 'http://definition.nexusformat.org/nxdl/3.1',
+               'xsi': 'http://www.w3.org/2001/XMLSchema-instance'}
+    )
```
rettigl, line 61: The XML header part "nsmap" in this format seems to not be nexus-conform, at least the punx-tool "tree" function does not work with this header. The nexus nxdl files on github seem to have as header:
xmlns="http://definition.nexusformat.org/nxdl/3.1"
using this, it is recognized by punx

nexusparser/tools/yaml2nxdl/yaml2nxdl_recursive_build.py
```diff
@@ -1,226 +1,246 @@
...
+            xml_handle_link(obj, keyword, value)
+        elif (kType in nx_clss) or (kType not in nx_type_keys + ['']):
+            print('nx_class')
+            # we can be sure we need to instantiate a new group
+            grp = ET.SubElement(obj, 'group')
+            if kName != '':  # use the custom name for the group
+                grp.set('name', kName)
+            else:
+                if kType != '':
+                    grp.set('name', kType)
```
rettigl, line 147: I think this else block should be removed, and no name defined if the object is not named (see my other comment)

nexusparser/tools/yaml2nxdl/yaml2nxdl.py
aalbino2: Thank you for pointing out this discrepancy to nexus format. We indeed are moving from lxml to xml library and some functions are lacking, like the one generating the correct namespace. This issue has been now resolved and, once this pull request is closed, you will find the correct header in the xml generated file

nexusparser/tools/yaml2nxdl/yaml2nxdl_recursive_build.py
```diff
@@ -1,226 +1,233 @@
...
+        elif keyword[-6:] == '(link)': 
+            xml_handle_link(obj, keyword, value)
+        elif (kType in nx_clss) or (kType not in nx_type_keys + ['']):
+            # we can be sure we need to instantiate a new group
+            grp = ET.SubElement(obj, 'group')
+            if kName != '':  # use the custom name for the group
+                grp.set('name', kName)
+            else:
+                if kType != '':
+                    grp.set('name', kType)
```
aalbino2, line 143: @mkuehbach Let's evaluate together the point raised by @rettigl

sanbrock:
- please fix the compliance with PEP8  
- please add an (empty) __init__.py file to each directory needed to be installed by pip

sanbrock:
I see pycodestye goes through - great! Please correct the code according to pylint suggestions, as well.  
Note that in vscode, you can use F2 to rename symbols, so you do not have to manually search for all their appearance and do the name modification everywhere.  
On the other hand, be careful and run the test-cases regularly inside vscode.

sanbrock:
@aalbino2 could you please try to fix the pylint issues in  
- [ ] nexusparser/parser.py,  
- [ ] nexusparser/tools/nexus.py,  
- [ ] nexusparser/tools/yaml2nxdl/nxdl2yaml.py,  
- [ ] nexusparser/tools/yaml2nxdl/yaml2nxdl_recursive_build.py,  
- [ ] nexusparser/metainfo/__init__.py  
- [ ] nexusparser/metainfo/nexus.py  
- [ ] tests/test_nexus.py  
- [ ] tests/test_parser.py  
The following files can actually be removed because we do not use the templates anymore:  
- [ ] nexusparser/metainfo/test_nexus.py  
- [ ] nexusparser/metainfo/generate_nexus.py
- [ ] nexusparser/metainfo/generate.py
- [ ] nexusparser/metainfo/templates

sanbrock:
LGTM

PR #8, opened 2022-02-09
Dc validation revamp

shabihsherjeel:
- [x] /@default is not defined in the NXDL files. This should be defined in the top most element for all NXDLs. Maybe in NXobject. I won't recommend changing


sanbrock:
LGTM

PR #12, opened 2022-03-28
Dataconverter virtual datasets

aalbino2:


sanbrock:
Ref_NXellipsometry.nxdl.xml seems to be missing and so test_xml_parsing fails

PR #14, opened 2022-05-30
Added json_map with example files

shabihsherjeel:
This reader allows you to map data in a .json or an xarray as a .pickle to be mapped to any given NXDL using a .mapping.json


sanbrock:
LGTM

PR #20, opened 2022-06-21
Removed pyaml as a requirement and added PyYAML

sherjeelshabih:


aalbino2:
I checkedout the branch, everything seems to be fine, no test is failing, so I would go for the merge

aalbino2 (approved):
I checkedout the branch, everything seems to be fine, no test is failing, so I would go for the merge

PR #23, opened 2022-06-27
Dataconverter param file

shabihsherjeel:
This adds support to recieve a parameter file for the data converter and also allows us to trigger the converter in the Nomad GUI by uploading such a parameter YAML file.

### DIFF
--- a/nexusparser/parser.py
+++ b/nexusparser/parser.py
@@ -19,9 +19,12 @@
 # limitations under the License.
 #
 
+from typing import Iterable, Union
+import os
+import pathlib
 import numpy as np
 from nomad.datamodel import EntryArchive
-from nomad.parsing import MatchingParser
+from nomad.parsing import Parser
 # from . import metainfo  # pylint: disable=unused-import
 from nexusparser.tools import nexus as read_nexus
 from nexusparser.metainfo import nexus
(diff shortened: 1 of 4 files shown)


nexusparser/parser.py
```diff
@@ -237,6 +249,29 @@ def parse(self, mainfile: str, archive: EntryArchive, logger=None, child_archive
...
+                    conv_params["input_file"] = [f"{base_dir}{os.sep}{check_path(file)}"
+                                                 for file in conv_params["input_file"]]
+                else:
+                    conv_params["input_file"] = (f"{base_dir}{os.sep}"
+                                                 f"{check_path(conv_params['input_file'])}")
+                conv_params["output"] = f"{base_dir}{os.sep}{check_path(conv_params['output'])}"
+                convert(**conv_params)
+                mainfile = conv_params["output"]
+
         nexus_helper = read_nexus.HandleNexus(logger, [mainfile])
```
sanbrock, line 275: Won't NOMAD try to parse the freshly generated new NeXus file anyway?

shabihsherjeel: The Nomad walk over the files doesn't win the race here. The user can press reprocess to have another entry for the NXS file. It's hard to enforce this and can lead to unexpected behavior.

I've went with this additional feature in h5web in Nomad that allows us to show the Nexus file assosciated with the dataconverter parameter file. Here is the relevant MR: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/734

PR #31, opened 2022-08-16
Fixed an issue where the converter was trying to convert link dict ob…

shabihsherjeel:
This adds a test and a fix for the issue where the converter was trying to convert links.

Thanks @domna for finding the bug.


domna (approved):
Works for my testing files and fields 👍️ Thanks for the fix!

PR #32, opened 2022-08-25
Adapt nexus to metainfo changes

TLCFEM:


markus1978:
The current branch does not yet contain the *astroid* version fix that @domna  mentioned. On the main nomad we have fixed the versions in the requirements.txt to `astroid==2.5.1` with the same pylint version that is used here.

PR #34, opened 2022-08-31
the line number of the yaml file is displayed in the error message

aalbino2:


aalbino2:
I also added the support to "extends" in the yaml file, as aske by @domna 

Note that I also changed the file used for testing:
/nomad/dependencies/parsers/nexus/tests/data/tools/dataconverter/readers/mpes/Ref_nexus_mpes.log

Because some field was changed in the nxdl from recommended or required to optional.

domna (approved):
The extends keyword works like a charm. Did not review anything else.

The Ref log output for mpes should be the same as in the main branch now, because I updated it only yesterday.

sanbrock (changes requested):
it would be great to have some test cases where the new feature is checked

nexusparser/tools/yaml2nxdl/yaml2nxdl.py
```diff
@@ -108,7 +114,10 @@ def yaml2nxdl(input_file: str, verbose: bool):
         del yml_appdef['category']
```
shabihsherjeel, line 114: This line and line no: 111 can be removed out of the two if statements. It is always called  anyways.

nexusparser/tools/yaml2nxdl/yaml2nxdl_forward_tools.py
```diff
@@ -387,38 +480,41 @@ def recursive_build(obj, dct, verbose):
...
-        elif keyword_name != '':
-            fld = not_empty_keyword_name_handle(obj, keyword_type, keyword_name)
-            second_nested_level_handle(verbose, fld, value)
-            second_level_attributes_handle(fld, keyword, value)
+        elif keyword_name != '' and '__line__' not in keyword_name:
+            fld = not_empty_keyword_name_handle(
+                obj, keyword_type, keyword_name)
+            second_nested_level_handle(verbose, value, fld)
+            second_level_attributes_handle(dct, fld, keyword, value)
         else:
```
shabihsherjeel, line 519: This else could be removed.

shabihsherjeel (changes requested):

shabihsherjeel (approved):

sanbrock (approved):

PR #37, opened 2022-09-06
Hall reader

domna:


sanbrock (approved):

domna:
@aalbino2 Is it working now with the Nexus ELN? If this is the case I would merge.

aalbino2 (approved):
Yes, I can now choose the NXroot schema!

PR #38, opened 2022-09-06
Json map plotting

shabihsherjeel:


aalbino2 (approved):
I think the reader looks nice

Issue #28, opened 2022-07-28
Required setting of non-primitive typed fields or groups are ignored


domna:
If there is a field entry of a non-primitive type it is not required by the reader. For example the field `preparation_description` of type NXnote in NXmpes is set to required, but the a nxs can be created w/o writing it. This also applies for groups, which should exist even if they're empty as long as they are required.

In the source it is stated that only fields and attributes are added to the template:
https://github.com/nomad-coe/nomad-parser-nexus/blob/5898de7f8ef2f5ba96c873d7d945e687bc2611ec/nexusparser/tools/dataconverter/helpers.py#L56-L57

PR #39, opened 2022-09-07
Fix to handle inheriting only groups in an AppDef

shabihsherjeel:
Related to #28


nexusparser/tools/dataconverter/helpers.py
```diff
@@ -67,6 +75,10 @@ def generate_template_from_nxdl(root, template, path=None, nxdl_root=None):
         if tag == "field" \
            and ("units" in root.attrib.keys() and root.attrib["units"] != "NX_UNITLESS"):
             template[optionality][f"{path}/@units"] = None
+    elif tag == "group":
+        if is_a_lone_group(root):
+            template[get_required_string(root)][path] = None
+            template["lone_groups"] = path
```
domna, line 81: I think this needs an array here to also reflect multiple empty and required groups. See failing test.

domna (changes requested):

domna (approved):

PR #44, opened 2022-09-14
Mpes reader fix

Arora0:


nexusparser/tools/dataconverter/readers/mpes/reader.py
```diff
@@ -307,18 +307,18 @@ def read(
...
-                            x_array_loaded.attrs,
-                            value,
-                        )
-
-                    except KeyError:
-                        print(
-                            f"[info]: Path {key} not found. "
-                            f"Skipping the entry.",
-                        )
+                    if key not in eln_data_dict:
```
domna, line 310: I think we should refine that at one point. That we just fill in the keys present in the xarray and the keys in the eln and let the internal logic of the parser throw the errors if anything is missing.

But I think for now this hot fix is fine.

domna (approved):

Issue #45, opened 2022-09-14
YAML2NXDL:  Definition name attribute wrongly placed in YAML


shabihsherjeel:
In all cases of NXDLs where the definition is given a "name" attribute, we wrap this in brackets in the YAML equivalent. Instead, the extends attribute should be in the brackets and name should be outside. This is normal behavior for groups and other cases of "inheritance".

For example:
The line here should be NXiv_temp(NXsensor_scan)
The equivalent in XML would be <definition name="NXiv_temp" extends="NXsensor_scan">...

https://github.com/FAIRmat-Experimental/data-modeling/blob/2eb2c0314e61234f36365b948e50fc28cb14fea8/Application_Definitions/new/NXiv_temp.yaml#L15

shabihsherjeel:
Fixed in #46

PR #46, opened 2022-09-14
Fix for YAML name(depends) convention

shabihsherjeel:
This relates to #45

### DIFF
--- a/tests/data/tools/yaml2nxdl_test_data/NXfilelineError1.yml
+++ b/tests/data/tools/yaml2nxdl_test_data/NXfilelineError1.yml
@@ -1,14 +1,14 @@
-#test case for attributes 
+#test case for attributes
 doc: documentation no. 0
 symbols:
   doc: documentation no. 1
   testnamesymbol: test description of symbol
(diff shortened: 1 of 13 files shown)


tests/data/tools/yaml2nxdl_test_data/NXfilelineError1.yml
```diff
@@ -1,14 +1,14 @@
-#test case for attributes 
+#test case for attributes
 doc: documentation no. 0
 symbols:
   doc: documentation no. 1
   testnamesymbol: test description of symbol
 category: application
 extends: my_test_extend
```
aalbino2, line 7: shouldn't be needed anymore

aalbino2: I checked why the test_yaml2nxdl didn't fail after the changes you made to the "extends" keyword. I found it to be okay and you can safely remove line 7

sanbrock (approved):
there is still one unresolved conversation with Andrea

Issue #49, opened 2022-09-21
Incompatible version


TLCFEM:
https://github.com/nomad-coe/nomad-parser-nexus/blob/b7eeda5dd4599d51bdceed8decd1523736d71210/requirements.txt#L17

v2022.6.0 requires python 3.8, conflicts with python 3.7 used in nomad.

shabihsherjeel:
I thought nomad was moving to python3.8. If that's not true, I'll revert it back to a version that works with 3.7. @fekad and I are trying to resolve this in  #48.

Issue #48, opened 2022-09-20
Removing Nomad as a dependency


shabihsherjeel:
While discussing some dependency issues with Adam, he mentioned that it will be nice to remove nomad as a dependency from this package.

We have to test if this package works as a standalone. When this package is installed as a parser, the Nomad package should already be installed. So as a parser it doesn't really need to have nomad as a requirement.

I'll make a branch to test this.

@fekad

PR #50, opened 2022-09-23
Better dependency management for dev, standalone, and a Nomad install

### LINKED ISSUE #40: Dependency management
I looked into different dependency management tools for python and I found three alternatives: poetry, pipenv and pip-compile (from the package pip-tools). 

As I compared them I found that poetry and pipenv try to be a solution for everything, creating environments, packages and doing dependency managements. A nice feature of pipenv is that it also looks into security vulnerabilities of the packages. However, pipenv is only very inconsistently updated and to me the current status of the project is not entirely clear.

Therefore, I found a nice and easy solution would be to use [pip-compile](https://pypi.org/project/pip-tools/). It generates a pinned/freezed `requirements.txt` file from a `requirements.in` (which is unpinned), so the pinned versions can be easily updated. You may also generate the requirements.txt directly from the setup.py requirements (but for nomad-parser they are just read from requirements.txt, so this is not a viable solution here). The changes from the current state would be minimal but it would freeze the whole dependency tree and would certainly solve dependency issues which keep appearing for us.

We started to use it in our ellipsometry tool, so [here](https://github.com/PyEllips/pyElli/tree/master/requirements) you can see some example files for a base installation and two extra requires dependencies (of the form {extra}-requirements.txt).

@shabihsherjeel @sanbrock @aalbino2 @aalbino2 What do you think? We may shortly discuss after DSM tomorrow.

shabihsherjeel:
Fixes #40 #48 and #49.

- [ ] Update all Readme's to reflect how to install for different use cases
- [ ] Make sure definitions folder is collected in the install package


setup.py
```diff
@@ -16,90 +16,6 @@
...
-                'read_nexus = nexusparser.tools.nexus:main',
-                'dataconverter = nexusparser.tools.dataconverter.convert:convert_cli',
-                'yaml2nxdl = nexusparser.tools.yaml2nxdl.yaml2nxdl:launch_tool'
-            ]
-        })
-
-
-if __name__ == '__main__':
-    main()
+setup()
```
sanbrock, line 21: will it ensure packaging the xml files from nexus definitinos?

shabihsherjeel: This is still in draft. There are many other developments happening in Nomad that I am waiting for. But I'll add this point as a TODO above so that we don't forget this before merging.

sanbrock (approved):

PR #57, opened 2022-11-22
Added support for HDF5 files in JSON Map reader

shabihsherjeel:


nexusutils/dataconverter/readers/json_map/reader.py
```diff
@@ -32,12 +33,7 @@ def parse_slice(slice_string):
     slices = slice_string.split(",")
     for index, item in enumerate(slices):
         values = item.split(":")
-        if len(values) == 1:
-            slices[index] = int(values[0])
-        else:
-            if len(values) < 3:
-                values.append("")
-            slices[index] = slice(map(int, filter(lambda x: not x, values)))
+        slices[index] = slice(*[int(x) for x in values])
```
domna, line 36: This solution would not filter empty strings in values as the previous solution. If this is not a problem this is fine.

domna (approved):

nexusutils/dataconverter/readers/json_map/reader.py
domna: Ok this is fine if the slice syntax is fully filled. But do you want to support short notation, e.g. 1: for indexing from one to the end of the array, too? This would fail here.

domna: ok, could be nice to throw an error or warning that this is not supported in case of `len(values) != 3 or not values[0] or not values[-1]`.

nexusutils/dataconverter/readers/json_map/reader.py
```diff
@@ -32,12 +33,7 @@ def parse_slice(slice_string):
     slices = slice_string.split(",")
     for index, item in enumerate(slices):
         values = item.split(":")
-        if len(values) == 1:
-            slices[index] = int(values[0])
-        else:
-            if len(values) < 3:
-                values.append("")
-            slices[index] = slice(map(int, filter(lambda x: not x, values)))
+        slices[index] = slice(*[None if x == '' else int(x) for x in values])
```
shabihsherjeel, line 36: @domna This works for now. I tried "::x", "x:", and ":x:". They all seem to work.

PR #58, opened 2022-12-13
Xps datamodel with cls ap1

RubelMozumder:
HI @domna and @sanbrock,
Still I am working on it trying to fix Errors come from pydocstyle, pylint and so on.


domna (changes requested):
I think the two files should be refactored to also solve the too-many-branches & too-many-locals issues from pylint.

Mainly I think you should check if assignments can be simplified (e.g. just use the value instead of a new variable if it is only used once) and better grouped and the functions should be split up into more subfunctions. We can also go through the files together to properly refactor it.

RubelMozumder:
Pull #58 is deprecated, instead, #61 contains all required changes.

PR #61, opened 2022-12-22
Xps datamodel v new

RubelMozumder:
@sanbrock and @domna, Here is the new pull request instead of #58. I have tried my best to follow @domna comments for the last pull request #58.


nexusutils/dataconverter/helpers.py
```diff
@@ -406,6 +409,9 @@ def validate_data_dict(template, data, nxdl_root: ET.Element):
                 index_of_at = nxdl_path.rindex("@")
                 nxdl_path = nxdl_path[0:index_of_at] + nxdl_path[index_of_at + 1:]
 
+            if nxdl_path in nxdl_path_set:
```
shabihsherjeel, line 412: This will skip over template entries with the same NXDL definition and will not validate the second occurrence. Are we sure we need this?

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,349 @@
...
+import numpy as np
+
+from nexusutils.dataconverter.readers.base.reader import BaseReader
+from nexusutils.dataconverter.readers.xps.reader_utils import XpsDataFileParser
+from nexusutils.dataconverter.readers.utils import flatten_and_replace, FlattenSettings
+from nexusutils.dataconverter.template import Template
+
+np.set_printoptions(threshold=sys.maxsize)
+
+XPS_TOCKEN = "@xps_tocken:"
```
shabihsherjeel, line 37: Consider making these class functions of XPSReader.

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,349 @@
...
+
+    elif dt_typ in (XPS_DATA_TOCKEN, XPS_DETECTOR_TOCKEN):
+        # entries_values = entry:{cycle0_scan0_chan0:xr.data}
+        entries_values = xps_data_dict["data"]
+
+    return entries_values
+
+
+# pylint: disable=too-many-locals
+def fill_data_group(key,
```
shabihsherjeel, line 106: Consider making these class functions as well. You won't need so many arguments and the silent pass by reference will not be needed.

nexusutils/dataconverter/helpers.py
RubelMozumder: @shabihsherjeel, In XPS if it takes around 2 hours to validate each NXDL definition. That is the reason to skip the validation step for the same nxdl occurrence. And I assumed that once the nxdl path is correct it will be correct for all nxdl paths.

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,360 @@
...
+
+        chan_count = "_chan"
+
+        # Iteration over scan
+        for data_var in xr_data.data_vars:
+
+            if chan_count in data_var:
+                detector_num = data_var.split("_chan")[-1]
+                detector_nm = f"detector{detector_num}"
+                DETECTOR_SET.add(detector_nm)
```
domna, line 212: The all uppercase notation is typically reserved for constants in python, i.e. they should not be changed here. Consider just specifying this as lower case.

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,360 @@
...
+                scan_num = data_var.split("_scan")[-1].split("_chan")[0]
+                scan_nm = f"scan_{scan_num}"
+                modified_key = key.replace("entry", entry)
+                modified_key = modified_key.replace("[detector]", f"[{detector_nm}]")
+                modified_key = modified_key.replace("[data]", f"[{scan_nm}]")
+                # key_nxclass = modified_key.replace("/raw", "/@NX_class")
```
domna, line 218: Should be removed if it is not used anymore.

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,360 @@
...
+                modified_key_unit = modified_key + "/@units"
+
+                template[modified_key] = xr_data[data_var].data
+                key_indv_chan_sginal = modified_key.replace("/raw", "/@signal")
+                template[key_indv_chan_sginal] = "raw"
+                # template[key_nxclass] = "NXdata"
```
domna, line 224: Should be removed, too.

nexusutils/dataconverter/template.py
```diff
@@ -165,3 +165,15 @@ def add_entry(self, entry_name):
         template = Template(self)
         template.rename_entry("entry", entry_name, False)
         self.update(template)
+
+    def __delitem__(self, key):
```
domna, line 169: This is not used anymore. Should it be kept or not?

nexusutils/dataconverter/readers/xps/reader.py
RubelMozumder: This gives pylint errors. That is the reason introducing module variables always uppercase.

nexusutils/dataconverter/template.py
RubelMozumder: My opinion is on behalf of keeping this.

nexusutils/dataconverter/readers/xps/reader.py
RubelMozumder: done

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,360 @@
...
+                                    xps_data_dict,
+                                    template,
+                                    ENTRY_SET)
+        if eln_data_dict:
+            fill_template_with_eln_data(eln_data_dict,
+                                        config_dict,
+                                        template,
+                                        ENTRY_SET)
+        else:
+            raise ValueError("Eln file must be submited with some required fileds and attributes.")
```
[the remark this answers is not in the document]
RubelMozumder: Done!

nexusutils/dataconverter/readers/xps/reader.py
RubelMozumder: Done!

nexusutils/dataconverter/readers/xps/reader.py
domna: ok, my bad I thought that it is then seen as a constant. But obviously I'm wrong

nexusutils/dataconverter/template.py
domna: Fine with me. I just wanted to make sure that it wasn't just forgotten

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,405 @@
...
+                entry_set.add(entry)
+                modified_key = key.replace("[entry]", f"[{entry}]")
+                template[modified_key] = ent_value
+                try:
+                    template[f"{modified_key}/@units"] = config_dict[f"{key}/@units"]
+                except KeyError:
+                    pass
+
+
+def extract_atom_types(formula: str) -> Set:
```
domna, line 266: I think you should also check against valid element names, as @mkuebach does in the apm reader.

nexusutils/dataconverter/helpers.py
```diff
@@ -406,6 +409,10 @@ def validate_data_dict(template, data, nxdl_root: ET.Element):
                 index_of_at = nxdl_path.rindex("@")
                 nxdl_path = nxdl_path[0:index_of_at] + nxdl_path[index_of_at + 1:]
 
+            if nxdl_path in nxdl_path_to_type \
+               and isinstance(data[path], nxdl_path_to_type[nxdl_path]):
+                continue
+
```
shabihsherjeel, line 415: This doesn't work as this skips validation. We want to skip asking for the "elem" object below. You can store the elem object in this new dictionary you made. So wo won't need to call get_node_at_nxdl_path as this will be the same. But we still do want to verify that the entry is valid based on the little set of rules we check for.

nexusutils/dataconverter/helpers.py
```diff
@@ -422,6 +429,7 @@ def validate_data_dict(template, data, nxdl_root: ET.Element):
                 if not is_valid_enum:
                     raise Exception(f"The value at {path} should be"
                                     f" one of the following strings: {enums}")
+            nxdl_path_to_type[nxdl_path] = type(data[path])
```
shabihsherjeel, line 432: The comment above is valid for this too.

nexusutils/dataconverter/readers/xps/reader.py
```diff
@@ -0,0 +1,405 @@
...
+
+        final_template = Template()
+        for key, val in template.items():
+            if ("/ENTRY[entry]" not in key) and (val is not None):
+                final_template[key] = val
+
+        return final_template
+
+
+READER = XPSReader
```
shabihsherjeel, line 409: I will vote for having all the functions added for this reader to be part of the class. It fits better and we won't need these ALL_CAPS variables anymore.  @domna what do you think?

nexusutils/dataconverter/readers/xps/reader_utils.py
```diff
@@ -0,0 +1,835 @@
...
+        element.attrib['__parent__'] = parent_element  # type: ignore[assignment]
+        element.attrib['__odr_siblings__'] = child_elmt_ind  # type: ignore[assignment]
+
+        if self.child_nm_reslvers not in parent_element.attrib.keys():
+            parent_element.attrib[self.child_nm_reslvers] = []  # type: ignore[assignment]
+
+        elmt_tag = element.tag
+
+        if child_elmt_ind <= skip_child:
+            pass
```
shabihsherjeel, line 118: This does nothing.

nexusutils/dataconverter/readers/xps/reader_utils.py
```diff
@@ -0,0 +1,835 @@
...
+
+        elif elmt_tag == "sequence":
+
```
shabihsherjeel, line 121: Why the empty space? Also in between these if-elif-blocks we have too many spaces. Not really needed are they?

nexusutils/dataconverter/readers/xps/reader_utils.py
```diff
@@ -0,0 +1,835 @@
...
+                              parent_path)
+
+        elif elmt_tag in name_val_elmt_tag:
+            self.last_element_parser(element,
+                                     parent_path)
+
+        else:
+            raise TypeError("Needs to parse to different type of parser")
+
+    # pylint: disable=too-many-locals
```
shabihsherjeel, line 136: This is not needed. You can remove skip_child below and also make the while loop part of the pass_child_through_parsers function with some renaming. The code will be easier to follow like that.

nexusutils/dataconverter/readers/xps/reader_utils.py
```diff
@@ -0,0 +1,835 @@
...
+        while child_num > 0:
+
+            self.pass_child_through_parsers(element_,
+                                            parent_path,
+                                            child_elmt_ind,
+                                            skip_child)
+            child_num -= 1
+            child_elmt_ind += 1
+
+    # pylint disable=too-many-branches
```
shabihsherjeel, line 175: This check is there for a reason. Don't disable this.

nexusutils/dataconverter/readers/xps/reader_utils.py
```diff
@@ -0,0 +1,835 @@
...
+
+            else:
+                # Check twin siblings
+
+                section_nm_reslvr = self.restructure_value(elmt_attr[key_type_name],
+                                                           "string")
+                section_nm_reslvr = section_nm_reslvr + "_" + str(elmt_attr['__odr_siblings__'])
+                # section_nm_reslvr = self.check_for_siblins_with_same_name(
+                #    section_nm_reslvr, element_
+                # )
```
shabihsherjeel, line 264: Can be removed.

tests/dataconverter/test_readers.py
```diff
@@ -99,7 +99,7 @@ def test_has_correct_read_func(reader):
             template = Template()
             generate_template_from_nxdl(root, template)
 
-            read_data = reader().read(template=Template(template), file_paths=tuple(input_files))
+            read_data = reader().read(template=Template(template), file_paths=list(input_files))
```
shabihsherjeel, line 102: This should be reverted. It's a tuple really. Not a list.

shabihsherjeel (changes requested):

nexusutils/dataconverter/readers/xps/reader.py
domna: Yes I agree that we should put the functions in the class. I feel the general pattern for the other readers is different, since we almost always use stand-alone functions but I think it's better to avoid this non-constant uppercase variables (which are also confusing) and have class variables instead.

nexusutils/dataconverter/helpers.py
RubelMozumder: Using this if clause is the reason not to use get_node_at_nxdl_path(). You right we can store the element in the dict, I would argue that indeed it is. The nxdl_path is the conclusion/summary of the xml element. So, storing an unnecessary XML element will cost more space. @sanbrock also did objection on it. But after discussing we decided to keep it as it is now.

nexusutils/dataconverter/readers/xps/reader.py
sanbrock: Yes, to avoid confusion, we will have to restructure all the readers.

shabihsherjeel (approved):

sanbrock:
LGTM

PR #69, opened 2023-02-01
Fixes linting

domna:


domna:
@shabihsherjeel It's just to quickly fix linting in the ci/cd. Can you have a quick look so I don't break anything (should be fine I think).

shabihsherjeel (approved):
Looks great.

PR #72, opened 2023-02-02
RII parser refinements

domna:
For the associated appdef/base classes see https://github.com/FAIRmat-Experimental/data-modeling/pull/79


domna:
@sanbrock @shabihsherjeel I introduced [lark](https://github.com/lark-parser/lark) as a dependency. I could also think of using this only as an optional dependency as it is just used for parsing the refractive index database rn. But we eventually could provide the formula parsing as a tool inside nexusututils as well. Then I'd keep the lark dependency as is.

nexusutils/dataconverter/helpers.py
```diff
@@ -169,6 +169,7 @@ def is_value_valid_element_of_enum(value, elem) -> Tuple[bool, list]:
     "NX_UINT": (np.ndarray, np.unsignedinteger),
     "NX_NUMBER": (int, float, np.ndarray, np.signedinteger, np.unsignedinteger, np.floating, dict),
     "NX_POSINT": (int, np.ndarray, np.signedinteger),  # > 0 is checked in is_valid_data_field()
+    "NX_COMPLEX": (complex, np.ndarray, np.complex128),
```
shabihsherjeel, line 172: Do you think we should add np.singlecomplex/np.csingle as well? And is this np.complex128 alias platform dependent? I see in the numpy docs that there is another alias which does not mention platform dependency. I guess this is just for x86_64. 

https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.csingle

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
```diff
@@ -55,23 +67,21 @@ def tabulated(table_input: TableInput):
...
     index_from = {
-        'tabulated nk': ['Wavelength', 'n', 'k'],
-        'tabulated n': ["Wavelength", "n"],
-        'tabulated k': ["Wavelength", "k"],
+        "tabulated nk": ["Wavelength", "n", "k"],
+        "tabulated n": ["Wavelength", "n"],
+        "tabulated k": ["Wavelength", "k"],
     }
 
     daf = pd.read_table(
```
shabihsherjeel, line 75: Consider renaming the variable `daf`. I don't know what it stands for. Maybe it's obvious for others.

nexusutils/dataconverter/helpers.py
domna: Yes it seems that `np.complex128` is platform depended. The platform independent equivalent would be `np.cdouble`, I think it's better to just use this instead of `np.complex128`.

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
```diff
@@ -134,54 +146,55 @@ def sellmeier_polynomial(model_input: ModelInput):
...
     """
-    model_input.add_model_name(
-        model_input.dispersion_path,
-        'Sellmeier'
-    )
+    model_input.add_model_name(model_input.dispersion_path, "Sellmeier")
 
     a_units = []
     # Check for valid parameters
     for exp1 in model_input.coeffs[2:7:4]:
```
shabihsherjeel, line 153: Is this `exp1` also an obvious name for these functions?

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
domna: It stands for dataframe. Typically in pandas `df` is used, but this is not to the liking of the linter. But I will rename it to something more descriptive.

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
domna: This stands for exponent 1, which is denoted as `e1` in the formula string. I can also write it out as `exponent1`.

nexusutils/dataconverter/readers/rii_database/reader.py
```diff
@@ -37,77 +43,307 @@ def read_yml_file(filename: str) -> Dict[Any, Any]:
...
+
+
+def read_metadata(yml_file: dict) -> Dict[str, Any]:
+    """Reads metadata from a dispersion yaml file"""
     entries: Dict[str, Any] = {}
+    if "REFERENCES" in yml_file:
+        pass
+
+    if "COMMENTS" in yml_file:
+        pass
```
shabihsherjeel, line 68: Are these if passes necessary? Or is this to be filled up later?

nexusutils/dataconverter/readers/rii_database/reader.py
```diff
@@ -37,77 +43,307 @@ def read_yml_file(filename: str) -> Dict[Any, Any]:
...
+            dispersion.FORMULA_PARSERS[dispersion_relation["type"]](model_input)
             return
 
-        entries[f'{path}/model_name'] = 'Table'
-        entries[f'{disp_path}/model_name'] = 'Table'
-        entries[f'{disp_path}/convention'] = 'n + ik'
-        entries[f'{disp_path}/wavelength'] = daf.index.values
-        entries[f'{disp_path}/wavelength/@units'] = 'micrometer'
-        entries[f'{disp_path}/refractive_index'] = daf['refractive_index'].values
+        raise NotImplementedError(f'No parser for type {dispersion_relation["type"]}')
```
shabihsherjeel, line 104: Consider using an if-elif-else here. We won't need the returns then. And the else can be `raise NotImple...`

nexusutils/dataconverter/helpers.py
shabihsherjeel: What about np.singlecomplex/np.csingle? Should I drop em in too? What do you think? h5py does support 8 byte complex numbers. So I'm hoping it supports np.csingle too.

shabihsherjeel (approved):
It looks great. :+1:

nexusutils/dataconverter/helpers.py
domna: Ah yes, I think it's useful to add all valid types for h5py. I just had my use case in mind where I should only have `np.complex128`. I should also refactor the code parts from `np.complex128` to `np.cdouble`, so I'll also add the `np.csingle`.

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
```diff
@@ -134,54 +146,55 @@ def sellmeier_polynomial(model_input: ModelInput):
...
+        model_input.coeffs[3:8:4],
+        [f"micrometer^(2/{e2})" for e2 in model_input.coeffs[4:9:4]],
     )
-    model_input.add_rep_param(path, 'e1', model_input.coeffs[2:7:4], [''])
-    model_input.add_rep_param(path, 'e2', model_input.coeffs[4:9:4], [''])
+    model_input.add_rep_param(path, "e1", model_input.coeffs[2:7:4], [""])
+    model_input.add_rep_param(path, "e2", model_input.coeffs[4:9:4], [""])
 
     # If there are just 9 parameters don't append polynomial part
     if not model_input.coeffs[9::]:
```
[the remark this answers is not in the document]
domna: No there is no specific reason. I think it just was in my line of thought from the indexing above. I will change it to `len()`. Good to know that len is faster I would have suspected that using not directly is faster (but probably it's slower due to the indexing).

nexusutils/dataconverter/readers/rii_database/reader.py
domna: This I'll add later in a new PR. Right now I don't parse any metadata, yet.

nexusutils/dataconverter/readers/rii_database/reader.py
domna: I actually like this `if...return` style. I somehow find it easier to read than if-elif-else especially for longer if-elif blocks. I can, however, change it if you think if-elif-else fits better to the general coding style in this package.

nexusutils/dataconverter/helpers.py
shabihsherjeel: Thank you!

nexusutils/dataconverter/readers/rii_database/dispersion_functions.py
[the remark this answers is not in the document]
shabihsherjeel: Yeah. That's why I checked. It's a very minor difference of course.

nexusutils/dataconverter/readers/rii_database/reader.py
shabihsherjeel: No, it's fine. I was just mentioning it in case you would prefer that. They both read fine to me.

PR #73, opened 2023-02-10
Dispersion refinements

domna:


domna:
@shabihsherjeel This is mainly refactored into classes (I thought all these nested functions are not too nice). I added, however, requests and requests-cache to the dependencies (requests was there as a sub-dependency before).
The most crucial change to review is probably the changes in `convert.py` I did.

I think with this the parser should also be final for now.

nexusutils/dataconverter/convert.py
```diff
@@ -106,12 +107,16 @@ def convert(input_file: Tuple[str],
...
+            file_paths=input_file,
+            objects=objects
+        )
     else:
-        data = data_reader().read(template=Template(template),  # type: ignore[operator]
-                                  file_paths=input_file)
+        data = data_reader(**kwargs).read(  # type: ignore[operator]
+            template=Template(template),
+            file_paths=input_file
+        )
```
shabihsherjeel, line 119: I'll check the rest later. I just wanted to check this out. 

So you're now passing some parmas to the init func as well. That's cool and nice that it's here now. But my question to you is, should we do the same with the call to the read func then? I thought what I did might be simplifying it for folks who write the readers. But of course this kwargs approach is better, if understood and used right. And it will get rid of this unnecessary if statement to check for `objects`.

domna: Yes, I think it would be nice to insert the **kwargs in the function, too. Then the readers could also define own parameter names, e.g. `key_value_map` or `raw_data`. If people don't really know how to deal with it we still can keep the objects in the example (which also should still just work if we pass `**kwargs` and remove the `objects` parameter from the `convert` functions parameters).

domna: Ah but we'd had to add **kwargs to each readers `read` function. Or do you have a nice approach in mind how to handle this in the base class exclusively?

domna: We could continue to pass it to the `__init__` and have a `read` wrapper which looks something like this
```python

def __init__(self, **kwargs):
    process_class_kwargs(**kwargs) # Not sure if **kwargs copies...
    del kwargs['processed_key'].   # But it's just some boilerplate for handling **kwargs here

   self._reduced_kwargs = kwargs

def read_wrapper(self, template, file_paths):
    return self.read(template, file_paths, **self._reduced_kwargs)
```

```process_class_kwargs``` just consumes the kwargs which should be processed at class initialisation. The rest is passed to read. We would then call `read_wrapper` in the `convert` function. We could probably also drop all kwargs which are not allowed for the parser and log a warning for that (we would need to get a list of params of the read functions... don't know if this is easily possible).

-- just a quick idea which would leave the writing of parsers and cli interfacing untouched.

Comment (shabihsherjeel) [reply nexusutils/dataconverter/convert.py:119]: Something like this can work. But it will be confusing, implementation wise, for newer `_reader.py`'s. It's our own secret implementation of something so common. If we implement **kwargs, then updating the reader codes will definitely be better for documentation. What do you think? If you like, I can help with it. Or we can do it separate to this PR. Your call.

File: nexusutils/dataconverter/readers/rii_database/citation_parser.py [hunk 2]
```diff
@@ -0,0 +1,113 @@
...
+            count = locals().get("count", 0) + 1
+            name = f"{basename}{count}"
+
+        base_path = f"{path}/{reference_identifier}[{name}]"
+        if self.url:
+            template[f"{base_path}/url"] = self.url
+        if self.doi:
+            template[f"{base_path}/doi"] = self.doi
+        if self.bibtex:
+            template[f"{base_path}/bibtex"] = self.bibtex
```
shabihsherjeel, line 110: Do you think we should add this behavior to the template as a default:

```python
__set__(self, instance, value):
   if value:
      self.value = value
```

domna: I feel it's better to keep it explicit. Here, it wouldn't even work as this is a "normal" dict with which the actual template of type `Template` gets updated with.

What would happen with `None` values in the template being returned by the read function? Would this result in an error or would the keys just be ignored for non-required fields? Maybe this could be a good default behaviour, it the field is explicitly None (and not None-like, like "") to just drop the entry if it is not required.

shabihsherjeel: The __set__ was wrong. I actually was imagining this one here: https://github.com/nomad-coe/nomad-parser-nexus/blob/16614122a281ce5f5d245c90130ede147c988510/nexusutils/dataconverter/template.py#L66

I say we check ```v``` there and if it is ```None``` we don't do anything. So the previously set key by the generation stays as is. No new ```None``` key will be added. I think I also check requirements explicitly in case someone makes a new template and sends that back in the read function.

nexusutils/dataconverter/convert.py
domna: I think it's good to have the **kwargs approach. Maybe it's better then to just pass the **kwargs to the read function and not to the `__init__` of the class, my solution was just thinking about how it could be if we pass it to both. In this case we wouldn't want to have the parameter in both, the init and the read function.

I can just do it. I think it won't be to hard to update. We can just quickly do it here, the merge is not too urgent.

nexusutils/dataconverter/readers/rii_database/citation_parser.py
domna: Ah yes this would work. I just realised that it is doing exactly what I wrote above, because updating the template would just use `__setitem__`. So yes, let's do it like this.

shabihsherjeel:
It looks all good to me. :+1:

nexusutils/dataconverter/convert.py
domna: It's updated. We don't even have to change anything in the `read` funcs as objects will be passed automatically and is set to None in the base reader. The only change is that `objects` now has to be passed explicitly where before it could also be passed as `*args`. I think nobody ever used a non-keyworded objects argument, so it'll be fine.

nexusutils/dataconverter/readers/rii_database/citation_parser.py
domna: I added a check. However, it lets the `test_add_entry` fail. I couldn't really see why, because it only adds to the internal dicts, which should not be affected by setitem. @shabihsherjeel any idea?

domna: I commited the failing change so you can see it in the ci/cd

shabihsherjeel: I'm fixing these. I'll push a commit in a bit.

PR #74, opened 2023-03-09
New versioning and ci/cd

domna:
I added also a ci/cd for pytesting with different python versions (so we actually know what versions to support). The pytest in `pylint.yml` is currently duplicate


shabihsherjeel (approved):
The changes seem fine. The CI is also failing on the master branch currently.

PR #75, opened 2023-03-13
Fixes on the CI and badges for tests, coverage, etc.

shabihsherjeel:

PR #78, opened 2023-03-22
Nyaml2nxdl bug

RubelMozumder:
This seams work perfectly still not taking care of comments in forward and backward convert.

For comments I have another branch `nyaml2nxdl_comment_conversion`:
 Self Comments for dev:
  1. yaml comments can be collected in CommentCollector class:
        TODO: Needs to track of parent yaml node such as NX-fields, NX-attributes and NX-Group etc
  2. XML comments also can be collected by CommentCollector class, need to test whether the class working for XML comments perfectly or not.


nexusutils/nyaml2nxdl/nyaml2nxdl_helper.py
```diff
@@ -0,0 +1,90 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# Yaml library does not except the keys (escapechar "\t" and yaml separator ":")
+# So the corresponding value is to skip them and
+# and also carefull about this order
+ESCAPE_CHAR_DICT = {":": "\':\'",
+                    "\t": "    "}
```
sanbrock, line 31: Does not the conversion  : -> ':' what Florian also mentioned cause any problems?

tests/data/nyaml2nxdl/NXellipsometry-docCheck.yaml
```diff
@@ -1,588 +1,543 @@
...
-# the document has the following main elements:
-# instrument used and is characteristics
-# measured data, the discription of the sample and what was measured about it
-# derived parameters: extra parameters derived in the measurement software
-
-# one would need a Class/ separate entry for the sample description which can link to and should contain a full description of the composition.
-# A free description as text might be not enough or unclear. Especially if one deals with many layer structured samples and optically anisotropic
-# samples, like triclinic one
-#
-# Comments are moved to the Development_notes.md file
```
sanbrock, line 13: do you plan to put comments back when the comment preservation is implemented?

tests/data/nyaml2nxdl/NXnested_symbols.yaml
```diff
@@ -1,12 +1,12 @@
...
 symbols:
   doc: teststring
   nfa: Number of fast axes (acquired simutaneously) e.g. emission angle, kinetic energy
   nsa: Number of slow axes (acquired scanning a physical quantity) e.g. lens voltage, photon energy or temperature
   nx: Number of points in the first angular direction
   ne: Number of points in the energy dispersion direction
-category: base
-doc: Test case for verifying handling of symbols inside a nexus class in nested layers of the hierarchy
-NXentry:
+NXentry(NXobject):
```
sanbrock, line 9: At definition level, a missing superclass could be automatically take NXobject, so the line above could be acceptable as it was: 'NXentry:'
This can be added after the SKM.

sanbrock (approved):

sanbrock (changes requested):
please make sure that ci/cd passes

nexusutils/nyaml2nxdl/nyaml2nxdl_helper.py
domna: @sanbrock This does not necessarily create problems it is just inconvenient and in my small tests it worked to use this in pyyaml. It would therefore be nice to not having to type this in the yaml files, but this can also be fixed at a later stage.

domna:
> please make sure that ci/cd passes

The problem is that the master branch is currently failing with the exact same error we have here. We need to merge #75 first and then rebase here to fix the ci/cd

nexusutils/nyaml2nxdl/nyaml2nxdl_helper.py
RubelMozumder: Because of ':' yaml to nxdl giving error in file NXlog_parsed.yaml for line:

   `deprecated: see: https://github.com/nexusformat/definitions/issues/639`

tests/data/nyaml2nxdl/NXellipsometry-docCheck.yaml
RubelMozumder: Yes, I have plan for that.

aalbino2 (approved):
I find the code much more robust now, it is perfectly dealing with all the subtleties of the nxdl fileformat. I also like the new stdoutput of the --verbose flag

RubelMozumder:
> 

Done!

PR #88, opened 2023-04-18
Rename to pynxtools

domna:
As the name `nexusutils` seems to be taken on pypi I propose using `nexustools`, which is free.

See the test upload at https://pypi.org/project/nexustools/.
I did, however, while testing create a version 0.1.0, which will never be usable again even after deletion (pypi rules). We just need to be aware of this and either skip it or make 0.1.1 the next version.

Maybe it would also be good to rename this repository then, too and move it to https://github.com/FAIRmat-NFDI? It would be good to do this before the release, because the badges also refer to the repository and would break otherwise (or we need to re-publish or keep the old repo until the next release).

@shabihsherjeel @sanbrock


shabihsherjeel:
I'm happy with the name `nexustools`.

I'm also happy to go with the lowest version possible. We can go with 0.1.1 and just imagine that all the versions before that we didn't publish on Pypi were up to 0.1.0. It will save us some trouble figuring this out again in the future. For anyone who installs the package it is just an incremental version number.

Let's move it before release as you suggest then.

shabihsherjeel (approved):
We should remember to change the badge url in the Readme when we move it to FAIRMAT-NFDI.

domna:
Should we also update the folder structure then? Otherwise users would still have to refer to nexusutils, which may be a little odd.

PR #90, opened 2023-04-18
Nexus read

RubelMozumder:
@shabihsherjeel, After you are done with your code you can merge this branch.

PR #91, opened 2023-04-18
Inheritance check dataconverter

shabihsherjeel:
Depends on #90 for some CI tests. But this branch has reader tests failing. These can be handled here. Once #90 has been merged to master, I'll pull in the changes.

### DIFF
--- a/pynxtools/dataconverter/helpers.py
+++ b/pynxtools/dataconverter/helpers.py
@@ -38,9 +38,32 @@ def is_a_lone_group(xml_element) -> bool:
     return True
 
 
-def generate_template_from_nxdl(root, template, path="", nxdl_root=None):
+def get_all_defined_required_children(nxdl_path, nxdl_name):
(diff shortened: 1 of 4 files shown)


nexusutils/nexus/nexus.py
```diff
@@ -1104,32 +1126,109 @@ def axis_helper(dim, nxdata, signal, axes, logger):
...
         self.in_file.close()
 
 
-def main():
+@click.command()
+@click.option(
+    '--nexus-file',
+    required=False,
+    default=None,
+    help=('NeXus file with extension .nxs to learn NeXus different concept'
```
domna, line 1214: I think these help strings should also contain the slightly more verbose explanation from the README

domna (approved):
Looks good. I did not review it in super much detail as I'm also not too involved in the internal workings around this part, but regarding the passing tests and structure of this it seems good to me.

PR #93, opened 2023-04-19
Renames to pynxtools

domna:
I closed #88 to work from a clean repo.

As agreed in the TF meeting yesterday we rename nexusutils to pynxtools.
Wherever we refer the package we use pynxtools (in lower case). When we state it somewhere in free from text we use pyNXtools.

@shabihsherjeel Do you think we should use pyNXtools in the name in pyproject toml. Then people would need to `pip install pyNXtools` and `import pynxtools`. I typically prefer all lowercase, so I don't need to use the shift keys but I think it comes down to preference.

For now I only replaced everything to the lower case variant and we may can update it as soon as we move this repo and fix the urls (mainly in the readme).

I also claimed pynxtools on pypi already (therefore the version 0.0.999 is not available anymore 😎)

Publishing should only be done after moving to the FAIRmat-NFDI organization.


shabihsherjeel:
I vote for keeping it all lower case for things like pip install, etc. We can use pyNXtools for titles to Readme's, papers, etc. This is also not a necessity or rule. It just looks easier to read quickly. 

So to answer concisely: No. We shouldn't use pyNXtools in the name in pyproject.toml.

shabihsherjeel (approved):

PR #94, opened 2023-04-20
Moving extract_atom types functoin from reader/xps to nexusutils/help…

RubelMozumder:
…er module.

### DIFF
--- a/nexusutils/dataconverter/helpers.py
+++ b/nexusutils/dataconverter/helpers.py
@@ -476,8 +476,9 @@ def convert_to_hill(atoms_typ):
         atom_list.append('C')
     if 'H' in atoms_typ:
         atom_list.append('H')
-    for char in atom_list:
-        atoms_typ.remove(char)
+    if atom_list != []:
+        for char in atom_list:
+            atoms_typ.remove(char)
     return atom_list + list(atoms_typ)
 
 
--- a/nexusutils/dataconverter/helpers.py
+++ b/nexusutils/dataconverter/helpers.py
@@ -476,7 +476,7 @@ def convert_to_hill(atoms_typ):
         atom_list.append('C')
     if 'H' in atoms_typ:
         atom_list.append('H')
-    if atom_list != []:
+    if atom_list:
         for char in atom_list:
             atoms_typ.remove(char)
     return atom_list + list(atoms_typ)
--- a/nexusutils/dataconverter/helpers.py
+++ b/nexusutils/dataconverter/helpers.py
@@ -17,7 +17,7 @@
 #
 """Helper functions commonly used by the convert routine."""
 
-from typing import Tuple, Callable, Union, Set, List
+from typing import Tuple, Callable, Union
 import re
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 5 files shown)


mkuehbach:
Handling of atoms does work which is logically unrelated to parsing so instead of placing these functions in the generic helpers.py better would be to make py file with a speaking name were all this atom handling is stored this file can be in the same place like helpers at the same level of the directory tree in this way one could then import it selectively where needed, I agree that such functions should be in the generic place for the dataconverter as it is indeed cross-reader

nexusutils/dataconverter/helpers.py
```diff
@@ -466,7 +466,22 @@ def check_for_valid_atom_types(atoms: Union[str, list]):
                              f"check for correct element name")
 
 
-def extract_atom_types(formula: str, mode: str = 'hill') -> Union[Set, List]:
+def convert_to_hill(atoms_typ):
+    """Convert list of atom into to hill."""
+    if not isinstance(atoms_typ, list) and isinstance(atoms_typ, set):
```
mkuehbach, line 471: there should be a check which catches the case when atoms_typ is empty or neither list nor set

nexusutils/dataconverter/helpers.py
```diff
@@ -99,7 +100,7 @@ def convert_nexus_to_suggested_name(nexus_name):
     return nexus_name[2:]
 
 
-def convert_data_converter_entry_to_nxdl_path_entry(entry) -> str:
+def convert_data_converter_entry_to_nxdl_path_entry(entry):
     """
```
sanbrock, line 104: The method is not modified, so how come that it does not return a str anymore?

nexusutils/dataconverter/helpers.py
mkuehbach: what if atoms_typ has D (deuterium)?

tests/dataconverter/test_helpers.py
```diff
@@ -291,3 +291,16 @@ def test_validate_data_dict(data_dict, error_message, template, nxdl_root, reque
...
     assert helpers.path_in_data_dict(nxdl_path, template) == expected
+
+
+def test_atom_type_extractor_and_hill_conversion():
+    """
+        Test atom type extractor and conversion to hill
+    """
+
+    test_chemical_formula = "(C38H54S4)n(NaO2)5(CH4)NH3B"
+    expected_atom_types = ['C', 'H', 'B', 'N', 'Na', 'O', 'S']
```
mkuehbach, line 302: should test with all words available in the Hill vocabulary

mkuehbach (changes requested):

nexusutils/dataconverter/helpers.py
```diff
@@ -449,3 +450,69 @@ def get_first_group(root):
...
+    return atom_list + list(atoms_typ)
+
+
+def extract_atom_types(formula, mode='hill'):
+    """Extract atom types form chemical formula."""
+
+    atom_types: set = set()
+    element: str = ""
+    # tested with "(C38H54S4)n(NaO2)5(CH4)NH3B"
+    for char in formula:
```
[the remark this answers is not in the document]
sanbrock: In fact lines 343-345 are doing the job there.  The question if this code should do the same or not? And also what Markus asked if and how 'D' is handled, etc.

sanbrock (approved):
questions should be answered after the merge

PR #101, opened 2023-04-28
Ellipsometry reader + mock function

cmmngr:
Changes in ellips reader:

- Optional use of mock function for ellipsometry data
- Automatic labelling of data labelsdepending on data type and angle of incidence
- Writing atom_types as list (if entered as string in eln_data.yaml)

Mock function:

- random atom types, chemical formula and sample name
- modifies spectral range, incident angles
- random choice of data type (psi/Delta, tan(psi)/cos(delta), Mueller matrix)
- distorts data + adds noise to data


sanbrock (approved):
LGTM

PR #104, opened 2023-05-03
apm-em-matwerk-sprint14-pre1

mkuehbach:


mkuehbach:
Apart from linting issues not related to any of the tools which are updated by this feature branch fixing and updating (apm, em_om, em_spctrscpy) there is a conflict now between used numpy and ase

mkuehbach:
The issue manifests in the code for synthesizing the atom probe examples for the search sprint, which internally heavily relies on ase functionalities. Specifically the issue is that lattice/bravais.py is used, in ase=3.19.0 this function uses (line 252ish) still np.float, but with numpy>=1.20 this became deprecated. Given that we currently would like to use ase==3.19.0 but also a numpy newer even than >=1.2x we face the conflict. Possible solutions have to be discussed:
- [ ] Either replace the ase calls by hand-written code
- [ ] Update ase to at least an as high version as when the ase developers fixed their code (not only in bravais.py) to become compliant with newer numpy versions, (preferred solution) the relevant commit which fixed this in ase is this one:
https://github.com/rosswhitfield/ase/commits/76ec28edffbc8cdd97d83b1e90bcdbe30fff150b/ase/lattice/bravais.py see the commits related to "more deprecation fixes" based on the time line of the ase project and the time stamp of this "more deprecation fixes" the fixes were included in say
https://github.com/rosswhitfield/ase/tags
3.20.1 and 3.21.0 so using an ase as 3.19.x is not sufficient, the reason why this likely was not a problem is because being more picky with ase versions started for pynxtools at about the same time when we hadnt had a clean deployment merge (but maybe I am wrong here on this last part, anyway the conflict is real
- [ ] make the synthetic structure generation for the apm reader a developer feature and take it out of the nomad deployment

sanbrock:
An alternative is to pip install a specific lib with its dependencies into a specific dir. This can be then loaded on demand from that location.  
https://skippd.medium.com/handling-multiple-versions-of-a-package-in-python-73055998ce73

mkuehbach:
Currently solved by explicit documentation for developers how to generate synthetic apm data but switched off for general use of pynxtools in nomad see commit 82ba83959f120e98fbfe0626fb4b9dd5d1099549

sanbrock (approved):
LGTM

PR #107, opened 2023-05-08
Changes is_mock to convert arg

domna:


pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -202,7 +202,7 @@ class EllipsometryReader(BaseReader):
     supported_nxdls = ["NXellipsometry"]
 
     @staticmethod
-    def populate_header_dict_with_datasets(file_paths):
+    def populate_header_dict_with_datasets(file_paths, is_mock=False):
```
cmmngr, line 205: 'is_mock' is set to 'False' here and below (in line 292)

cmmngr: this line should read:
def populate_header_dict_with_datasets(file_paths, is_mock):

PR #109, opened 2023-05-09
DEBUG: Deleting intermeadiate generated file.

RubelMozumder:
Note: The temp files created in test 'test_fileline_error' mainly come from the intermediate step and the test was intended to fail. So, the created tmp file was not deleted as the entire process was not done. For that reason, the file has been created from the test by hard coding.


domna (approved):
LGTM and problem of residual files is solved for me locally.

PR #110, opened 2023-05-10
Reader for native SPECS SLE format

lukaspie:
* Implement reader for different versions of .sle file. 
    *  These parsers read the schedule (as xml) as well as the data from the SLE datafiles (which are SQLlite databases)
    *  The output of the SleParsers is a flattened list of dictionaries, which each dict containing the data and metadata of one 
        spectrum (including the individual channel scans).

* Mapping from parser output  to a dictionary that has a similar structure as the output of the XmlSpecs parser, which can then be used to write the nxmpes structure using config files.

* Different config files for xml and sle parsing are read automatically based on the file extension.
    * Config file for SLE is already compatible with NXmpes_xps application definition.

* Keyword arguments can be passed from convert to the individual parsers. For SLE files, one can e.g. remove the initial position alignment scans which improves the reading time.

Not working:
* Channel readout and calibration (see method _get_calibrated_data in xps_reader_utils). This is related to the initial parsing of the data from the SLE file and can be fixed later. At the moment, the "calibrated" data is currently just the scan of the first channel.
* For the data field, some information that is now hardcoded in the config_file is also available in the file, but parsing from the file currently breaks the code (because of XPS_DATA_TOCKEN).


pynxtools/dataconverter/readers/xps/reader.py
```diff
@@ -46,18 +46,19 @@
...
     'Source': 'SOURCE[source]',
+    'Beam': 'BEAM[beam]',
+    'Analyser': 'ELECTRONANALYSER[electronanalyser]',
     'Collectioncolumn': 'COLLECTIONCOLUMN[collectioncolumn]',
     'Energydispersion': 'ENERGYDISPERSION[energydispersion]',
-    'Detector': 'DETECTOR[detector]'
+    'Detector': 'DETECTOR[detector]',
+    'Manipulator': 'MANIPULATOR[manipulator]',
+    'Sample': 'SAMPLE[sample]',
+    'Data': 'DATA[data]',
```
sanbrock, line 61: Any reason for these changes, since only only the order has been modified.

lukaspie: The order was changed to reflect the order in the application definition. Probably not necessary, but may be easier to adjust once the app-def changes.

sanbrock (approved):
LGTM

PR #112, opened 2023-05-19
Transmission reader: Fixes reading of detector ranges

domna:
There was an error in the reading of asc files if the measurement was only performed in one detector range. In this case no separate slits, gain and times are noted in the file which was failing during reading of the file.

### DIFF
--- a/pynxtools/dataconverter/readers/transmission/reader.py
+++ b/pynxtools/dataconverter/readers/transmission/reader.py
@@ -33,10 +33,14 @@ from pynxtools.dataconverter.readers.utils import parse_json, parse_yml
 METADATA_MAP: Dict[str, Any] = {
     "/ENTRY[entry]/SAMPLE[sample]/name": 8,
     "/ENTRY[entry]/start_time": mpars.read_start_date,
-    "/ENTRY[entry]/instrument/sample_attenuator/attenuator_transmission": mpars.read_sample_attenuator,
-    "/ENTRY[entry]/instrument/ref_attenuator/attenuator_transmission": mpars.read_ref_attenuator,
-    "/ENTRY[entry]/instrument/spectrometer/GRATING[grating]/wavelength_range": mpars.read_uv_monochromator_range,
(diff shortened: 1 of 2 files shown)


pynxtools/dataconverter/readers/transmission/reader.py
```diff
@@ -149,25 +146,36 @@ def read_detectors(metadata: list) -> Dict[str, Any]:
...
         )
     )
 
-    for name, idx in zip(["PbS", "InGaAs"], [1, 0]):
+    for name, idx, slit, time, gain in zip(
+        ["PbS", "InGaAs"],
+        [1, 0],
+        detector_slits[1:],
+        detector_times[1:],
+        detector_gains[1:],
```
shabihsherjeel, line 169: I'm guessing this is the important change here? This looks fine how its being used below. If so, this is ready to merge. I'll approve this unless something else needs attention to.

shabihsherjeel (approved):

pynxtools/dataconverter/readers/transmission/reader.py
```diff
@@ -149,25 +146,36 @@ def read_detectors(metadata: list) -> Dict[str, Any]:
...
+    wavelength_ranges = (
         [mpars.MIN_WAVELENGTH] + detector_changes[::-1] + [mpars.MAX_WAVELENGTH]
+    )
 
     template.update(
         convert_detector_to_template(
-            "PMT", detector_slits[2], detector_times[2],
-            None, 2, [wavelength_ranges[0], wavelength_ranges[1]]
+            "PMT",
+            detector_slits[-1],
```
domna, line 156: This is also relevant to use -1 instead of 2 here. This line was actually failing if there was only one detector

pynxtools/dataconverter/readers/transmission/reader.py
domna: Yeah, this and the other line I added a comment for. Sorry that I committed my auto-formatting here, too

PR #113, opened 2023-05-22
Full visit and support for attribute inheritances

sanbrock:


pynxtools/nexus/nexus.py
```diff
@@ -574,6 +583,72 @@ def get_node_concept_path(elem):
...
+        # NX_class is a compulsory attribute for groups in a nexus file
+        # which should match the type of the corresponding NXDL element
+        if attr == 'NX_class' and not isinstance(hdf_node, h5py.Dataset) and elem_index == 0:
+            elem = None
+            logger, doc, attr = write_doc_string(logger, doc, attr)
+            new_elem = elem
+            break
+        # units category is a compulsory attribute for any fields
+        if attr == 'units' and isinstance(hdf_node, h5py.Dataset):
+            req_str = "<<REQUIRED>>"
```
RubelMozumder, line 602: I am wondering if it is possible to use `get_required_string()` instead of hard coding <<REQUIRED>>.

RubelMozumder (approved):

pynxtools/nexus/nexus.py
sanbrock: Curently, get_required_string(nxdl_elem) only returns requirements defined in the nxdl definition files. On he other hand the above requirements come from NeXus conventions which are not written in the nxdl files.

PR #117, opened 2023-05-26
Apm em sprint14 p2

mkuehbach:


mkuehbach:
From my side I am waiting now for Benedikt to grant permission for uploading the example file, in the meantime I continue testing this with nomad, please observe that the definitions have been updated also to point now to the latest

mkuehbach:
Timeout from coveralls :(

domna:
Please check if the updated dependencies still work with the dependencies in nomad before merging this.

pynxtools/dataconverter/convert.py
```diff
@@ -75,7 +75,8 @@ def convert(input_file: Tuple[str],
     # Reading in the NXDL and generating a template
     definitions_path = nexus.get_nexus_definitions_path()
     if nxdl == "NXtest":
-        nxdl_path = os.path.join("tests", "data", "dataconverter", "NXtest.nxdl.xml")
+        nxdl_path = os.path.join(f"{definitions_path}/../../", "tests", "data",
```
sanbrock, line 78: rather use the module path, because the locations of the nexus definition can be overridden by environment variable

tests/data/dataconverter/readers/apm/nomad_oasis_eln_schema_for_nx_apm/nxapm.schema.archive.yaml
```diff
@@ -3,16 +3,12 @@
...
-  sections: # section definitions what were back in the old days msection base classes
-    # Operator:
-    # Specimen:
+  # 'ELN/matching NXapm NeXus appdef data schema for atom probe'
+  sections: # section definitions what were back in the old days msections
     AtomProbeMicroscopy: # the actual schema
-      # nomad.datamodel.metainfo.eln.NexusParser
       base_sections:
-        - 'nomad.datamodel.metainfo.eln.NexusParser'
+        - 'nomad.datamodel.metainfo.eln.NexusDataConverter'
```
sanbrock, line 10: shall it be modified in the other ELN schemas when updating the NOMAD modules?

sanbrock (changes requested):

tests/data/dataconverter/readers/apm/nomad_oasis_eln_schema_for_nx_apm/nxapm.schema.archive.yaml
mkuehbach: Yes, this function was renamed as I can confirm also using the update-north nomad-FAIR branch on the 2023/05/28 (with which I am testing the examples)

pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -160,6 +160,14 @@ def read(self,
 
         ensure_all_required_fields_exist(template, new_template)
 
+        debugging = False
+        if debugging is True:
+            print("Reporting state of template before passing to HDF5 writing...")
+            for keyword, value in new_template.items():
+                print(f"{keyword}, {value}")
```
sherjeelshabih, line 167: Do we really need these kind of debug messages left? If we do need them, let's use the logging module so they can be correctly handled by executing code like Nomad running this instead of having print statements over an if statement.

pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -160,6 +160,14 @@ def read(self,
...
+                # print(type(new_template[keyword]))
+                # print(new_template[keyword])
```
sherjeelshabih, line 169: Commented out code is just unnecessary. It's not executing anything.

tests/dataconverter/test_helpers.py
```diff
@@ -17,6 +17,9 @@
 #
 """Test cases for the helper functions used by the DataConverter."""
 
+# pylint: disable=deprecated-module
+# distutils will become deprecated with py3.12
+
 import xml.etree.ElementTree as ET
 import os
 from distutils import file_util
```
domna, line 25: We should start using:
```suggestion
from setuptools import distutils
```

and then using `distutils.file_util` in the code

domna:
This line will probably fail in nomad:
https://github.com/FAIRmat-NFDI/pynxtools/blob/9143bc086bb3ac19b20b3cbc895f7e4abe2cd914/pynxtools/nexus/nexus.py#L1225

Should be:
` d_inq_nd=None, c_inq_nd=None): `

This is fixed in master already. So after merging this it should be fine (but the respective MR in nomad with this branch will fail).

pyproject.toml
```diff
@@ -37,6 +37,9 @@ dependencies = [
     "kikuchipy>=0.8.2",
     "pyxem>=0.14.2",
     "zipfile37==0.1.3",
+    "nionswift==0.16.8",
+    "nionswift_tool==0.4.18",
+    "tzlocal<=4.3",
```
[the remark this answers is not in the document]
sherjeelshabih: I did find this: https://github.com/FAIRmat-NFDI/pynxtools/blob/9143bc086bb3ac19b20b3cbc895f7e4abe2cd914/pynxtools/dataconverter/readers/em_nion/utils/swift_zipped_project_parser.py#L30

I believe it just uses the nionswift package and not the "_tool"?

pynxtools/dataconverter/readers/json_map/reader.py
mkuehbach: I see your point and agree for end users; given its set to False it will neither be seen nor have an effect upon execution. On the flipside none of these parsers are in yet a state where I would blindly start the verification for new developments and always first check that there is something in the template. As in few cases it works out of the box that code got removed and added in with each revision, yes might not be such a big deal to copy it in all the time, the better the parsers get the more I am becoming supportive for your point to remove this.

pyproject.toml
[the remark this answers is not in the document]
mkuehbach: You are right, I removed nionswift_tool and it continues to work, tzlocal is used by nionswift, the reason to constrain it was explained in the commits 0461... and de34... this branch, fact is nionswift does not constrain their time managing libs strong enough to reflect changes in the interface which I thought caused problems with the CI, anyway it turned out this was not the case but nonetheless nionswift needs tzlocal

tests/dataconverter/test_helpers.py
mkuehbach: On the distutils, agree, implemented and tested your suggestion, tests run fine with this and no longer deprecation warnings

pynxtools/dataconverter/convert.py
mkuehbach: Handled in by the following new commit 45f257

mkuehbach: https://stackoverflow.com/questions/247770/how-to-retrieve-a-modules-path

sanbrock (approved):

PR #119, opened 2023-05-30
Fixes interface for failing nomad tests

domna:


pynxtools/nexus/nexus.py
```diff
@@ -1325,7 +1330,11 @@ def full_visit(self, root, hdf_node, name, func):
     def process_nexus_master_file(self, parser):
         """Process a nexus master file by processing all its nodes and their attributes"""
         self.parser = parser
-        self.in_file = h5py.File(self.input_file_name, 'r')
+        self.in_file = h5py.File(
```
sanbrock, line 1333: if the input files are in a list, we should then process them all in a loop

sanbrock (changes requested):

pynxtools/nexus/nexus.py
domna: The question is if we expect a list here. As far as I see it we just manually created a list in nomad as I now see by the change @mkuehbach did in nomad: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1322/diffs#9461f20970192a34cd19eacc2124c25f818d9f91_289_288

sanbrock: Is not it now fixed in NOMAD? so it sends a single string? Actually, this is also called from the command line, where --nexus-file xxx is used. 
I have no problem with processing only a single files as also in the docstring

sanbrock (approved):

PR #120, opened 2023-05-30
Ellips reader and eln_data

cmmngr:
Updated ellips reader and metadata file (eln_data.yaml)

### DIFF
--- a/examples/ellipsometry/eln_data.yaml
+++ b/examples/ellipsometry/eln_data.yaml
@@ -2,7 +2,7 @@ Data:
   data_identifier: 1
   data_software: WVASE
   data_software/@url: https://www.jawoollam.com/ellipsometry-software/wvase
-  data_software/@version: '3.882'
+  data_software/version: '3.882'
   data_type: Psi/Delta
   spectrum_type: wavelength
   spectrum_unit: Angstroms
@@ -35,9 +35,10 @@ Instrument:
   rotating_element_type: compensator (source side)
   software: CompleteEASE
   software/@url: https://www.jawoollam.com/ellipsometry-software/completeease
-  software/@version: '6.37'
+  software/version: '6.37'
 Sample:
   atom_types: Si, O
+  backside_roughness: false
   chemical_formula: SiO2
   layer_structure: 2nm SiO2 on Si
   sample_history: Commercially purchased sample
--- a/pynxtools/dataconverter/readers/ellips/reader.py
+++ b/pynxtools/dataconverter/readers/ellips/reader.py
@@ -40,7 +40,7 @@ CONVERT_DICT = {
     'Environment': 'environment_conditions',
     'Instrument': 'INSTRUMENT[instrument]',
     'Sample': 'SAMPLE[sample]',
-    'Stage': 'sample_stage',
+    'Sample_stage': 'sample_stage',
     'User': 'USER[user]',
     'Instrument/angle_of_incidence': 'INSTRUMENT[instrument]/angle_of_incidence',
     'Instrument/angle_of_incidence/unit': 'INSTRUMENT[instrument]/angle_of_incidence/@units',
--- a/tests/data/dataconverter/readers/ellips/eln_data.yaml
+++ b/tests/data/dataconverter/readers/ellips/eln_data.yaml
@@ -2,7 +2,7 @@ Data:
   data_identifier: 1
   data_software: WVASE
   data_software/@url: https://www.jawoollam.com/ellipsometry-software/wvase
-  data_software/@version: '3.882'
+  data_software/version: '3.882'
   data_type: Psi/Delta
   spectrum_type: wavelength
   spectrum_unit: Angstroms
@@ -35,9 +35,10 @@ Instrument:
   rotating_element_type: compensator (source side)
   software: CompleteEASE
   software/@url: https://www.jawoollam.com/ellipsometry-software/completeease
-  software/@version: '6.37'
+  software/version: '6.37'
 Sample:
   atom_types: Si, O
+  backside_roughness: false
   chemical_formula: SiO2
   layer_structure: 2nm SiO2 on Si
   sample_history: Commercially purchased sample


pynxtools/dataconverter/readers/ellips/mock.py
```diff
@@ -47,10 +47,10 @@ class MockEllips():
 
     def __init__(self, data_template) -> None:
         self.data = data_template["measured_data"]
-        self.wavelength = data_template["spectrometer/wavelength"]
+        self.wavelength = data_template["data_collection/NAME_spectrum[wavelength_spectrum]"]
```
sanbrock, line 50: Is `data_collection` used in the appdef or it should rather be something like `DATA[data_collection]` here?

cmmngr: `data_collection` is used in the app def. It's `data_collection(NXprocess)`.

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -24,10 +24,78 @@
...
+    'parameters',
+    'plot_name',
+    'sep',
+    'skip',
+    'spectrum_type',
+    'spectrum_unit'
+]
+
+REPLACE_NESTED = {
+    'INSTRUMENT[instrument]': 'INSTRUMENT[instrument]',
```
domna, line 95: This seems to do nothing here.

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -192,6 +283,87 @@ def data_set_dims(whole_data):
...
+        index += counts[0]
+        block_idx.append(index)
+
+    # derived parameters:
+    # takes last but one column from the right (skips empty columns):
+    data_index = 1
+    temp = whole_data[header["colnames"][-data_index]].to_numpy()[
+        block_idx[-1] - 1].astype("float64")
+
+    while temp * 0 != 0:
```
domna, line 308: This is always False.

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -217,6 +389,10 @@ def populate_header_dict_with_datasets(file_paths, is_mock=False):
         """
         header, data_file = populate_header_dict(file_paths)
 
+        # read first N lines of the data file:
+        # with open(data_file) as input_file:
```
domna, line 393: Commented code should just be removed.

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -225,63 +401,33 @@ def populate_header_dict_with_datasets(file_paths, is_mock=False):
...
-                    val.append(f"{key}_{int(angle[1])}deg")
-                index += counts[angle[0]]
-                block_idx.append(index)
 
+        if "atom_types" not in header:
+            header["atom_types"] = extract_atom_types(header["chemical_formula"])
         # Atom types: Convert str to list if atom_types is not a list:
-        if isinstance(header["atom_types"], str):
-            header["atom_types"] = header["atom_types"].split(",")
+        # if isinstance(header["atom_types"], str):
```
domna, line 427: Commented code should just be removed

pynxtools/dataconverter/readers/utils.py
```diff
@@ -41,6 +41,7 @@ class FlattenSettings():
     parent_key: str = "/ENTRY[entry]"
     sep: str = "/"
     is_in_section: bool = False
+    black_list: Optional[list] = None
```
domna, line 44: I'm thinking that we may should use a better name. Something like `ignore_keys`. What do you think?

cmmngr: agreed

pynxtools/dataconverter/readers/ellips/reader.py
cmmngr: Right. This is a placeholder (for an empty list I got an error message...). After updating the ELN schema, some keys to be replaced will be added here.

pynxtools/dataconverter/readers/ellips/reader.py
cmmngr: not if `temp = nan`. There's probably a better way to write that. Suggestions?

domna: Yes, use [`math.isnan`](https://docs.python.org/3/library/math.html#math.isnan)

pynxtools/dataconverter/readers/ellips/reader.py
domna: Ok, if the empty list produces an error we should fix this in the `flatten_and_replace` function as it should also be possible to leave this empty.

I just looked at the code there and it just uses the `items()` method on this replace nested dict, which should be fine for an empty list. So I don't understand where the error was coming from. Could you post your error message you get?

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -43,21 +83,16 @@ def load_header(filename, default):
...
-            clean_header[key] = val
+    clean_header = header
 
     for key, value in default.items():
         if key not in clean_header:
             clean_header[key] = value
+
+    for key, val in header.items():
+        if key == 'sep':
+            clean_header[key] = val.encode("utf-8").decode("unicode_escape")
```
domna, line 94:
```suggestion
    if 'sep' in header:
        clean_header['sep'] = header['sep'].encode("utf-8").decode("unicode_escape")
```

domna (approved):

domna:
Please also remember to update the `eln_data.yaml` and eln files in the north repository (https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-remote-tools-hub/-/tree/develop/docker/ellips/example)

pynxtools/dataconverter/readers/ellips/reader.py
domna:
```suggestion
    while math.isnan(temp):
```

cmmngr: already done

domna: Ah sorry. The diff was alread showing this. But gh diff sometimes gets stuck if there are merge conflicts... (or maybe it's a feature I'm to stupid to understand 🤔)

examples/ellipsometry/eln_data.yaml
```diff
@@ -1,71 +1,73 @@
-acquisition_program/\@url: https://www.jawoollam.com/ellipsometry-software/wvase
```
domna, line 1: I would suggest to use an eln_data.yaml here which was generated by nomad

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -18,16 +18,56 @@
...
+    'Instrument': 'INSTRUMENT[instrument]',
+    'Sample': 'SAMPLE[sample]',
+    'Stage': 'sample_stage',
+    'User': 'USER[user]',
+    'Instrument/angle_of_incidence': 'INSTRUMENT[instrument]/angle_of_incidence',
+    'Instrument/angle_of_incidence/unit': 'INSTRUMENT[instrument]/angle_of_incidence/@units',
+    'column_names': 'data_collection/column_names',
+    'data_error': 'data_collection/data_error',
+    'depolarization': 'derived_parameters/depolarization',
+    'measured_data': 'data_collection/measured_data',
```
domna, line 50:
```suggestion
    'measured_data': 'data_collection/measured_data',
    'software': 'software/program',
    'data_software': 'data_software/program',
```

PR #122, opened 2023-05-31
Fixes #121. Raises exception when wrong dict found.

### LINKED ISSUE #121: Code for handling dicts in Writer should be improved
The code for handling dicts [here](https://github.com/FAIRmat-NFDI/pynxtools/blob/f8dcf39e1ed89b8576763f9bbd9d76b301b2a81f/pynxtools/dataconverter/writer.py#L107) doesn't have a default fallback else in the if-elif-else block. This causes any dictionary being sent to the writer not handled correctly. 

A correct implementation will raise an error explaining which entry the wrong dict was placed in.

sherjeelshabih:


sanbrock (approved):
LGTM

pynxtools/dataconverter/writer.py
```diff
@@ -158,6 +159,11 @@ def handle_dicts_entries(data, grp, entry_name, output_path, path):
                                compression_opts=strength)
         else:
             grp.create_dataset(entry_name, data=data["compress"])
+    else:
+        print("comes here!!")
```
domna, line 163: @sherjeelshabih Is this intentional?

pynxtools/dataconverter/writer.py
```diff
@@ -260,6 +266,8 @@ def add_units_key(dataset, path):
                                                          self.output_path, path])
                     else:
                         dataset = grp.create_dataset(entry_name, data=data)
+            except InvalidDictProvided as exc:
+                print(str(exc))
```
domna, line 270: Wouldn't it be better to re-raise the error here?

sherjeelshabih: This we have resolved over slack.

pynxtools/dataconverter/writer.py
sherjeelshabih: Nope. I'll fix this in a separate PR now.

PR #127, opened 2023-06-14
Adds aoi as scan axis in ellipsometry reader

domna:


pynxtools/dataconverter/helpers.py
```diff
@@ -242,7 +242,7 @@ def is_value_valid_element_of_enum(value, elem) -> Tuple[bool, list]:
     "NX_DATE_TIME": (str,),
     "NX_FLOAT": (float, np.ndarray, np.floating),
     "NX_INT": (int, np.ndarray, np.signedinteger),
-    "NX_UINT": (np.ndarray, np.unsignedinteger),
+    "NX_UINT": (int, np.ndarray, np.unsignedinteger),
```
sherjeelshabih, line 245: This is not compatible as a Python int is either an C int or long if I'm not wrong. POSINT should suffice if you want a definitely positive number. UINT is more about data representation and increasing the range of the possible numbers that can be represented.  If you are actually reading in a UINT, then I would prefer you bring it as is here for validation.

Or is there something else I'm missing here?

pynxtools/dataconverter/helpers.py
```diff
@@ -313,7 +313,7 @@ def is_valid_data_field(value, nxdl_type, path):
             raise ValueError(f"The value at {path} should be of Python type: {accepted_types}"
                              f", as defined in the NXDL as {nxdl_type}.") from exc
 
-    if nxdl_type == "NX_POSINT" and not is_positive_int(value):
+    if nxdl_type in ["NX_POSINT", "NX_UINT"] and not is_positive_int(value):
```
sherjeelshabih, line 316: Related to the comment above. I wouldn't check for a virtual UINT representation. But I'd rather recommend using NX_POSINT.

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -361,7 +363,17 @@ def populate_header_dict_with_datasets(file_paths, is_mock=False):
         header[f"data_collection/NAME_spectrum[{spectrum_type}_spectrum]"] = (
             whole_data[spectrum_type].to_numpy()[0:counts[0]].astype("float64"))
 
+        def write_scan_axis(name: str, values: list, units: str):
+            base_path = f"Env_Conditions/PARAMETER[{name}]"
+            header[f"{base_path}/values"] = values
+            header[f"{base_path}/values/@units"] = units
+            header[f"{base_path}/number_of_parameters"] = len(values)
+            header[f"{base_path}/number_of_parameters/@units"] = ""
```
sherjeelshabih, line 371: Should the number of parameters have units? Is there a mistake in the template generation or NXDL here?

pynxtools/dataconverter/helpers.py
domna: No, I actually just followed what was written in NXopt, which just requires that its a positive number. I would just change it to NX_POSINT in the appdef and revert the changes here.

pynxtools/dataconverter/readers/ellips/reader.py
domna: This I discussed with @sanbrock already. I was also arguing for not having a unit specified on this field but @sanbrock s argument was to set it to `NX_UNITLESS` to not accidentally set a wrong number.

sherjeelshabih (approved):

PR #130, opened 2023-06-16
Add option to nyaml2nxdl to add space to self-closing tags

PeterC-DLS:
Also add short options

### DIFF
--- a/pynxtools/nyaml2nxdl/nyaml2nxdl.py
+++ b/pynxtools/nyaml2nxdl/nyaml2nxdl.py
@@ -180,12 +180,12 @@ possible issues in yaml files'
 )
 @click.option(
     '-s',
-    '--self-closing-tag-space',
+    '--self-closing-tag-nospace',
     is_flag=True,
     default=False,
-    help='Ensure there is a single space character before all self-closing XML tag'
+    help='Omit a single space character before all self-closing XML tag'
 )
-def launch_tool(input_file, verbose, append, check_consistency, self_closing_tag_space):
+def launch_tool(input_file, verbose, append, check_consistency, self_closing_tag_nospace):
     """
         Main function that distiguishes the input file format and launches the tools.
     """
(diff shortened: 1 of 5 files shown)


pynxtools/nyaml2nxdl/nyaml2nxdl.py
```diff
@@ -148,33 +148,44 @@ def split_name_and_extension(file_name):
...
 possible issues in yaml files'
 )
-def launch_tool(input_file, verbose, append, check_consistency):
+@click.option(
+    '-s',
+    '--self-closing-tag-space',
+    is_flag=True,
+    default=False,
+    help='Ensure there is a single space character before all self-closing XML tag'
+)
```
sherjeelshabih, line 187: Do we need this as a flag? Wouldn't it be better to have this as the default?

RubelMozumder (commented):
!!

pynxtools/nyaml2nxdl/nyaml2nxdl.py
PeterC-DLS: I will change and flip the logic.

sherjeelshabih:
Thank your for the changes, @PeterC-DLS. Flipping the logic looks good too. I'll leave the final review to @RubelMozumder and @sanbrock before the merge.

I was thinking that we could just not have such a flag and just add the space in self-closing tags as that seems to be the norm usually. But it really is a choice so I leave it up to the three of you to decide what is the best way to proceed.

RubelMozumder:
It is a matter of choice to have space before the self-closing tag. Now, I assume that with the '-s' option no space before the closing tag, otherwise a default. So, the choice is up to the users on how they want to see XML files.
I would say this makes sense, right? 
So, my flag is ready to go to the master branch. Thanks to @sherjeelshabih for keeping me in conversation.

sanbrock:
the code has already been migrated to a for of nexus_definitions ([FAIRmat-NFDI:nyaml2nxdl-migration](https://github.com/FAIRmat-NFDI/nexus_definitions/tree/nyaml2nxdl-migration)). Can you please migrate this PR there?

PR #133, opened 2023-06-22
Simple verification

domna:
This is a simple verification. I just load everything from a provided file and write it into a dict. There are still several issues:

- [ ] Large datasets shouldn't actually be loaded
- [ ] units are not checked as they are not checked by pynxtools
- [ ] types are not checked

### DIFF
--- a/pynxtools/dataconverter/verify.py
+++ b/pynxtools/dataconverter/verify.py
@@ -1,26 +1,101 @@
 """Verifies a nxs file"""
 import os
+import sys
+from typing import Dict, Union
 import click
 import xml.etree.ElementTree as ET
+import logging
+from h5py import File, Dataset, Group
 
 from pynxtools.dataconverter import helpers
 from pynxtools.dataconverter.template import Template
 from pynxtools.nexus import nexus
 
+logger = logging.getLogger(__name__)
+
+DEBUG_TEMPLATE = 9
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.StreamHandler(sys.stdout))
+
 
 @click.command()
-@click.argument('file')
+@click.argument("file")
 def verify(file: str):
     """Verifies a nexus file"""
-    nxdl = 'NXellipsometry'  # TODO: Read from file
-    definitions_path = nexus.get_nexus_definitions_path()
-    nxdl_path = os.path.join(definitions_path, "contributed_definitions", f"{nxdl}.nxdl.xml")
-    if not os.path.exists(nxdl_path):
-        nxdl_path = os.path.join(definitions_path, "applications", f"{nxdl}.nxdl.xml")
-    if not os.path.exists(nxdl_path):
-        raise FileNotFoundError(f"The nxdl file, {nxdl}, was not found.")
+    def_map: Dict[str, str] = {}
+    with File(file, "r") as h5file:
+        for entry in h5file.keys():
+            if h5file[entry].attrs.get("NX_class") == "NXentry":
+                def_map = {
+                    entry: (
+                        definition := h5file[f"/{entry}/definition"][()].decode("utf8")
+                    )
+                }
+                logger.debug(f"Reading entry '{entry}': {definition}'")
 
-    nxdl_root = ET.parse(nxdl_path).getroot()
+    for entry, nxdl in def_map.items():
+        definitions_path = nexus.get_nexus_definitions_path()
+        nxdl_path = os.path.join(
+            definitions_path, "contributed_definitions", f"{nxdl}.nxdl.xml"
+        )
+        if not os.path.exists(nxdl_path):
+            nxdl_path = os.path.join(
+                definitions_path, "applications", f"{nxdl}.nxdl.xml"
+            )
+        if not os.path.exists(nxdl_path):
+            raise FileNotFoundError(f"The nxdl file, {nxdl}, was not found.")
 
-    template = Template()
-    helpers.generate_template_from_nxdl(nxdl_root, template)
+        nxdl_root = ET.parse(nxdl_path).getroot()
+
+        empty_template = Template()
+        template = Template(empty_template)
+        helpers.generate_template_from_nxdl(nxdl_root, template)
+
+        logger.log(DEBUG_TEMPLATE, "Generated template: %s", template)
+
+        class_map: Dict[str, str] = {}
+
+        def collect_groups(name: str, dataset: Union[Group, Dataset]):
+            if isinstance(dataset, Group) and (
+                nx_class := dataset.attrs.get("NX_class")
+            ):
+                entry_name = name.rsplit("/", 1)[-1]
+                clean_nx_class = nx_class[2:].upper()
+
+                class_map[entry_name] = clean_nx_class
+                logger.debug("Adding class %s to %s", clean_nx_class, entry_name)
+
+        def collect_fields_and_attrs(name: str, dataset: Union[Group, Dataset]):
+            for nx_class in class_map:
+                if name.startswith(nx_class):
+                    name = name.replace(
+                        f"{nx_class}/", f"{class_map[nx_class]}[{nx_class}]/"
+                    )
+
+            if isinstance(dataset, Dataset):
+                logger.debug("Adding field %s/%s", entry_path, name)
+                if isinstance(read_data := dataset[()], bytes):
+                    read_data = read_data.decode("utf-8")
+                template[f"{entry_path}/{name}"] = read_data
+
+            for attr_name, val in dataset.attrs.items():
+                if attr_name == "NX_class":
+                    continue
+                logger.debug("Adding attribute %s/%s/@%s", entry_path, name, attr_name)
+                template[f"{entry_path}/{name}/@{attr_name}"] = val
+
+        entry_path = f"/ENTRY[{entry}]"
+        with File(file, "r") as h5file:
+            # TODO: Check whether h5py does graph traversal
+            # which would ensure visiting groups before their fields.
+            # In this case one visititems is enough.
+            h5file[f"/{entry}"].visititems(collect_groups)
+            h5file[f"/{entry}"].visititems(collect_fields_and_attrs)
+
+        logger.log(DEBUG_TEMPLATE, "Processed template %s", template)
+        helpers.validate_data_dict(empty_template, Template(template), nxdl_root)
+
+        logger.info(
+            f"The entry `{entry}` in file `{file}` is a valid file"
+            f" according to the `{nxdl}` application definition."
+        )
--- a/pynxtools/dataconverter/verify.py
+++ b/pynxtools/dataconverter/verify.py
@@ -47,53 +47,62 @@ def verify(file: str):
 
         nxdl_root = ET.parse(nxdl_path).getroot()
 
-        empty_template = Template()
-        template = Template(empty_template)
-        helpers.generate_template_from_nxdl(nxdl_root, template)
+        ref_template = Template()
+        data_template = Template()
+        helpers.generate_template_from_nxdl(nxdl_root, ref_template)
 
-        logger.log(DEBUG_TEMPLATE, "Generated template: %s", template)
+        logger.log(DEBUG_TEMPLATE, "Reference template: %s", ref_template)
 
         class_map: Dict[str, str] = {}
 
-        def collect_groups(name: str, dataset: Union[Group, Dataset]):
+        def replace_group_names(path: str):
+            for nx_class in class_map:
+                if f"/{nx_class}/" in path or path.startswith(f"{nx_class}/"):
+                    path = path.replace(
+                        f"{nx_class}/", f"{class_map[nx_class]}[{nx_class}]/"
+                    )
+            return path
+
+        def collect_entries(name: str, dataset: Union[Group, Dataset]):
+            clean_name = replace_group_names(name)
             if isinstance(dataset, Group) and (
                 nx_class := dataset.attrs.get("NX_class")
             ):
                 entry_name = name.rsplit("/", 1)[-1]
                 clean_nx_class = nx_class[2:].upper()
 
-                class_map[entry_name] = clean_nx_class
+                is_variadic = True
+                clean_name = replace_group_names(name)
+                for ref_entry in ref_template:
+                    if ref_entry.startswith(f"{entry_path}/{clean_name}"):
+                        is_variadic = False
+                        break
+
+                if is_variadic:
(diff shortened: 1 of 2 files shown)


domna:
This nyaml entry still gives an error `SOLUTE(NXsample_component)`.
The template verification demands the path

```
ValueError: The data entry corresponding to /ENTRY[entry]/SAMPLE[sample]/SOLUTE/solvent 
is required and hasn't been supplied by the reader.
```

while the script here adds this key:
```
"/ENTRY[entry]/SAMPLE[sample]/SAMPLE_COMPONENT[solute2]/solvent": "solvent1"
```

Which one of these should it be: `SOLUTE[solute2]`, `SAMPLE_COMPONENT[solute2]` or `SOLUTE`?

domna:
@sherjeelshabih @sanbrock What do you think of this approach? It is roughly working already (except for the error in the previous comment but I'm actually not sure if this is a general bug in the template generation, @sherjeelshabih ?) and IMO it would be helpful to have a verification tool for nexus files. If you think this is a good path I would continue adding more verification features also in other parts of the code (i.e., the points addressed above) and tests.

sherjeelshabih:
> This nyaml entry still gives an error `SOLUTE(NXsample_component)`. The template verification demands the path
> 
> ```
> ValueError: The data entry corresponding to /ENTRY[entry]/SAMPLE[sample]/SOLUTE/solvent 
> is required and hasn't been supplied by the reader.
> ```
> 
> while the script here adds this key:
> 
> ```
> "/ENTRY[entry]/SAMPLE[sample]/SAMPLE_COMPONENT[solute2]/solvent": "solvent1"
> ```
> 
> Which one of these should it be: `SOLUTE[solute2]`, `SAMPLE_COMPONENT[solute2]` or `SOLUTE`?

I think SOLUTE[solute2] might work. The parts outside the brackets are used to look for the NXDL node. So the function get_node_at_nxdl_path receives a concatenation of what is outside the brackets. You can check if that function finds this NXDL path correctly: /ENTRY/SAMPLE/SOLUTE/solvent. I keep forgetting the exact caps/no caps format that goes in there. But as far as the converter sees this, the outside will be used to get the nxdl node for nexus info it neds. Then it just writes whatever you have in the [brackets] to the hdf5 file.

sherjeelshabih:
> @sherjeelshabih @sanbrock What do you think of this approach? It is roughly working already (except for the error in the previous comment but I'm actually not sure if this is a general bug in the template generation, @sherjeelshabih ?) and IMO it would be helpful to have a verification tool for nexus files. If you think this is a good path I would continue adding more verification features also in other parts of the code (i.e., the points addressed above) and tests.

I really like this approach. I already see that you are using the same validation routine used while converting. This will help improve verification there too then. That's awesome.

I think the not loading of large datasets should have a nice wrapper to allow readers, from the other side, to also somehow tell this verfication routine to not load large sets until it actually writes the hdf5. Then for writing we can have a streamIO object or whatever one calls it. For the verification of hdf5 files this will be easier as we can rely on the h5py's lazyload functionalities. Maybe for this PR we can focus on the simple hdf5 case and just have this in mind.

I think shape and dim aren't checked but types were being checked. Or you mean the types don't deal with what comes out of h5py.dtype?

pynxtools/dataconverter/verify.py
```diff
@@ -0,0 +1,139 @@
...
+@click.command()
+@click.argument("file")
+def verify(file: str):
+    """Verifies a nexus file"""
+    def_map = _get_def_map(file)
+
+    if not def_map:
+        logger.info("Could not find any valid entry in file %s", file)
+    ref_template = Template()
+    data_template = Template()
```
sherjeelshabih, line 67: Just a side note: Can these lines be ommitted?

domna: Yes they can but linting complained that I initialized them in the for-loop. Anyways, I will clean this all up a little before actually merging it. It's was more of a proof-of-concept now.

domna:
> I think SOLUTE[solute2] might work. The parts outside the brackets are used to look for the NXDL node. So the function get_node_at_nxdl_path receives a concatenation of what is outside the brackets. You can check if that function finds this NXDL path correctly: /ENTRY/SAMPLE/SOLUTE/solvent. I keep forgetting the exact caps/no caps format that goes in there. But as far as the converter sees this, the outside will be used to get the nxdl node for nexus info it neds. Then it just writes whatever you have in the [brackets] to the hdf5 file.

ok I will just check what works. SOLUTE[solute2] will be a bit trickier to implement in the reverse way as I'm currently just reading the `NX_class` attribute for this. I would need to get the actual group name from the nxdl to resolve this from the file. Maybe it would be helpful to allow `SAMPLE_COMPONENT[solute2]` for a group named `SOLUTE` in the validation step and just check if the actual naming is correct for the upper/lower combinations (e.g. something like `SOLUTE_fixedpart` should just allow names like `test1_fixedpart`, `test2_fixedpart` etc).

domna:
> > @sherjeelshabih @sanbrock What do you think of this approach? It is roughly working already (except for the error in the previous comment but I'm actually not sure if this is a general bug in the template generation, @sherjeelshabih ?) and IMO it would be helpful to have a verification tool for nexus files. If you think this is a good path I would continue adding more verification features also in other parts of the code (i.e., the points addressed above) and tests.
> 
> I really like this approach. I already see that you are using the same validation routine used while converting. This will help improve verification there too then. That's awesome.

Yes this is why I started this because I felt most of the parts for validation are already there. I just did not have an easy way to use this from the command line.

> I think the not loading of large datasets should have a nice wrapper to allow readers, from the other side, to also somehow tell this verfication routine to not load large sets until it actually writes the hdf5. Then for writing we can have a streamIO object or whatever one calls it. For the verification of hdf5 files this will be easier as we can rely on the h5py's lazyload functionalities. Maybe for this PR we can focus on the simple hdf5 case and just have this in mind.

Yes sounds good 👍️This would also help building this for verification as we could just build a small class to pass instead of the actual data and let the verification routine check its contents against the nxdl.

> I think shape and dim aren't checked but types were being checked. Or you mean the types don't deal with what comes out of h5py.dtype?

I did not actually check the details of what is being checked and what not by the code already (the points are also a somehow chaotic collection of the points I did not want to forget to address here 😅)

domna:
There are still some problems with the upper/lowercase notation in pynxtools which blocks proper validation.
If I try to verify the minimal mpes example (https://box.hu-berlin.de/f/888b21ec9852470fb5e4/) I get the following output:

```
❯ verify_nexus NXmpes_minimal.nxs
The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beam_TYPE/incident_energy is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beam_TYPE/incident_energy/@units is required and hasn't been supplied by the reader.
The unit, /ENTRY[entry]/INSTRUMENT[instrument]/BEAM[beam_probe]/incident_energy/@units = eV, is being written but has no documentation.
The unit, /ENTRY[entry]/data/data/@units = counts, is being written but has no documentation.
The unit, /ENTRY[entry]/data/energy/@units = eV, is being written but has no documentation.
The unit, /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/ENERGYDISPERSION[energydispersion]/pass_energy/@units = eV, is being written but has no documentation.
The entry `entry` in file `NXmpes_minimal.nxs` is a valid file according to the `NXmpes` application definition.
```

I think the unit parts is that the units fix we made is not included here (as this also added support for _checking_ units here). I'll investigate.
@sherjeelshabih Any idea how to fix the `beam_TYPE/incident_energy` problem? Also I think it is related to the error we had in the previous discussion in this thread.

sherjeelshabih:
> There are still some problems with the upper/lowercase notation in pynxtools which blocks proper validation. If I try to verify the minimal mpes example (https://box.hu-berlin.de/f/888b21ec9852470fb5e4/) I get the following output:
> 
> ```
> ❯ verify_nexus NXmpes_minimal.nxs
> The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beam_TYPE/incident_energy is required and hasn't been supplied by the reader.
> The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beam_TYPE/incident_energy/@units is required and hasn't been supplied by the reader.
> The unit, /ENTRY[entry]/INSTRUMENT[instrument]/BEAM[beam_probe]/incident_energy/@units = eV, is being written but has no documentation.
> The unit, /ENTRY[entry]/data/data/@units = counts, is being written but has no documentation.
> The unit, /ENTRY[entry]/data/energy/@units = eV, is being written but has no documentation.
> The unit, /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/ENERGYDISPERSION[energydispersion]/pass_energy/@units = eV, is being written but has no documentation.
> The entry `entry` in file `NXmpes_minimal.nxs` is a valid file according to the `NXmpes` application definition.
> ```
> 
> I think the unit parts is that the units fix we made is not included here (as this also added support for _checking_ units here). I'll investigate. @sherjeelshabih Any idea how to fix the `beam_TYPE/incident_energy` problem? Also I think it is related to the error we had in the previous discussion in this thread.

This of course needs to be fixed. The relevant logic could be added here:

https://github.com/FAIRmat-NFDI/pynxtools/blob/6b3b3d5c372b3d0a062bdf32097929211e6c2e86/pynxtools/dataconverter/helpers.py#L392-L399

There are possible group names where the CAPS_nocaps syntax isn't dictating renamability. We should add those as exceptions. I can help you with this if you don't get to this before I get time. :sweat_smile:

sanbrock:
Note that `NXmpes/ENTRY/INSTRUMENT/beam_TYPE` is actually a specialisation of `NXinstrument/BEAM`  as it is properly picked up in  
[documentation](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXmpes.html#nxmpes-entry-instrument-beam-type-group)  
The problem is that `read_nexus -d /entry/instrument/beam_probe` does not recognise it as  `NXmpes/ENTRY/INSTRUMENT/beam_TYPE` but rather as `NXinstrument/BEAM`:  
`$ read_nexus -d /entry/instrument/beam_probe -f ~/Downloads/NXmpes_minimal.nxs 
DEBUG: ===== GROUP (//entry/instrument/beam_probe [NXmpes::/NXentry/NXinstrument/NXbeam]): <HDF5 group "/entry/instrument/beam_probe" (1 members)>
DEBUG: classpath: ['NXentry', 'NXinstrument', 'NXbeam']
DEBUG: classes:
NXinstrument.nxdl.xml:/BEAM
NXbeam.nxdl.xml:
DEBUG: <<OPTIONAL>>
DEBUG: documentation (NXinstrument.nxdl.xml:/BEAM):
DEBUG: 
DEBUG: documentation (NXbeam.nxdl.xml:):
DEBUG: 
         Properties of the neutron or X-ray beam at a given location.
         
         This group is intended to be referenced
         by beamline component groups within the :ref:`NXinstrument` group or by the :ref:`NXsample` group. This group is
         especially valuable in storing the results of instrument simulations in which it is useful
         to specify the beam profile, time distribution etc. at each beamline component. Otherwise,
         its most likely use is in the :ref:`NXsample` group in which it defines the results of the neutron
         scattering by the sample, e.g., energy transfer, polarizations. Finally, There are cases where the beam is
         considered as a beamline component and this group may be defined as a subgroup directly inside
         :ref:`NXinstrument`, in which case it is recommended that the position of the beam is specified by an
         :ref:`NXtransformations` group, unless the beam is at the origin (which is the sample).
         
         Note that incident_wavelength and related fields can be a scalar values or arrays, depending on the use case.
         To support these use cases, the explicit dimensionality of these fields is not specified, but it can be inferred
         by the presense of and shape of accompanying fields, such as incident_wavelength_weights for a polychromatic beam.
    
DEBUG: ===== ATTRS (//entry/instrument/beam_probe@NX_class)
DEBUG: value: NXbeam 
DEBUG: classpath: ['NXentry', 'NXinstrument', 'NXbeam']
DEBUG: classes:
NXinstrument.nxdl.xml:/BEAM
NXbeam.nxdl.xml:
DEBUG: @NX_class [NX_CHAR]
DEBUG: 
`

sanbrock:
In fact, only leading UPPERs are checked:  
`def get_nx_namefit(hdf_name, name, name_any=False):
    """Checks if an HDF5 node name corresponds to a child of the NXDL element
    uppercase letters in front can be replaced by arbitraty name, but
    uppercase to lowercase match is preferred,
    so such match is counted as a measure of the fit"""`

domna:
> In fact, only leading UPPERs are checked: `def get_nx_namefit(hdf_name, name, name_any=False): """Checks if an HDF5 node name corresponds to a child of the NXDL element uppercase letters in front can be replaced by arbitraty name, but uppercase to lowercase match is preferred, so such match is counted as a measure of the fit"""`

Thank you for pointing to this function. This is very useful. I will investigate whether I can just tune the accepted name there and see if it works then.

sanbrock:
Note that these functions should have been removed from `pynxtools/nexus/utils/nxdl.py` and should be imported and used from `definitions/dev_tools/utils/nxdl.utils.py`.

pynxtools/nexus/nexus.py
```diff
@@ -776,7 +776,7 @@ def process_nexus_master_file(self, parser):
 )
 def main(nexus_file, documentation, concept):
     """The main function to call when used as a script."""
-    logging_format = "%(levelname)s: %(message)s"
+    logging_format = "%(message)s"
```
sanbrock, line 779: Should not this change require the modification of the reference files for the nexus tests?

sanbrock (changes requested):

pynxtools/nexus/nxdl_utils.py
```diff
@@ -95,32 +96,41 @@ def get_nx_class(nxdl_elem):
 
 
 def get_nx_namefit(hdf_name, name, name_any=False):
-    """Checks if an HDF5 node name corresponds to a child of the NXDL element
-    uppercase letters in front can be replaced by arbitraty name, but
-    uppercase to lowercase match is preferred,
-    so such match is counted as a measure of the fit"""
+    """
```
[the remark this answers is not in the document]
domna: Yes, I will move this. Please note that this PR is just a draft, so I'm just testing that everything is working rn. Cleaning and refactoring I will do towards this changes become stable.

domna:
This is superseeded by #333, which includes everything of this PR but uses the new validation algorithm.

PR #137, opened 2023-07-03
Fixes linting

domna:


RubelMozumder (approved):
I think everything looks fine. But still have linter issue. After, fixing that it is ready to merge.

PR #141, opened 2023-07-20
update-north-markus

mkuehbach:
Merge features developed within June (partially during sprint 15) for creating deployable apm-specific OASES in Leoben and Sydney to master so that these can appear in a new nomad oasis deployment


README.md
```diff
@@ -71,6 +71,18 @@ Especially relevant for developers, there exists a basic test framework written
...
+## Run examples
+
+A number of examples exist which document how the tools can be used. For a standalone
+usage convenient jupyter notebooks are available for each tool. To use them jupyter
+and related tools have to be installed in the development environment as follows:
+
+```shell
+python -m pip install jupyter
+python -m pip install jupyterlab
+python -m pip install jupyterlab_h5web
```
sanbrock, line 83: Isn't jupyterlab_h5web enough alone, so the others are installed automatically as dependencies?

tests/data/dataconverter/readers/mpes/Ref_nexus_mpes.log
```diff
@@ -702,25 +665,15 @@ DEBUG - value: 0.11
 DEBUG - classpath: ['NXentry', 'NXinstrument', 'NXbeam', 'NX_NUMBER']
 DEBUG - classes:
 NXmpes.nxdl.xml:/ENTRY/INSTRUMENT/BEAM/incident_energy_spread
-NXbeam.nxdl.xml:/incident_energy_spread
 DEBUG - <<RECOMMENDED>>
 DEBUG - documentation (NXmpes.nxdl.xml:/ENTRY/INSTRUMENT/BEAM/incident_energy_spread):
 DEBUG - 
-DEBUG - documentation (NXbeam.nxdl.xml:/incident_energy_spread):
```
sanbrock, line 709: Are these missing because base class modifications have been removed, like incident_energy_spread from NXbeam?
They will come back when merging Florian's corrections...

sanbrock (approved):
LGTM, but please answer the questions

README.md
mkuehbach: When I worked on this branch in a local environment I was for some reason unable to get jupyterlab running when installing inside a fresh env using only jupyterlab_h5web, I made the same exercise right now and indeed jupyterlab_h5web again try to start jupyter lab out of the box without the two above does not work despite using the right local python3.10 env and not my conda base py3.9 env. Thus, I will keep the above suggestions to users for now as it reads above.

tests/data/dataconverter/readers/mpes/Ref_nexus_mpes.log
mkuehbach: You are right, this work on this branch happened during an interval in time when these already discussed about mpes base class fields were already removed but development for Leoben had to move on and thus I had to adapt to the situation of using a commit of nxdefs during this pullback interval. It is also my assumption that when Florians mpes base class back branch is merged into nxdefs we can and should update the nxdefs on the fairmat anyway and this will demand that I test again then all examples if they work with the current combination of nxdefs + up-to-date and update-north-markus merged into fairmat etc. happy to do this after the merge of both according to how tomorrow's meeting these testing tasks I think is what is needed and what I meant with I will have time this week to prioritize it.

README.md
sanbrock: thanks

mkuehbach: ran into this again today and learned that indeed for py310 the second line is not needed as the pip install jupyter will install jupyterlab and two other dependencies which the jupyterlab_h5web does not pick up

PR #142, opened 2023-07-31
Enable Rubel, updated spm deps and fixing pytests

mkuehbach:


RubelMozumder (approved):
Nice all conflicts are resolved!

PR #136, opened 2023-06-28
Stm version 1

RubelMozumder:
This PR implies:

1. **STM and STS reader**: At this the reader supports two version (Generic 5e, Generic 4.5) of STM and STS experiment (from Sabine and Carlos).
2. **generate_eln** tools with option **--eln-type (allowed values eln, echem_eln)** for simple eln and scheme eln.
3. nexus converter checks the existence of the concept that is referred to be linked.

PR #147, opened 2023-08-24
Revert "Stm version 1"

RubelMozumder:
Reverts FAIRmat-NFDI/pynxtools#136


RubelMozumder:
This is for test purposes on sts reader with any intention to modification.

PR #149, opened 2023-08-25
A quick fix for a bug with req group and opt parent

sherjeelshabih:
@rettigl I hope this fixes your issue. If not feel free to make modifications or comments here for me to fix.

### DIFF
--- a/pynxtools/dataconverter/helpers.py
+++ b/pynxtools/dataconverter/helpers.py
@@ -424,7 +424,7 @@ def does_group_exist(path_to_group, data):
     return False
 
 
-def ensure_all_required_fields_exist(template, data):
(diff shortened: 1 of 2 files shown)


rettigl:
This fix solves the problem that required child groups of optional parent groups become required independent of the presence of elements in the parent group.
However, another bug remains in this context. In the same configuration, even if I provide the required child group and some fields in it, there is still an error message stating that not all required children are present in the optional parent group.
The reason is that when you create the group with fields, you only create the fields as explicit elements in the dictionary, whereas the groups exists as "None" in the template dicitonary, and is not being overwritten. The code currently picks this up as a missing required child.

sherjeelshabih:
#159 supercedes all changes here. I'll close this PR.

PR #155, opened 2023-09-07
Adds test for example warnings and adds `--undocumented` flag

domna:
This ensures that changes do not change the warnings generated when converting the example files.

Currently, this is only for mpes but is parametrized and can be extended to other readers as well. We could also add this for all readers and add a script to generate the keys as a regression file to compare to.

It adds an `--undocumented` field, too. This enables the display of undocumented fields while writing the file. It also defaults to suppress warnings due to undocumented fields, but maybe it's better to default to showing them?

### DIFF
--- a/pynxtools/dataconverter/convert.py
+++ b/pynxtools/dataconverter/convert.py
@@ -36,6 +36,7 @@ from pynxtools.nexus import nexus
 
 
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
+UNDOCUMENTED = 9
 logger.setLevel(logging.INFO)
 logger.addHandler(logging.StreamHandler(sys.stdout))
 
@@ -70,6 +71,7 @@ def convert(input_file: Tuple[str],
             output: str,
             generate_template: bool = False,
             fair: bool = False,
+            undocumented: bool = False,
             **kwargs):
     """The conversion routine that takes the input parameters and calls the necessary functions."""
     # Reading in the NXDL and generating a template
@@ -88,6 +90,9 @@ def convert(input_file: Tuple[str],
             raise FileNotFoundError(f"The nxdl file, {nxdl}, was not found.")
 
     nxdl_root = ET.parse(nxdl_path).getroot()
+
+    if undocumented:
+        logger.setLevel(UNDOCUMENTED)
 
     template = Template()
     helpers.generate_template_from_nxdl(nxdl_root, template)
@@ -120,7 +125,11 @@ def convert(input_file: Tuple[str],
     for path in data.undocumented.keys():
         if "/@default" in path:
             continue
-        logger.warning("The path, %s, is being written but has no documentation.", path)
+        logger.log(
+            UNDOCUMENTED,
+            "The path, %s, is being written but has no documentation.",
+            path
+        )
 
     Writer(data=data, nxdl_path=nxdl_path, output_path=output).write()
 
@@ -170,20 +179,28 @@ def parse_params_file(params_file):
     is_flag=True,
     default=False,
     help='Let the converter know to be stricter in checking the documentation.'
-)  # pylint: disable=too-many-arguments
+)
 @click.option(
     '--params-file',
     type=click.File('r'),
     default=None,
     help='Allows to pass a .yaml file with all the parameters the converter supports.'
 )
+@click.option(
+    '--undocumented',
+    is_flag=True,
+    default=False,
+    help='Shows a log output for all fields which are undocumented'
+)
+# pylint: disable=too-many-arguments
 def convert_cli(input_file: Tuple[str],
                 reader: str,
                 nxdl: str,
                 output: str,
                 generate_template: bool,
                 fair: bool,
-                params_file: str):
+                params_file: str,
+                undocumented: bool):
     """The CLI entrypoint for the convert function"""
     if params_file:
         try:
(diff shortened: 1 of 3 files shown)


sherjeelshabih:
I like the --undocumented flag you've added. Everything looks good. 

I just wanted to know what exactly are we trying to achieve with the regression test here. Do we need to lock the content in? It seems we are trying to ensure consistency in what fields we expect to be undocumented in our NXDL. But this doesn't really test the functionality of the tool to figure out whether these fields are the ones we want to show as warning.

What this test really is checking for is whether our readers/NXDL still have the same undocumented combo as we had before. All I see this doing is adding another step to correct before we push commits. If we really want to enforce some sort of expected content from our NXDL/filled template, we can do that. But then why limit it to just the undocumented? I guess the other "nexus" test we had was already doing a content test on NXmpes.

I would say we can, instead, make an example Template and NXDL, not related to any technique, where some fields in the Template are not documented in the accompanying NXDL. We can use the NXtest.nxdl.xml here. Then we test for that extra added list of fields. So we know this tool is correctly identifying undocumented fields.

Am I missing something here? Probably, you have a better idea on why such a test is necessary.

domna:
> I just wanted to know what exactly are we trying to achieve with the regression test here. Do we need to lock the content in? It seems we are trying to ensure consistency in what fields we expect to be undocumented in our NXDL. But this doesn't really test the functionality of the tool to figure out whether these fields are the ones we want to show as warning.

Yes, the background to the test is that we want to ensure consistency of the warnings we see in the mpes example. Lately, we had the problems that the changes to the base classes vanished and we got new undocumented warnings in the mpes example, which we only found as soon as we ran the examples again.


> What this test really is checking for is whether our readers/NXDL still have the same undocumented combo as we had before. All I see this doing is adding another step to correct before we push commits. If we really want to enforce some sort of expected content from our NXDL/filled template, we can do that. But then why limit it to just the undocumented? I guess the other "nexus" test we had was already doing a content test on NXmpes.

Yes, we do have a test which generates a regression file and compares it to the current output. I agree that this already does what we want, but in the case above it was just regenerated and overwritten. This procedure is often done because people don't really see what has changed and with lots of tiny changes to other base classes at the NIAC pullback it made sense to just regenerate the regression file. 
However, this changes here are just an additional layer that we don't introduce any unforeseen changes to the examples. I agree that it is incomplete (because there could be fields missing which are not in the example) and redundant (because we have the log output file), but I think its still helpful in cases where the undocumented fields change to have an easy comparison. Also we can use it for other readers to ensure example consistency.

If we want to invest a little more time in this we could instead check the full template we generate for the example and check whether this is what we expect (including what is undocumented etc.). I just wanted to add this as a quick check so we don't run into the problems we had with the base class changes again.

> I would say we can, instead, make an example Template and NXDL, not related to any technique, where some fields in the Template are not documented in the accompanying NXDL. We can use the NXtest.nxdl.xml here. Then we test for that extra added list of fields. So we know this tool is correctly identifying undocumented fields.

That's not really what I want, because I want to specifically check the examples output.

sherjeelshabih (approved):

rettigl:
Let me also put my 5cents here, as I was the one asking for this kind of test. We already a couple of times had the situation that things were changed on either the code or the definitions that lead to warnings popping up during conversions in the examples, which however nobody noticed, because nobody of us actually ever runs these examples, and as mentioned above the typical practice with the existing tests is that the thousans-line-long regression logs are just overwritten when a new version comes, and be done with, and nobody notices if e.g. a nexus field disappears. 
Thus, I wanted a test that explicitly tests that the output of the conversion conforms to an expected output, which typically would be that no warnings pop up.
Indeed, even the undocumented fields you included now are a product of such a not-detected change in the NXdata base class, which should have been corrected in our config files long ago, but was not, because nobody noticed... => exactly why we need this test.

rettigl:
As for the implementation, I would even really do this at the level of captureing the output of the conversion process (either command line or python function), and comparing it to a (very short) regression log. This way, you also catch output from the reader itself, and any output from the writer that might change...

domna:
> Indeed, even the undocumented fields you included now are a product of such a not-detected change in the NXdata base class, which should have been corrected in our config files long ago, but was not, because nobody noticed... => exactly why we need this test.

I corrected the config_file (thanks for pointing it out!)
Surprisingly, these fields are not found by the regression log (I did not change it now and did not need to update the log output). So such changes seem to even go unnoticed by the log regression test.

> As for the implementation, I would even really do this at the level of captureing the output of the conversion process (either command line or python function), and comparing it to a (very short) regression log. This way, you also catch output from the reader itself, and any output from the writer that might change...

There were several reasons I did it like this. 

- First one is that this test is now more specific and cleaner on what we actually test. So it will not go off when we, e.g., change the info text of the reader which would be confusing. 
- Second one is that we do not need an additional file someone needs to update and compare, but we can do it directly in an array. My feeling is that people changing this are more aware of what is actually changing, e.g., actively adding undocumented fields may give more the feeling of changing something unwanted then simply updating a file - but maybe it's not a big problem because the regression file would be very small. 
- The last one is not a real valid argument but I initially tried to simply add it in the `test_has_correct_read_function` but the some of the logs did not show up and I just let it be because I liked the current option better. However, I do have the feeling that file based regression testing is not too reliable, also with respect to the problems we sometimes have with the other regression log files (sometimes we get scientific notation for numeric fields, sometimes just plain zeros and we had problems with local vs. ci/cd compatibility in the past). Again, this probably is no big problem for the small test file but I still have the feeling that avoiding files is the better choice here.

sherjeelshabih:
Thank you for the explanation. I get the motivation for this test. I think this is a good implementation. The regular reader test does check for paths that are documented. So if there is anything missing that test will fail. And now this covers the willfully ignored undocumented paths.

I will still recommend having a detached to technique test for the functionality that spits out undocumented fields. But it's not necessarily in the scope of this PR.

domna:
> I will still recommend having a detached to technique test for the functionality that spits out undocumented fields. But it's not necessarily in the scope of this PR.

Yes, I agree we should add such a test, I created #156 for this.

domna:
@rettigl I will merge this now. If you still prefer the file log based approach I can add this in a separate PR.

PR #157, opened 2023-09-12
Adds passing of `objects` to transmission reader

domna:
This adds the functionality of passing a dict to the transmission readers convert function.

@hampusnasstrom


sherjeelshabih:
So you pass Dict type objects and append them to the template? No other object types are accepted currently, right?

sherjeelshabih (approved):

domna:
> So you pass Dict type objects and append them to the template? No other object types are accepted currently, right?

Exactly, I expect an iterable of dicts. Every other element is simply warned about and ignored

Issue #145, opened 2023-08-07
Attributes are written without the actual fields


domna:
If the internal templates gets some attribute keys, e.g., `/my/key/@units`, without the actual `/my/key` it writes this attribute without the actual key.
The reader should check whether the field is present before actually writing any units.

@Arora0 @rettigl

lukaspie:
Solved with new validation

PR #159, opened 2023-09-13
Adds test for group optionality

sherjeelshabih:
I've restructured the validation of groups nested in optional groups, etc. This can still use more example cases to test against. I'll leave this open for further work and discussion into this.

Currently, we are testing the following cases:

- optional group with required field missing
- optional group with no child provided
- empty required group
- empty required group under an optional group

If there is anything more that can be added, please let me know.

### DIFF
--- a/pynxtools/dataconverter/helpers.py
+++ b/pynxtools/dataconverter/helpers.py
@@ -424,7 +424,7 @@ def does_group_exist(path_to_group, data):
     return False
 
 
-def ensure_all_required_fields_exist(template, data):
+def ensure_all_required_fields_exist(template, data, nxdl_root):
     """Checks whether all the required fields are in the returned data object."""
     for path in template["required"]:
         entry_name = get_name_from_data_dict_entry(path[path.rindex('/') + 1:])
@@ -432,7 +432,11 @@ def ensure_all_required_fields_exist(template, data):
             continue
         nxdl_path = convert_data_converter_dict_to_nxdl_path(path)
         is_path_in_data_dict, renamed_path = path_in_data_dict(nxdl_path, data)
+
         if path in template["lone_groups"] and does_group_exist(path, data):
+            continue
+
+        if check_for_optional_parent(path, nxdl_root) != "<<NOT_FOUND>>":
             continue
 
         if not is_path_in_data_dict or data[renamed_path] is None:
@@ -475,7 +479,7 @@ def validate_data_dict(template, data, nxdl_root: ET.Element):
     nxdl_path_to_elm: dict = {}
 
     # Make sure all required fields exist.
-    ensure_all_required_fields_exist(template, data)
+    ensure_all_required_fields_exist(template, data, nxdl_root)
     try_undocumented(data, nxdl_root)
 
     for path in data.get_documented().keys():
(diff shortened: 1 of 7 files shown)


pynxtools/dataconverter/helpers.py
```diff
@@ -424,17 +426,25 @@ def does_group_exist(path_to_group, data):
...
-        if path in template["lone_groups"] and does_group_exist(path, data):
-            continue
 
+        renamed_path = path if renamed_path is None else renamed_path
+        if path in template["lone_groups"]:
+            opt_parent = check_for_optional_parent(path, nxdl_root)
+            if not does_group_exist(renamed_path, data):
+                raise ValueError(f"The required group, {path}, hasn't been supplied.")
+            if opt_parent != "<<NOT_FOUND>>" and does_group_exist(opt_parent, data):
+                if not does_group_exist(renamed_path, data):
```
domna, line 444: Will this not always be catched by the first if-clause already?

tests/data/dataconverter/readers/json_map/data.mapping.json
```diff
@@ -18,5 +18,6 @@
     "/ENTRY[entry]/optional_parent/required_child": "/required_child",
     "/ENTRY[entry]/program_name": "Example for listing exact data in the map file: Nexus Parser",
     "/ENTRY[entry]/required_group/description": "An example description",
-    "/ENTRY[entry]/required_group2/description": "An example description"
```
domna, line 21: I think it would also help to have another test without specifying the path `/ENTRY[entry]/optional_parent` to see whether it works to leave out the `optional_parent` group entirely.

pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -24,7 +24,7 @@
 
 from pynxtools.dataconverter.readers.base.reader import BaseReader
 from pynxtools.dataconverter.template import Template
-from pynxtools.dataconverter.helpers import ensure_all_required_fields_exist
+# from pynxtools.dataconverter.helpers import ensure_all_required_fields_exist
```
domna, line 27:
```suggestion
```

domna (changes requested):
Looks good, but I think it would be helpful to test also when the entire optional group is not present. Do you think this would also work for #145? Currently, I think we just stop it because we continue if the path ends on `@units`.

pynxtools/dataconverter/helpers.py
sherjeelshabih: Yes. Thanks. I fixed it now. :pray:

tests/data/dataconverter/readers/json_map/data.mapping.json
sherjeelshabih: I added this test as well. It was quite needed. It works i guess.

tests/dataconverter/test_helpers.py
```diff
@@ -269,12 +295,13 @@ def test_validate_data_dict(data_dict, error_message, template, nxdl_root, reque
...
                                     "link-dict-instead-of-bool",
-                                    "allow-required-and-empty-group"):
+                                    "allow-required-and-empty-group",
+                                    "opt-group-completely-removed"):
         helpers.validate_data_dict(template, data_dict, nxdl_root)
     else:
         with pytest.raises(Exception) as execinfo:
             helpers.validate_data_dict(template, data_dict, nxdl_root)
-
+        print(execinfo.value)
```
domna, line 304: Is this planned or just a residual from debugging?

```suggestion
```

domna (approved):
Thanks for the changes.
Looks good now.

tests/dataconverter/test_helpers.py
sherjeelshabih: Ah. Residual ofc...

PR #175, opened 2023-10-31
Acumulating errors for groups, fields and attributes.

RubelMozumder:
- [ ] Accumulatting all the error for required fields, groups and attributes, later raise an error.


RubelMozumder:
@domna here is the primary approach. Did you mean something like this? 
With this change, the converter will not stop for individual error but rather collect all required fields, groups, and attributes that have not been supplied. Finally will raise the error.

pynxtools/dataconverter/helpers.py
```diff
@@ -456,15 +459,28 @@ def ensure_all_required_fields_exist(template, data, nxdl_root):
             opt_parent = check_for_optional_parent(path, nxdl_root)
             if opt_parent != "<<NOT_FOUND>>":
                 if does_group_exist(opt_parent, data) and not does_group_exist(renamed_path, data):
-                    raise ValueError(f"The required group, {path}, hasn't been supplied"
-                                     f" while its optional parent, {path}, is supplied.")
+                    missing_groups_and_fields['groups'].appends(path)
```
domna, line 462: I would just use logging here and do a `logger.warning`.

pynxtools/dataconverter/helpers.py
```diff
@@ -456,15 +459,28 @@ def ensure_all_required_fields_exist(template, data, nxdl_root):
...
             if not does_group_exist(renamed_path, data):
                 raise ValueError(f"The required group, {path}, hasn't been supplied.")
             continue
         if not is_path_in_data_dict or data[renamed_path] is None:
-            raise ValueError(f"The data entry corresponding to {path} is required "
-                             f"and hasn't been supplied by the reader.")
+            if entry_name.startswith('@'):
+                missing_groups_and_fields['attributes'].append(path)
+            else:
+                missing_groups_and_fields['fields'].append(path)
```
domna, line 471: Also can just be a `logger.warning`

pynxtools/dataconverter/helpers.py
```diff
@@ -456,15 +459,28 @@ def ensure_all_required_fields_exist(template, data, nxdl_root):
...
+
+    # Rasing errors for any kind of missing required fields, gorup and attributes:
+    def raise_required_concept_missing_error(missing_groups_and_fields):
+        error = ''
+        for concept, path_list in missing_groups_and_fields.items():
+            error = error + f"Required {concept} are not provided: \n"
+            if path_list:
+                for path in path_list:
+                    error = error + f"    \n\u2022{path}\n"
+        if error:
```
domna, line 481:
```suggestion
        if missing_groups_and_fields:
```

domna (commented):
This works nicely for me already. However, I would just use pythons logging to report on the fields which are missing and just raise a general warning (e.g., `Missing required fields`), maybe with a custom Error for this (so we can specifically catch this). That way we can also suppress the warnings externally if we want (e.g., when we want to write `@partial`).

I think it would also be useful to have a flag in pynxtools config to switch failing on or off here in case we just want to write a `@partial` appdef or we just catch a custom Error. When we add support for writing partials we could add a cli argument for this.

We could also add the `@partial` resolving here, but I think it is more involved, because we actually have to check which concepts we supply and it might be easier to request this via cli (i.e., we don't say `--nxdl NXmy_nxdl` but rather `--nxdl NXmy_nxdl/NXinstrument,NXmy_nxdl/NXsample`)

I think @sherjeelshabih might also have an opinion on this.

RubelMozumder:
> This works nicely for me already. However, I would just use pythons logging to report on the fields which are missing and just raise a general warning (e.g., `Missing required fields`), maybe with a custom Error for this (so we can specifically catch this). That way we can also suppress the warnings externally if we want (e.g., when we want to write `@partial`).
> 
> I think it would also be useful to have a flag in pynxtools config to switch failing on or off here in case we just want to write a `@partial` appdef or we just catch a custom Error. When we add support for writing partials we could add a cli argument for this.
> 
> We could also add the `@partial` resolving here, but I think it is more involved, because we actually have to check which concepts we supply and it might be easier to request this via cli (i.e., we don't say `--nxdl NXmy_nxdl` but rather `--nxdl NXmy_nxdl/NXinstrument,NXmy_nxdl/NXsample`)
> 
> I think @sherjeelshabih might also have an opinion on this.

I see that there are two points:
1. Having logger warning and finally raise `a general (with out nofitying any missing fields)` error to get the reader caught. 
   Still I see a pitfall in `general error massage`, the nomad loger does not catch the logger info from pynxtools. So the a general error massage would not convey any valuable info. One thing, now we can do having logger massage as you mentioned and error massage with the list of missing required field, groups, and attributes. So the error massage will also be meaningful in Noamd. Probably, @sherjeelshabih may have something to include. 
   
   Another thing, I want to mention here, we need create a general logger ogject, that can take other logger object and pass the log massages from pynxtools. But this must come through another  separate development and PR. @sherjeelshabih what do you think?
   
2. The `@partial` resolution needs more discussion and a refactor lots of code and files.
   Because, from the `template genration` point of view it is fine. But the conflicts come after manipulation of template inside reader. Because, the motive of reader to work with `application definition`, not with `base definition`. So before getting any log massage or error reader will block the process. So, this kind of modification needs a different PR with lots of changes. @sherjeelshabih what is you opinion?

pynxtools/dataconverter/helpers.py
RubelMozumder: > 

It does not work, because `missing_group_and_fields` is not an empty dict.
```ipython
In [1]: dic = {'xx':[]}

In [2]: if dic:
   ...:     print('somthing')
   ...: 
somthing
```

pynxtools/dataconverter/helpers.py
RubelMozumder: Agree!

pynxtools/dataconverter/helpers.py
RubelMozumder: So, you do not want the converter blocked, even the required fields are not rendered.

sherjeelshabih: This is what I had in mind. No need to collect all these errors like this. We can just output the warnings.

sherjeelshabih:
I would say once we have the warning messages, let's not even throw an error. We want to be as permissive as possible now. And in another PR we will automatically add the @partial attribute when we detect something missing. So with those two, I guess we won't need to raise anything. 

@RubelMozumder we should definitely check how to get the logger right from Nomad. I thought this was already being done. It will be best to not raise any errors and just let the Nomad GUI display warnings for missing fields. The workflow shouldn't be stopped like that.

domna:
> I would say once we have the warning messages, let's not even throw an error. We want to be as permissive as possible now. And in another PR we will automatically add the @partial attribute when we detect something missing. So with those two, I guess we won't need to raise anything.
> 
Did not fully read all the comments yet but just wanted to leave my two cents on this: This is why I suggested using a specific error message for this. That way we could also catch it in nomad and just ignore it, but we could also fail in pynxtools if you want to fail on this (like a non-partial mode, e.g., I want to reliably write a full appdef). But I would be fine with non-failing by default (also that way we can directly throw it into nomad without any modification).

RubelMozumder:
> I would say once we have the warning messages, let's not even throw an error. We want to be as permissive as possible now. And in another PR we will automatically add the @partial attribute when we detect something missing. So with those two, I guess we won't need to raise anything.
> 
> @RubelMozumder we should definitely check how to get the logger right from Nomad. I thought this was already being done. It will be best to not raise any errors and just let the Nomad GUI display warnings for missing fields. The workflow shouldn't be stopped like that.

pynxtools/dataconverter/convert.py
```diff
@@ -72,8 +72,12 @@ def convert(input_file: Tuple[str],
             generate_template: bool = False,
             fair: bool = False,
             undocumented: bool = False,
+            logger_: logging.Logger = None,
```
domna, line 75: The best practice for logging is different. The docs recommend to use a modul level logger with a name, e.g., creating `logger` variable at the start of the file (see https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial):

```python
import logging

logger = logging.getLogger(__file__)
```

Then you can just use `logger.warning("My warning")` inside the file. Also use `%s` syntax when formatting the string, e.g., `logger.warning("Invalid value: %s", my_var)` when logging my_var, because logging does not evaluate my_var in case the logging level is disabled.

This allows another tool using pynxtools to hook into the logger with our name and display the emitted logs. I don't know if we have to follow naming conventions for nomad (I had this for uvicorn where I needed to create a name like `logging.getLogger(f'uvicorn.{__file__}')` and then everything was displayed automagically).

RubelMozumder: I will try to follow it.

RubelMozumder:
nomad_logger

So, there is no blocker, pynxtools reader is now passing the missing required concepts with warning.

pynxtools/dataconverter/convert.py
RubelMozumder: Yah, when you mentioned about `logger` warning in this, I also thought to have something like this as you mentioned here. I see the beauty of logger.

RubelMozumder: One thing, `logger.warning("Invalid value: %s", my_var)` does not replace %s when in nomad logger. But the same syntax works in console when calling reader.

RubelMozumder: Module level logger in pynxtools is also working nicely.

domna: > One thing, `logger.warning("Invalid value: %s", my_var)` does not replace %s when in nomad logger. But the same syntax works in console when calling reader.

Ok, this is odd. Normally, this is recommended (see https://docs.python.org/3/howto/logging.html#optimization). Maybe this behaviour is deactivated for some reason in nomad, because there I just see f-strings being used (and the pylint errors are also deactivated). So we can just follow the nomad way: use f-strings and add `logging-not-lazy`, `logging-format-interpolation`, `logging-fstring-interpolation` to `.pylintrc`. But I would be interested why it is like this in nomad

pynxtools/dataconverter/convert.py
```diff
@@ -34,11 +34,7 @@
 from pynxtools.dataconverter.template import Template
 from pynxtools.nexus import nexus
 
-
-logger = logging.getLogger(__name__)  # pylint: disable=C0103
-UNDOCUMENTED = 9
-logger.setLevel(logging.INFO)
-logger.addHandler(logging.StreamHandler(sys.stdout))
+from .logger import logger
```
sherjeelshabih, line 37: We should use absolute imports like we do in the rest of the project. The only upside is we can move this file anywhere and it will still work with an absolute import. Also it looks cleaner because we already do it everywhere else.

pynxtools/dataconverter/readers/xrd/README.md
```diff
@@ -0,0 +1,40 @@
+# XRD reader
```
domna, line 1: Why is this part of this PR?
Please remove the xrd part and make it separate PR if necessary.

dev-requirements.txt
```diff
@@ -510,6 +510,7 @@ typing-extensions==4.3.0
     #   astroid
     #   mypy
     #   numcodecs
+    #   pylint
```
domna, line 513: This changes can be reverted, because it is not changing anything.

pynxtools/dataconverter/convert.py
```diff
@@ -34,18 +34,14 @@
...
 if sys.version_info >= (3, 10):
     from importlib.metadata import entry_points
 else:
     from importlib_metadata import entry_points
 
 
-logger = logging.getLogger(__name__)  # pylint: disable=C0103
-UNDOCUMENTED = 9
-logger.setLevel(logging.INFO)
-logger.addHandler(logging.StreamHandler(sys.stdout))
```
domna, line 46: Why has the `UNDOCUMENTED` level been removed?
I introduced to be able to separate info and undocumented loggings.

pynxtools/dataconverter/logger.py
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import logging
+
+logger = logging.getLogger("pynxtools")
```
domna, line 22: Is there any benefit of having a central `logger` instance?

pynxtools/dataconverter/helpers.py
```diff
@@ -32,7 +33,7 @@
 from pynxtools.nexus import nexus
 from pynxtools.nexus.nexus import NxdlAttributeError
 
-logger = logging.getLogger(__name__)
+pynx_logger = logging.getLogger(__name__)
```
domna, line 36: This is inconsistent. In other places you use the central logger and here you use a local logger.

pynxtools/dataconverter/helpers.py
```diff
@@ -446,7 +447,9 @@ def does_group_exist(path_to_group, data):
     return False
 
 
-def ensure_all_required_fields_exist(template, data, nxdl_root):
+# pylint: disable=W1203
+def ensure_all_required_fields_exist(template, data,
+                                     nxdl_root, logger=pynx_logger):
```
domna, line 452: Normally, it shouldn't be necessary to pass the logger here. `logging` keeps track of the available loggers and integrates them

pynxtools/dataconverter/helpers.py
```diff
@@ -497,7 +500,9 @@ def try_undocumented(data, nxdl_root: ET.Element):
             pass
 
 
-def validate_data_dict(template, data, nxdl_root: ET.Element):
+def validate_data_dict(template, data,
+                       nxdl_root: ET.Element,
+                       logger=pynx_logger):
```
domna, line 505: Passing of logging shouldn't also be required here.

pynxtools/dataconverter/helpers.py
```diff
@@ -589,7 +594,7 @@ def convert_to_hill(atoms_typ):
     return atom_list + list(atoms_typ)
 
 
-def add_default_root_attributes(data, filename):
+def add_default_root_attributes(data, filename, logger=pynx_logger):
```
domna, line 597: Also not required to be passed here.

pynxtools/nexus/nexus.py
```diff
@@ -639,19 +643,28 @@ def full_visit(self, root, hdf_node, name, func):
...
+            if self.d_inq_nd is None and self.c_inq_nd is None:
+                get_default_plotable(self.in_file, self.logger)
+            # To log the provided concept and concepts founded
+            if self.c_inq_nd is not None:
+                for hdf_path in self.hdf_path_list_for_c_inq_nd:
+                    self.logger.info(hdf_path)
+        finally:
+            # To test if hdf_file is open print(self.in_file.id.valid)
+            self.in_file.close()
+            # To test if hdf_file is open print(self.in_file.id.valid)
```
domna, line 667: This seems also not related to the logging

pynxtools/dataconverter/template.py
```diff
@@ -125,6 +125,27 @@ def __contains__(self, k):
...
+        val = self.optional.get(key, None)
+        if val is None:
+            val = self.recommended.get(key, None)
+            if val is None:
+                val = self.required.get(key, None)
+                if val is None:
+                    val = self.undocumented.get(key, None)
+        if val is None:
+            return return_value
+        return val
```
domna, line 147: This is also not related to the logging parts

pynxtools/dataconverter/helpers.py
```diff
@@ -650,3 +655,72 @@ def extract_atom_types(formula, mode='hill'):
...
+                                        parent_path=''):
+    """Convert nested dict into slash separeted path upto certain level."""
+    sep = '/'
+
+    for key, val in nested_dict.items():
+        path = parent_path + sep + key
+        if isinstance(val, dict):
+            nested_dict_to_slash_separated_path(val, flattened_dict, path)
+        else:
+            flattened_dict[path] = val
```
domna, line 726: This seems to be not related to the logging feature.

domna (changes requested):
In this PR are a lot of additional features. Please cleanup this PR to only add the features relevant for proper logging of the writing errors.

pynxtools/dataconverter/helpers.py
RubelMozumder: That's came from rebase. I will fixed
 it

RubelMozumder:
This PR is replaced by PR #191 .

Issue #165, opened 2023-10-24
Merging of multiple files


domna:
When we have multiple nexus files with an @partial attribute, they should be merged into one file according to an appdef (or report which parts are missing).

PR #176, opened 2023-11-01
Merging of multiple files

sherjeelshabih:
I have broken down an NXiv_temp NeXus file.


sherjeelshabih:
- [ ] Add merge and map as CLI options piping those calls through the JSON Map Reader
- [ ] Add docs for the new options

pynxtools/dataconverter/hdfdict.py
```diff
@@ -123,7 +123,12 @@ def _recurse(hdfobject, datadict):
             elif isinstance(value, h5py.Dataset):
                 if not lazy:
                     value = unpacker(value)
-                datadict[key] = value
+                datadict[key] = value.asstr()[...] if h5py.check_string_dtype(value.dtype) else value  # pylint: disable=line-too-long
```
domna, line 126:
```suggestion
                datadict[key] = (
                    value.asstr()[...]
                    if h5py.check_string_dtype(value.dtype)
                    else value
                )
```

examples/json_map/README.md
```diff
@@ -0,0 +1,25 @@
...
+This reader is designed to allow users of pynxtools to convert their existing data with the help of a map file. The map file tells the reader what to pick from your data files and convert them to FAIR NeXus files. The following formats are supported as input files:
+* HDF5 (any extension works i.e. h5, hdf5, nxs, etc)
+* JSON
+* Python Dict Objects Pickled with [pickle](https://docs.python.org/3/library/pickle.html). These can contain [xarray.DataArray](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html) objects as well as regular Python types and Numpy types.
+
+## How to run these examples?
+
+### Automatically merge partial NeXus files
+```console
+user@box:~$ dataconverter --nxdl NXiv_temp --input-file voltage_and_temperature.nxs --input-file current.nxs --merge-partial --output auto_merged.nxs
```
domna, line 14: Could this not be the default route when no reader is specified?

examples/json_map/merge_copied.mapping.json
```diff
@@ -0,0 +1,35 @@
...
+    "/ENTRY[entry]/DATA[data]/current": "/entry/data/current",
+    "/ENTRY[entry]/DATA[data]/current_295C": "/entry/data/current_295C",
+    "/ENTRY[entry]/DATA[data]/current_300C": "/entry/data/current_300C",
+    "/ENTRY[entry]/DATA[data]/current_305C": "/entry/data/current_305C",
+    "/ENTRY[entry]/DATA[data]/current_310C": "/entry/data/current_310C",
+    "/ENTRY[entry]/DATA[data]/temperature": "/entry/data/temperature",
+    "/ENTRY[entry]/DATA[data]/voltage": "/entry/data/voltage",
+    "/ENTRY[entry]/INSTRUMENT[instrument]/ENVIRONMENT[environment]/voltage_controller/calibration_time": "/entry/instrument/environment/voltage_controller/calibration_time",
+    "/ENTRY[entry]/INSTRUMENT[instrument]/ENVIRONMENT[environment]/voltage_controller/run_control": "/entry/instrument/environment/voltage_controller/run_control",
+    "/ENTRY[entry]/INSTRUMENT[instrument]/ENVIRONMENT[environment]/voltage_controller/value": "/entry/instrument/environment/voltage_controller/value",
```
domna, line 12: Would it be possible to add group copying?
This would mean writing
```json
"/ENTRY[entry]/INSTRUMENT[instrument]/ENVIRONMENT[environment]/voltage_controller": "/entry/instrument/environment/voltage_controller",
```
here. Or did we just wanted to add group copying later?

domna: I would also find the linking syntax here helpful. So we always specify which file this comes from (or at least are able to do it) for cases where we might have duplicate fields.

This is what I'm thinking of:
```json
"/ENTRY[entry]/INSTRUMENT[instrument]/ENVIRONMENT[environment]/voltage_controller/run_control": 
"voltage_and_temperature.nxs:/entry/instrument/environment/voltage_controller/run_control",
```

pynxtools/dataconverter/README.md
```diff
@@ -37,9 +37,28 @@ Options:
                                   parameters the converter supports.
   --undocumented                  Shows a log output for all undocumented
                                   fields
+  --merge-partial                 Merges partial NeXus files provided as
+                                  different --input-file to the converter.
+  --mapping TEXT                  Takes a <name>.mapping.json file and
+                                  converts data from given input files.
```
domna, line 43: Does it *have* to be a `.mapping.json` file or would the reader also take a `.json` file? In the latter case I would just say json file but add a sentence that the convention is `.mapping.json`.

domna (commented):
Works nicely. Except from a view things this looks nice already. I have two requests though (which we might also move to separate issues if they are not already) which might be quick to do for you:

- Do we also want to add writing `@partial` files in this PR? I think this would also be a needed feature, but I think we don't have a separate issue for this yet.
- Could you write a small tutorial/how to for this feature? Just an mkdocs md file and then we can integrate it where we need it (nomad, future pynxtools mkdocs?). It doesn't need to be supper eloquent but I think it's good to have a basis as this will be very relevant for the technology partners. https://github.com/FAIRmat-NFDI/nexus_definitions/issues/110

pynxtools/dataconverter/README.md
sherjeelshabih: It has to be a .mapping.json. A regular json file can be used as a data file. So should I keep it as it is or you recommend a better phrasing?

examples/json_map/README.md
sherjeelshabih: We can have it as the default. I just didn't wanna take big steps like that without discussion. There isn't any other generic reader that could benefit from a default, right? So I can set this as the default.

pynxtools/dataconverter/README.md
domna: No then the phrasing is fine.

examples/json_map/merge_copied.mapping.json
domna: ok, sounds good

examples/json_map/README.md
domna: Yeah, now I'm not fully sure about it. Could be that users might just have forgotten to specify the reader, so if we use it as default we might show an additional info text that the partial merging was used (even though the reader could figure it out from the source file it got: nxs files -> merging; any other file -> probably user forgot to specify the reader. but I don't know how cumbersome it is to implement this)

sherjeelshabih: It's done. Commit coming up in a minute.

domna (approved):
Thanks for the additional docs. The adding of the `@partial` should be handled in #175

lukaspie:
Sorry for the delay. I just tested all the new features and didn't find any bugs. I agree that group copying might become an important feature. As an example if I have an instrument in Nomad which complies to a certain nxdl version and I also want to use this in another experiment, I could just copy all the static instrument groups (e.g. NXtransformations) over, without needing to map every single detail. But, this is out of the scope of this PR, so let's add it later.

sherjeelshabih:
Fixes #165

PR #177, opened 2023-11-10
Python 3.11 support

domna:
I think it would be good to start extending the support matrix to newer versions.


lukaspie:
Currently, [ifes_apt_tc_data_modeling](https://github.com/atomprobe-tc/ifes_apt_tc_data_modeling) is blocking the tests because [v0.0.9](https://pypi.org/project/ifes-apt-tc-data-modeling/) requires Python <3.11, >=3.8. @mkuehbach could you check if newer python versions could also work for this tool?

mkuehbach:
Hi @lukaspie, in principle that lib should also work for 3.11, at the point when the v0.0.9 was written that was not tested though and thus I switched py311 off. As I was planning to start on Monday anyways on implementing the ATO and APT parser for the FAU Erlangen examples, I would then test this with 3.11 and 3.12 and bump the ifes lib accordingly then

pyproject.toml
```diff
@@ -12,11 +12,12 @@ authors = [
 description = "Extend NeXus for materials science experiment and serve as a NOMAD parser implementation for NeXus."
```
mkuehbach, line 12: Extend NeXus for experiments and characterization in Materials Science and Materials Engineering ...

mkuehbach (approved):

pyproject.toml
domna:
```suggestion
description = "Extend NeXus for experiments and characterization in Materials Science and Materials Engineering and serve as a NOMAD parser implementation for NeXus."
```

PR #178, opened 2023-11-19
Fix mpes data indices

domna:
The data indices in NXdata should be int and not char. This fixes it.

### DIFF
--- a/pynxtools/dataconverter/readers/mpes/reader.py
+++ b/pynxtools/dataconverter/readers/mpes/reader.py
@@ -199,20 +199,6 @@ def handle_h5_and_json_file(file_paths, objects):
             )
 
         if not os.path.exists(file_path):
-            file_path = os.path.join(
-                os.path.dirname(__file__),
-                "..",
-                "..",
-                "..",
-                "..",
-                "tests",
-                "data",
-                "dataconverter",
-                "readers",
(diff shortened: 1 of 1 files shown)


pynxtools/dataconverter/readers/mpes/reader.py
```diff
@@ -198,20 +198,6 @@ def handle_h5_and_json_file(file_paths, objects):
                 f"but {file_path} does not match.",
             )
 
-        if not os.path.exists(file_path):
```
Arora0, line 201: If this if statement is removed then one needs to mention the entire data path which can be quite long.

Arora0 (approved):

pynxtools/dataconverter/readers/mpes/reader.py
domna: Yes, I agree that this may require to write out the full path for the testing files. However, I think it is odd for users when they get an error about some testing data path when they used a wrong file path which actually happened multiple times to me. Additionally, I think for the test files this is covered in the tests anyways and one should just run the tests to check to files. That way it isn't necessary to type the full path, too.

Arora0: Makes sense as it only effects the test data.

PR #182, opened 2023-11-21
Mpes automate data indices

Arora0:
As per an earlier discussion, this replaces the data indices in the config with an "*" and fills the config with the indices from the xarray before filling the template dictionary.

### DIFF
--- a/pynxtools/dataconverter/readers/mpes/reader.py
+++ b/pynxtools/dataconverter/readers/mpes/reader.py
@@ -243,6 +243,25 @@ def rgetattr(obj, attr):
     return reduce(_getattr, [obj] + attr.split("."))
 
 
+def fill_data_indices_in_config(config_file_dict, x_array_loaded):
+    """Add data indices key value pairs to the config_file
+    dictionary from the xarray dimensions if not already
+    present.
+    """
+    for key in list(config_file_dict):
+        if "*" in key:
+            value = config_file_dict[key]
+            for dim in x_array_loaded.dims:
+                new_key = key.replace("*", dim)
+                new_value = value.replace("*", dim)
+
+                if new_key not in config_file_dict.keys() \
+                        and new_value not in config_file_dict.values():
+                    config_file_dict[new_key] = new_value
+
+            config_file_dict.pop(key)
+
+
 class MPESReader(BaseReader):
     """MPES-specific reader class"""
 
@@ -251,7 +270,7 @@ class MPESReader(BaseReader):
     # Whitelist for the NXDLs that the reader supports and can process
     supported_nxdls = ["NXmpes"]
 
-    def read(
+    def read(  # pylint: disable=too-many-branches
             self,
             template: dict = None,
             file_paths: Tuple[str] = None,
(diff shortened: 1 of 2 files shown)


pynxtools/dataconverter/readers/mpes/reader.py
```diff
@@ -269,6 +269,13 @@ def read(
             eln_data_dict,
         ) = handle_h5_and_json_file(file_paths, objects)
 
+        for key in list(config_file_dict):
+            if "*" in key:
+                value = config_file_dict[key]
+                for dim in x_array_loaded.dims:
+                    config_file_dict[key.replace("*", dim)] = value.replace("*", dim)
```
domna, line 276: What happens if this key is already present?
Is it fine if this is overwritten like it currently is or should it be treated differently? Should an info or warning be logged in this case?

IMO it would be the cleanest that `*` acts as the default if there is no explicit key/value pair. If there is an explicit key value pair this should be taken. In that way you could always overwrite a certain axis from somewhere else if necessary.

Like so:

```
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@*_indices]": "@data:*.index",
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@kx_indices]": "5",
```

this ends updated dict
```
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@ky_indices]": "@data:*ky.index",
...
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@kx_indices]": "5",
```

instead of 
```
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@kx_indices]": "@data:*kx.index",
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@ky_indices]": "@data:*ky.index",
...
```

domna (approved):
Looks good. It's already approved. If you decide to change the overwriting behaviour I can re-review it.

rettigl (approved):
Apart from the overwriting issue, LGTM.

rettigl:
For the linting error add an ignore comment, as discussed.

pynxtools/dataconverter/readers/mpes/reader.py
Arora0: Done. I also added another condition which allows the indices to be renamed in the config. But for this to work, one has to also manually change the @axes attribute to a list containing the new indices instead of "@data:dims" as it is normally.
For example, to rename "energy" to "E", you do: 
```
"/ENTRY[entry]/DATA[data]/@axes": ["kx", "ky", "E", "delay"],
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@*_indices]": "@data:*.index",
"/ENTRY[entry]/DATA[data]/AXISNAME_indices[@E_indices]": "@data:energy.index"
...
```
and so on for "axisname" field as well.

rettigl (approved):
LGTM

domna (approved):
Looks good, thanks!

rettigl:
@domna Shall we merge this, and make a new release? I would like to integrate this into the nomad containers soon, if possible.

domna:
Sure, this should be possible. The mpes analysis containers are quickly updatable. Into nomad itself it would take a moment because we would need to merge the submodule there, too.

PR #184, opened 2023-11-24
Fix build via `python -m build`

domna:
This fixes the build via the python build method for wheel and sdist generation. It fails if the `nexus-version.txt` is not already present, so the code should just return to not write the file again if the code is not in a github repository anymore.
Previously, this was unnoticed because it works when the file was generated before.


domna:
@rettigl The package generation currently fails (see https://github.com/FAIRmat-NFDI/pynxtools/actions/runs/6982526131/job/19001851868). This PR fixes it and afterwards I can publish the package and re-generate the package.

sherjeelshabih (approved):

PR #185, opened 2023-11-28
Allow loading of plugins via entry_points

domna:
The reader can basically refactored out as is, we just have to add a `pyproject.toml` with the following entrypoint specification:

```
[project.entry-points."pynxtools.reader"]
test = "test_reader:MyReaderClass"
```

where `test` is the name of the reader referred by `--reader` and `MyReaderClass` is the reader class in the file.

A working example can be found in https://github.com/FAIRmat-NFDI/pynxtools-stm for the sts/stm reader.


sherjeelshabih:
This works really nicely. I have tested the stm reader and it gets loaded correctly. 

As we discussed, the next thing would be to have a testing strategy.

Also, we will need a list of reader plugins in this repo to help users find them.

domna:
We could build a [cookiecutter](https://github.com/cookiecutter/cookiecutter) template to get people started with developing their own readers

PR #186, opened 2023-11-30
Fix nexus version in published package

domna:
The nexus version cannot be found in the published package, because we only shallowly checkout the repo.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,6 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


sherjeelshabih (approved):

.github/workflows/publish.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          fetch-depth: 0
```
sanbrock, line 22: do we want to checkout different branches? tht is why we need this?

PR #190, opened 2023-12-02
Adds a get method to Dataconverter Template

sherjeelshabih:


pynxtools/dataconverter/template.py
```diff
@@ -125,6 +125,13 @@ def __contains__(self, k):
             k in self.required
         ])
 
+    def get(self, key: str, default=None):
+        """Proxies the get function to our internal __getitem__"""
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
```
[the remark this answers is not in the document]
sherjeelshabih: No, Rubel. You got it wrong. It only calls get_optionality() when the if statements are not satisfied. When a key isn't found it does raise a KeyError. Have another look. Try a template.get("/key/doesn't/exist", "somekey") and you'll get the right results.

The last call to get_optionality is a way to conveniently access your inner dicts from template based on optionality. Template only has keys starting with "/".

Also, on slack you just mentioned that you want to keep the duplicate code you added instead of using this fix we did together on Wednesday.

pynxtools/dataconverter/template.py
```diff
@@ -125,6 +125,13 @@ def __contains__(self, k):
...
     def __getitem__(self, k):
```
RubelMozumder, line 135: @sherjeelshabih I tried it out on the day before yesterday.
Here I added the code output:
```ipython
template = Template()
In [18]: generate_template_from_nxdl(nxdl_root, template)

In [19]: template['hl']
Out[19]: 
{'/ENTRY[entry]/title': None,
 '/ENTRY[entry]/definition': None,
 '/ENTRY[entry]/start_time': None,
 '/ENTRY[entry]/definition/@version': None,
 '/ENTRY[entry]/USER[user]/name': None,
 '/ENTRY[entry]/USER[user]/email': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/energy_resolution': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/energy_resolution/@units': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/SOURCE[source]/type': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/SOURCE[source]/probe': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/SOURCE[source]/name': None,
 '/ENTRY[entry]/INSTRUMENT[instrument]/BEAM[beam]/distance': None,
 ```
template['hl'] expected to raise an ValueError.

Comment (RubelMozumder) [reply pynxtools/dataconverter/template.py:135]: Yes, it is not expected to return a value for a key that is not even exist. so it should return an error, KeyError not an irrelevant dict.

File: pynxtools/dataconverter/template.py [hunk 3]
```diff
@@ -141,7 +148,10 @@ def __getitem__(self, k):
                         return self.required[k]
                     except KeyError:
                         return self.undocumented[k]
-        return self.get_optionality(k)
+        if k in ("required", "optional", "recommended", "undocumented"):
+            return self.get_optionality(k)
+        raise KeyError("Only paths starting with '/' or one of [optional_parents, "
+                       "lone_groups, required, optional, recommended, undocumented] can be used.")
```
RubelMozumder, line 154: `get` method is working perfectly.

RubelMozumder (approved):

PR #195, opened 2023-12-04
Always install importlib-metadata (not only for <3.10) and provide dummy entry_points

domna:
The version dependent install of the backporting `importlib-metadata` sparked some problems with nomad's dependency resolving. Here, I bring that `importlib-metadata` is always installed, regardless of the python version.

It also provides a dummy `entry_points` if importlib-metadata cannot be imported.

### DIFF
--- a/pynxtools/dataconverter/convert.py
+++ b/pynxtools/dataconverter/convert.py
@@ -42,7 +42,8 @@ else:
     except ImportError:
         # If importlib_metadata is not present
         # we provide a dummy function just returning an empty list.
-        def entry_points(_):
+        # pylint: disable=W0613
+        def entry_points(group, _):
             """Dummy function for importlib_metadata"""
             return []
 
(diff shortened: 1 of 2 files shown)


RubelMozumder:
**The Nomad CI/CD pipeline is disgusting.** 
Now, Nomad `CI/CD` is running properly.

pynxtools/dataconverter/convert.py
```diff
@@ -37,7 +37,15 @@
...
     from importlib.metadata import entry_points
 else:
-    from importlib_metadata import entry_points
+    try:
+        from importlib_metadata import entry_points
+    except ImportError:
+        # If importlib_metadata is not present
+        # we provide a dummy function just returning an empty list.
+        # pylint: disable=W0613
+        def entry_points(group):
```
RubelMozumder, line 46: Here I did some minor modification so that, nomad CI/CD pipeline pass the all test. No issue in nomad.

RubelMozumder (approved):

PR #201, opened 2023-12-07
Reformatted files + pre-commit hook

domna:


README.md
```diff
@@ -62,6 +62,14 @@ python -m pip install -e .
 python -m pip install -e ".[dev]"
 ```
 
+There is also a [pre-commit hook](https://pre-commit.com/#intro) available
+which formats the code and checks the linting before actually commiting.
+It can be installed with
+```shell
```
lukaspie, line 68: You should mention here that pre-commit needs to be pip installed.

sherjeelshabih: It is in the requirements file already. We will need to mention on scrum to not miss this for all of us already used to working with the project.

lukaspie: Right, I missed that

domna: Yes, we should moderate all of this here before doing the merges anyways. Because this will also affect all open PRs. We can also switch with nomad (dec 20th)

sherjeelshabih (approved):

PR #204, opened 2024-01-09
APM refactoring / APMv2

mkuehbach:


pynxtools/dataconverter/readers/apm/utils/apm_create_nx_default_plots.py
```diff
@@ -31,73 +41,61 @@ def create_default_plot_reconstruction(template: dict, entry_id: int) -> dict:
...
-    imx = np.ceil(bounds[1, 1]) + resolution
-    yedges = np.linspace(imi, imx, num=int(np.ceil((imx - imi) / resolution)) + 1,
-                         endpoint=True)
-    imi = np.floor(bounds[2, 0]) - resolution
-    imx = np.ceil(bounds[2, 1]) + resolution
-    zedges = np.linspace(imi, imx,
-                         num=int(np.ceil((imx - imi) / resolution)) + 1,
-                         endpoint=True)
+    # make the bounding box a quadric prism, discretized using cubic voxel edge in nm
+    aabb: dict = {"x": [0., 0.],
```
lukaspie, line 45: Is it possible to find a better (i.e., more self-explanatory) name than `aabb`? Maybe not, I don't really know the use case.

lukaspie: In general, variable names in this function could be improved.

mkuehbach: "aabb" is short-hand for axis_aligned bounding-box which is an accepted term in the computational geometry community. imi and imx indicate extrema values for dimensions where the i is a placeholder for x, y, z, admittedly maybe not pythonic enough

mkuehbach:
Before this PR can be merged the current state of the master has to be merged in again but this is not a blocker for people to review the implementation of the parser as it is proposed in this PR.

pynxtools/dataconverter/readers/apm/utils/apm_create_nx_default_plots.py
lukaspie: Maybe add some comments here in the code to briefly explain these terms.

apm.dev.ipynb
```diff
@@ -0,0 +1,87 @@
+{
```
sanbrock, line 1: Should not it be moved to pynxtools/dataconverter/readers/apm/dev or debug?
Also the whole debug directory could go there, couldn't it?

sanbrock (changes requested):

apm.dev.ipynb
sherjeelshabih: I will even say that instead of having "debug" code. Whatever this is debugging should become test code for that same functionality. This way it is easy to confirm that it is "de-bugged" in the CI without having to run anything locally for reviewers.

mkuehbach:
The code in this feature branch is taken as the starting point for the creation of the pynxtools-apm plugin.
All future development, including the linting with ruff is implemented and discussed here:
www.github.com/FAIRmat-NFDI/pynxtools-apm

apm.dev.ipynb
mkuehbach: I agree that is all what this directory on my computer is about, I have collected a corpus of about 160 real world examples from atom probe groups worldwide (published research) against which I am running the converter via this directory. Some of these datasets are openly accessible and could thus indeed be placed or used for writing tests, the apm dataconverter code has been made a pynxtools plugin and that directory has been removed

mkuehbach:
Information for end users:
With https://github.com/FAIRmat-NFDI/pynxtools-apm/commit/5645326895122586d24569683e5b9b26a8e47565
can be used again as a plugin to standardize atom probe via NeXus, we are currently working on documentation and integrating the plugin into NOMAD Oasis

PR #209, opened 2024-01-17
Set logger levels through handler

### LINKED ISSUE #208: dataconverter does not correctly pick up --generate template flag
The `--generate-template` flag of `pynxtools/dataconverter` seems to not be working for me at the moment. If I pass that flag, the template is not printed to the console.

Probably comes from changes to the logger in https://github.com/FAIRmat-NFDI/pynxtools/pull/191. This seems to be related to the setting of the logger level, see [here](https://stackoverflow.com/a/59705351).

Quick fix: Change to `logger.error(template)` in https://github.com/FAIRmat-NFDI/pynxtools/blob/c4d1b9cf8fdabe1a9d49ec54b09a78e5b773e28b/pynxtools/dataconverter/convert.py#L248

We should check that all other flags (like `--fair`, `--undocumented`) are still picked up correctly.

@domna @sherjeelshabih @RubelMozumder

lukaspie:
Configures logging with a [handler](https://docs.python.org/3/library/logging.html#logging.Handler). If you only configure a level for your logger, but no handlers, the internal handler is used which is set to output just the message (with no other formatting) at the `WARNING ` level (see [this reference](https://stackoverflow.com/a/59705351)). Currently, this prevents the printing of `INFO` level loggings to the console (including when using the `generate-template` flag).

Fixes #208.


pynxtools/dataconverter/logger.py
```diff
@@ -22,4 +22,8 @@
 logger = logging.getLogger("pynxtools")
 
 # Lowest level log allows to other levels erros, crittical, info and debug
+handler = logging.StreamHandler()
+formatter = logging.Formatter("%(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
```
sherjeelshabih, line 28: I think some code, imports maybe, are setting the basic loglevel to warning again. If I set logging.basicconfig(level=debug) it works again.

Will these changes here affect Nomad's log output?

lukaspie: If I understand correctly, this should only effect the pynxtools level logger, but I am not sure.

sherjeelshabih:
@lukaspie @domna I could test running the converter locally and everything works fine for me. 

I couldn't find any outputs in the Nomad GUI both before and after these changes. And I also couldn't find any call from Nomad to the dataconverter with a logger being passed.

domna:
> @lukaspie @domna I could test running the converter locally and everything works fine for me.
> 
> I couldn't find any outputs in the Nomad GUI both before and after these changes. And I also couldn't find any call from Nomad to the dataconverter with a logger being passed.

Here is the comment on Rubel where he reports that this works in nomad: https://github.com/FAIRmat-NFDI/pynxtools/pull/191#issuecomment-1839535098
However, it looks like he just tested this locally and we still need to pass the logger to the convert function in nomad. @sherjeelshabih Did you also try to pass it to convert in nomad and see if it is then logged in the gui?

sherjeelshabih:
> > @lukaspie @domna I could test running the converter locally and everything works fine for me.
> > I couldn't find any outputs in the Nomad GUI both before and after these changes. And I also couldn't find any call from Nomad to the dataconverter with a logger being passed.
> 
> Here is the comment from Rubel where he reports that this works in nomad: [#191 (comment)](https://github.com/FAIRmat-NFDI/pynxtools/pull/191#issuecomment-1839535098) However, it looks like he just tested this locally and we still need to pass the logger to the convert function in nomad. @sherjeelshabih Did you also try to pass it to convert in nomad and see if it is then logged in the gui?

I changed the way Nomad deals with this. I do not want to have a hot potato being passed around in function arguments. :smile: 

I just access the logging object of pynxtool during runtime from Nomad and switch it to what Nomad needs it to work. This way we don't have to change much in our code or add a special check if we are running under Nomad etc. 

See this MR on Nomad: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1638
Let me know what you think.

domna:
> > > @lukaspie @domna I could test running the converter locally and everything works fine for me.
> > > I couldn't find any outputs in the Nomad GUI both before and after these changes. And I also couldn't find any call from Nomad to the dataconverter with a logger being passed.
> > 
> > 
> > Here is the comment from Rubel where he reports that this works in nomad: [#191 (comment)](https://github.com/FAIRmat-NFDI/pynxtools/pull/191#issuecomment-1839535098) However, it looks like he just tested this locally and we still need to pass the logger to the convert function in nomad. @sherjeelshabih Did you also try to pass it to convert in nomad and see if it is then logged in the gui?
> 
> I changed the way Nomad deals with this. I do not want to have a hot potato being passed around in function arguments. 😄
> 
> I just access the logging object of pynxtool during runtime from Nomad and switch it to what Nomad needs it to work. This way we don't have to change much in our code or add a special check if we are running under Nomad etc.
> 
> See this MR on Nomad: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1638 Let me know what you think.

That's perfect 👍️

pynxtools/dataconverter/convert.py
```diff
@@ -351,6 +344,7 @@ def convert_cli(
     mapping: str,
 ):
     """The CLI entrypoint for the convert function"""
+    logger.addHandler(logging.StreamHandler())
```
domna, line 347: Is it needed to set this manually? I always thought logging always sets this as standard

domna (approved):
LGTM, except one small question whether we need to set the streamhandler manually (but there is probably a good reason).

pynxtools/dataconverter/convert.py
sherjeelshabih: It seems we do need to add this for levels below warn. Logging.warn and above have a fallback handler if I remember correctly. 

We can also do logging.basicConfig. That just adds this as part of the config setting routine.

lukaspie: We want to go at least to `logging.INFO`, even if we don't use it, right? Then this solution works. 

I thought `logging.basicConfig` [sets the log level for the root logger](https://docs.python.org/3/library/logging.html#logging.basicConfig), would that not effect the NOMAD logging? Maybe I misunderstand..

lukaspie:
LGTM, feel free to merge anytime you guys are happy with it.

pynxtools/dataconverter/convert.py
domna: For nomad it should be fine as we should never end up in this part of the code anyways. Generally, I'm just more used to seeing this being set via `basicConfig` but I think just adding the streamHandler is also fine and might be safer in case this function is execution by nomad for whatever reason.

So just go ahead :)

lukaspie: Alright, thanks for explaining 👍

PR #217, opened 2024-01-23
Adds link to wilkinson et al

sherjeelshabih:


domna (approved):
Thanks for the PR. I fixed two other errors but I think now everything should be fine

Issue #219, opened 2024-01-23
[Bug]: Handling of links in XPS reader


lukaspie:
### What happened?

When having links of the form `"group_a/field_a":" {"link": "group_b/field_b"}"` in the config file, the XPS reader is supposed to populate the template with for an entry `<entry1>` with `/ENTRY[<entry1>]/group_a/field_a: {'link': '/<entry1>/group_b/field_b'}`.

However, when there are multiple entries (`<entry1>,<entry2>`), the template for `<entry2>` is filled with `/ENTRY[<entry2>]/group_a/field_a: {'link': '/<entry1>/group_b/field_b'}`. That is, it links back to `<entry1>` instead of `<entry2>`.

PR #220, opened 2024-01-23
Hotfixes for XPS reader

lukaspie:
- Fixes issue with channel reading
 - Linking with multiple entries partially resolved (see also #219)


lukaspie:
Right, my bad. Thanks for the review.

PR #229, opened 2024-02-02
XPS vamas sub-reader: enhance data extraction

lukaspie:
This is a draft for enhancing the existing Vamas XPS reader.
The Vamas reader reads .vms files, which is the current ISO standard data transfer format ([ISO 14976](https://www.iso.org/standard/24269.html)).

- [x] Write CPS data to file
- [ ]  Extract units for quantities
- [ ] Parse vamas files of type IRREGULAR


lukaspie:
Closed as this will be addressed in the xps reader plugin: https://github.com/FAIRmat-NFDI/pynxtools-xps/pull/10

PR #237, opened 2024-02-12
Fixes NXsensor_scan and axes attribute error

domna:
This re-enables NXsensor_scan as an appdef and fixes a bug in which the the reading fails with an error when the axes array length is smaller than the dim of the data array.


sanbrock (approved):
LGTM

PR #242, opened 2024-02-20
Switches to NXscan for eln test

domna:
The eln test should not depend on the NXmpes definition as this may change.
I switched it to NXscan as this is a definition which is already contributed and will likely not change in the coming months.


RubelMozumder (approved):
This is just the chnage of the file to be subject in this tool and the change in the corresponding reference to the file. Everything looks fine

Issue #225, opened 2024-01-31
Implement proper semantic versioning and release v1.0.0


domna:
As requested yesterday in our TF meeting by @rettigl we should follow semantic versioning guidelines to not introduce breaking changes.

My suggestion is to follow proper [semantic versioning guidelines](https://semver.org). Currently, we are in some `0.x` version which is fine to break anything from semver perspective. We should, however, think towards a 1.0.0 release to allow users to reliably refer to pynx.

The semver guidelines required a detailed API of the software so it's easy to understand when something breaks. They say it can also be implemented in code, which I would interpret as having tests. We also wanted to test the api towards nomad as well, so this might be the starting point of getting a proper test coverage with a following 1.0.0 release.

An associated but more tricky question is how to keep the definitions used flexible. We would then need a specifier to accept certain versions a version of pynx works with.

What do you think @FAIRmat-NFDI/areab ? We should probably discuss this a bit further at one of our TF meetings and collect our decisions here.

rettigl:
My suggestion for the latter part would be to provide the definitions of a certain version also as python package. Then we can make a certain pynxtools version require a certain nexus definition version.

lukaspie:
See https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/#public-apis

PR #243, opened 2024-02-20
Pluginise mpes reader

### LINKED ISSUE #234: Pluginise mpes reader
We could also think about having a generic igor reader that we can then map for the different techniques.

domna:
The MPES reader is moving into its own plugin at https://github.com/FAIRmat-NFDI/pynxtools-mpes

Fixes #234


domna:
@sherjeelshabih I removed all the mpes related tests and move this into the reader plugin. Do you think we need to recreate some tests for the example reader to cover functionality?

@rettigl This is mainly fyi. When everything is moved you need to do an additional `pip install pynxtools-mpes` and everything should be picked up automatically from there. I will also add this as an install extra as [mpes]

lukaspie:
> @rettigl This is mainly fyi. When everything is moved you need to do an additional pip install pynxtools-mpes and everything should be picked up automatically from there. I will also add this as an install extra as [mpes]

That's a good idea also for the other plugins. Maybe we could also make an install extra as [all-readers] or something that install all plugins (at least the ones that we are aware of or for which we provide its own install extra)?

domna:
> > @rettigl This is mainly fyi. When everything is moved you need to do an additional pip install pynxtools-mpes and everything should be picked up automatically from there. I will also add this as an install extra as [mpes]
> 
> That's a good idea also for the other plugins. Maybe we could also make an install extra as [all-readers] or something that installs all plugins (at least the ones that we are aware of or for which we provide its own install extra)?
> 

Yes, this is a good idea. We should have this for all plugins we support and I think it's also fine if people want to add their readers there.
> EDIT: I think in general it would be nice to also list the supported plugins in the pynxtools main repo somewhere (maybe also in README?)

Yes, maybe we can get the list from the [pynxtools-plugin](https://github.com/topics/pynxtools-plugin) topic. That way it would automatically update :) If not we can just keep a list but link to the topic.

domna:
#247

sherjeelshabih:
> @sherjeelshabih I removed all the mpes related tests and move this into the reader plugin. Do you think we need to recreate some tests for the example reader to cover functionality?

I'll have a look on what tests etc will be needed separately. You can keep this PR for dealing with the MPES reader. I'll make an issue to check test coverage again for the JOSS paper.

lukaspie (approved):
LGTM, will do the same for https://github.com/FAIRmat-NFDI/pynxtools-xps.

rettigl:
I tested it, and it works, however this changes the functionality of pynxtools, so we should discuss how to adjust versioning accordingly (we are probably the only ones using the mpes reader right now, but the next version of pynxtools without extras definition will break the behavior).

domna:
> I tested it, and it works, however this changes the functionality of pynxtools, so we should discuss how to adjust versioning accordingly (we are probably the only ones using the mpes reader right now, but the next version of pynxtools without extras definition will break the behavior).

Yes, we should follow proper semantic versioning for pynxtools to indicate when we introduce breaking changes. See #225. Ideally, we would now do the main changes until a v1.0.0 release so that we don't have to introduce breaking changes for a long time.

I would do a v0.1.0 release for the next version to indicate that we break something.

PR #246, opened 2024-02-21
Add files as argument in CLI

### LINKED ISSUE #238: Rework the CLI interface and support folder reading
I think it would be helpful to use the dataconverter like this: `dataconverter --reader my_reader --nxdl NXtest FILENAME FILENAME FILENAME`, i.e., without the `--input-file`. This should be supported be click (there is a lot offered for convenience file interaction, so I think we can also use more of what click supplies). Ideally, we would make this backwards compatible such that `--input-file` is still supported.

Also being able to read complete folders would be useful, i.e., the reader would then read any files inside that folder and/or it is possible to design readers which work on folders directly. This is useful for config/data file combinations.

domna:
Fixes #238

- Supports adding files via `dataconverter --reader my_reader --nxdl NXmynxdl filename1 filename2`.
- You can also pass a folder and it will recursively get all the files in it and pass it to the reader function.
- Shell pattern replacement works out of the box `dataconverter --reader my_reader --nxdl NXmynxdl *.h5` or `dataconverter ... my_folder/**/*` for files in folders
- Keeps the old `--input-file` option but shows a deprecation warning.


domna (commented):
I set the whole Area-B team in review so you are aware of this change. Please give it a run and see if it works well for you

RubelMozumder:
LGTM, I have two points to mention here.
1. Could we also keep the old syntax `--input-file`, alongside with the new way of converter,  but given a warning that we are going replace the syntax with the new sytax. If you implement it directly all the `NORTH` will get infected and make the users confused.
2. Do you think to keep `--input-file` only for once for all the given input files. Which would be intuitive, what is the input file and what is the `nxdl`.

tests/dataconverter/test_convert.py
```diff
@@ -17,15 +17,17 @@
 #
 """Test cases for the convert script used to access the DataConverter."""
```
RubelMozumder, line 19: I do not see any reflection on the test. The test can also be modified if needed.

lukaspie:
> > 2. Do you think to keep `--input-file` only for once for all the given input files. Which would be intuitive, what is the input file and what is the `nxdl`.
> 
> I don't understand what you mean. The nxdl is clearly specified by `--nxdl NXmynxdl` so I think it should be obvious which argument is the nxdl.

If I understand @RubelMozumder correctly, you want to still keep `--input-file` and then allow to pass multiple file? That would be very confusing. It would need to be changed to `--input-files` at least and I don't think there is big improvement over just having no flag (like what Florian is proposing here).

tests/dataconverter/test_convert.py
domna: I removed the `--input-file` inputs except for `test_cli` as we cannot test this like this. click is already checking that the file we pass is existing, therefore we never start parsing there. But I think it's fine to still test the option there.

domna:
> > > 2. Do you think to keep `--input-file` only for once for all the given input files. Which would be intuitive, what is the input file and what is the `nxdl`.
> > 
> > 
> > I don't understand what you mean. The nxdl is clearly specified by `--nxdl NXmynxdl` so I think it should be obvious which argument is the nxdl.
> 
> If I understand @RubelMozumder correctly, you want to still keep `--input-file` and then allow to pass multiple file? That would be very confusing. It would need to be changed to `--input-files` at least and I don't think there is big improvement over just having no flag (like what Florian is proposing here).

Yes, I agree with you @lukaspie. In this case the direct file arguments should be used (it should also work now with shell pattern replacement, e.g., something like `dataconverter *.h5` should only run for all the h5 files in the folder). This is also much more aligned with how common cli tools work.

lukaspie:
> > > > 3\. Do you think to keep `--input-file` only for once for all the given input files. Which would be intuitive, what is the input file and what is the `nxdl`.
> > > 
> > > 
> > > I don't understand what you mean. The nxdl is clearly specified by `--nxdl NXmynxdl` so I think it should be obvious which argument is the nxdl.
> > 
> > 
> > If I understand @RubelMozumder correctly, you want to still keep `--input-file` and then allow to pass multiple file? That would be very confusing. It would need to be changed to `--input-files` at least and I don't think there is big improvement over just having no flag (like what Florian is proposing here).
> 
> Yes, I agree with you @lukaspie. In this case the direct file arguments should be used (it should also work now with shell pattern replacement, e.g., something like `dataconverter *.h5` should only run for all the h5 files in the folder). This is also much more aligned with how common cli tools work.

Great, then we all agree to use positional arguments. I will test now with the XPS reader and the data that I have here locally.

sherjeelshabih:
I just checked it for a couple of runs of the dataconverter I had in my terminal history. They work well. 

I just have one suggestion. In the --input-file help text, can you bring the Deprecated warning before the actual doc? This way it's more obvious for someone who is quickly doing a --help and not reading thoroughly. It is at the end currently without any space after the bracket.

```console
--input-file TEXT               The path to the input data file to read.
                                  (Repeat for more than one file.)Deprecated:
                                  Please use the positional file arguments
                                  instead.
```

domna:
> I just checked it for a couple of runs of the dataconverter I had in my terminal history. They work well.
> 
> I just have one suggestion. In the --input-file help text, can you bring the Deprecated warning before the actual doc? This way it's more obvious for someone who is quickly doing a --help and not reading thoroughly. It is at the end currently without any space after the bracket.
> 
> ```
> --input-file TEXT               The path to the input data file to read.
>                                   (Repeat for more than one file.)Deprecated:
>                                   Please use the positional file arguments
>                                   instead.
> ```

I change it, now it looks like this:
```
 --input-file TEXT               Deprecated: Please use the positional file
                                  arguments instead. The path to the input
                                  data file to read. (Repeat for more than one
                                  file.)
```

tests/dataconverter/test_convert.py
RubelMozumder: There are some `--input-file` options in the test file and they can be removed.

RubelMozumder (approved):

RubelMozumder (approved):
I approved it, if all are fixed it is ready to merge.

tests/dataconverter/test_convert.py
domna: No the only occurrence left is https://github.com/FAIRmat-NFDI/pynxtools/blob/f982c0e16636a6a5f87870f868014ab47188b871/tests/dataconverter/test_convert.py#L126-L127 which cannot be removed, because click automatically checks if `test_input` exists (and it doesn't). We can remove this test entirely but I would keep it until we completely remove `--input-file`.

lukaspie:
As discussed with Florian already, making `--nxdl` required breaks the usage of `params-file`. When I use `dataconverter --params-file params.yaml`, it complains that you need to pass an nxdl.

sherjeelshabih:
> As discussed with Florian already, making `--nxdl` required breaks the usage of `params-file`. When I use `dataconverter --params-file params.yaml`, it complains that you need to pass an nxdl.

Yeah. That's why I added that custom check. I see that Florian has already replaced the IOError with clicks own set. This should be fine with having `--nxdl` not required in click's config.

pynxtools/dataconverter/convert.py
```diff
@@ -348,29 +356,44 @@ def convert_cli(
...
-        if mapping:
-            reader = "json_map"
-            if mapping:
-                input_file = input_file + tuple([mapping])
-        convert(input_file, reader, nxdl, output, generate_template, fair, undocumented)
+    if nxdl is None:
+        raise click.UsageError("Missing option '--nxdl'")
+    if mapping:
+        reader = "json_map"
+        input_file = input_file + tuple([mapping])
```
sherjeelshabih, line 374: I think this should be moved back in the else statement. This will also not work with the --params if an --nxdl is not provided, right?

domna: It will be fine, I added a `return` statement in the `try` block for the params config to get rid of the additional else level.

domna: But maybe let's add a testing case for the params file before we merge this. To not break it again

sherjeelshabih: You're right. I just saw it a second ago. Thank you. I'll mark this as resolved.

domna: I added the test @lukaspie @sherjeelshabih

lukaspie: Great, feel free to merge then 👍

PR #248, opened 2024-02-21
Update lint configuration in pyproject.toml

domna:
With newer ruff versions we get this warning

```
❯ ruff format .
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'fixable' -> 'lint.fixable'
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
141 files left unchanged
```

~Maybe we can also check the alignment with nomad here again. As I think some settings have changed there.~ I checked, we are still aligned with nomad (except for the version pinning).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,12 +87,12 @@ local_scheme = "node-and-date"
 
 [tool.ruff]
 include = ["pynxtools/*.py", "tests/*.py"]
-select = [
+lint.select = [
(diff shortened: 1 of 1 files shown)


lukaspie (approved):
LGTM, I have already done this with the lint.* locally (and in another private repo) and it does not break anything.

PR #250, opened 2024-02-21
Adds multi format reader

### LINKED ISSUE #213: Streamline reading of json and yaml files
Many of our readers read json and yaml files to add additional metadata.
Some of these features are made generally available in the `pynxtools.utils` but reader implementations differ.

1. I think we should streamline yaml and json reading and offer this as a general solution for adding additional keys to an appdef. There is a generic `YamlJsonReader` already going in that direction, but it is just used by few readers.
2. I think this would also be useful for plugin developers as they can concentrate on writing the actual data conversion and just can use the general handling of additional metadata through pynxtools.

domna:
Fixes #213

ToDo:
- [x] Add support for config file parsing (helper in utils + default parsing)
- [x] Add support for nomad eln file parsing (helper in utils + default parsing)

## Will not be implemented here
- Add string parsing support from config file (as in #252)
- Add formula parsing support from config file (as in #252)

## Considerations
- Probably this should replace the YamlJsonReader entirely
- For that the hall and transmission reader need to be changed accordingly


domna (commented):
Hey @lukaspie, I started to work on a generic format reader. We should align here and come up with a common json config file notation and then we can do the parsing from the this reader instead directly in our plugins.

sherjeelshabih:
This sounds close to what the JsonMapReader does. It already supports multiple formats. Maybe we can bring that in as well.

domna:
> This sounds close to what the JsonMapReader does. It already supports multiple formats. Maybe we can bring that in as well.

Indeed, it is very similar. The main difference I see is that the json map reader uses a mapping file where we want to use a combination file. It is a json which either contains literal values or loads data from another source via `@data:...` or similar.

However, the way I want to design this is that you can always change how certain formats are read, i.e., that you specify a dict of file-endings + read functions (actually, this is how the YamlJsonReader already works). That way the json map reader could also be implemented in this fashion as an implementation for reading `.mapping.json` + `.pickle` files. But we could also directly integrate mapping and pickling support (even though I would be happy to not rely on pickling at all).

sherjeelshabih:
> > This sounds close to what the JsonMapReader does. It already supports multiple formats. Maybe we can bring that in as well.
> 
> Indeed, it is very similar. The main difference I see is that the json map reader uses a mapping file where we want to use a combination file. It is a json which either contains literal values or loads data from another source via `@data:...` or similar.
> 
> However, the way I want to design this is that you can always change how certain formats are read, i.e., that you specify a dict of file-endings + read functions (actually, this is how the YamlJsonReader already works). That way the json map reader could also be implemented in this fashion as an implementation for reading `.mapping.json` + `.pickle` files. But we could also directly integrate mapping and pickling support (even though I would be happy to not rely on pickling at all).

I think we should get rid of pickling eventually. I'll need to change the NXiv_temp example. 

If we will have a generic reader that can be customized like this, then we should streamline it and direct users to one unified place for such functions. I'll have a thorough look later into how the YamlJsonReader works. I don't mind making big changes to bring the two together. The less clutter we have the less we need to work on and users need to figure out.

domna:
> > > This sounds close to what the JsonMapReader does. It already supports multiple formats. Maybe we can bring that in as well.
> > 
> > 
> > Indeed, it is very similar. The main difference I see is that the json map reader uses a mapping file where we want to use a combination file. It is a json which either contains literal values or loads data from another source via `@data:...` or similar.
> > However, the way I want to design this is that you can always change how certain formats are read, i.e., that you specify a dict of file-endings + read functions (actually, this is how the YamlJsonReader already works). That way the json map reader could also be implemented in this fashion as an implementation for reading `.mapping.json` + `.pickle` files. But we could also directly integrate mapping and pickling support (even though I would be happy to not rely on pickling at all).
> 
> I think we should get rid of pickling eventually. I'll need to change the NXiv_temp example.
> 
> If we will have a generic reader that can be customized like this, then we should streamline it and direct users to one unified place for such functions. I'll have a thorough look later into how the YamlJsonReader works. I don't mind making big changes to bring the two together. The less clutter we have the less we need to work on and users need to figure out.

Yes, this is also my main endeavour with this PR. You can check the hall or transmission reader to see an implementation using this (I think it's easier to grasp from there). Here is the actual reader for hall:

```python
class HallReader(YamlJsonReader):
    """HallReader implementation for the DataConverter
    to convert Hall data to Nexus."""

    supported_nxdls: List[str] = ["NXroot"]
    extensions = {
        ".txt": lambda fname: parse_txt(fname, "iso-8859-1"),
        ".json": parse_json,
        ".yml": lambda fname: parse_yml(fname, None, None),
        ".yaml": lambda fname: parse_yml(fname, None, None),
    }
```

I think it's easy and straightforward to understand and all of the parsing is offloaded into the `parse_...` functions.

The current plan with this MultiFormatReader is the following:

- We allow using an extensions dict in an inheriting reader, which maps file ending to parse function.
- The reader should pickup common functionality for standard file endings (json config files, nomad eln yaml's), which can be overridden with the extensions dict (we already have a `parse_json` helper in pynxtools.dataconverter.readers.utils so readers can also re-use them along other files. We should also add one for yaml and set them as default in MultiFormatReader)
- We should advocate this generic reader for most cases and the base reader if you want to do something special

lukaspie:
FWIW, I have taken quite a different approach in the [XPS reader](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/reader.py), with different classes for the individual sub-readers for different formats and the selection based on the file format happening in a different [script](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/file_parser.py). So if we implement this, I will have to change the XPS reader quite a bit. I would be happy to do so if we get more aligned across the different readers and if the config/eln file parsing is unified.

domna:
> FWIW, I have taken quite a different approach in the [XPS reader](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/reader.py), with different classes for the individual sub-readers for different formats and the selection based on the file format happening in a different [script](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/file_parser.py). So if we implement this, I will have to change the XPS reader quite a bit. I would be happy to do so if we get more aligned across the different readers and if the config/eln file parsing is unified.

I think we might be able to support both. As far as I understood it from a first glance you keep a dict with a reader class. So we could use this in the multi format reader to instantiate the class and invoke its read function whenever we find a class instead of a function. I think this is also a nice feature, that you can either specify a simple read function or a complete reader.

But I would anyways keep the config parsing as a separate function. So you could also just reuse this in your code.

lukaspie:
> > FWIW, I have taken quite a different approach in the [XPS reader](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/reader.py), with different classes for the individual sub-readers for different formats and the selection based on the file format happening in a different [script](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/pynxtools_xps/file_parser.py). So if we implement this, I will have to change the XPS reader quite a bit. I would be happy to do so if we get more aligned across the different readers and if the config/eln file parsing is unified.
> 
> I think we might be able to support both. As far as I understood it from a first glance you keep a dict with a reader class. So we could use this in the multi format reader to instantiate the class and invoke its read function whenever we find a class instead of a function. I think this is also a nice feature, that you can either specify a simple read function or a complete reader.
> 
> But I would anyways keep the config parsing as a separate function. So you could also just reuse this in your code.

Yeah, it's not actually a "reader" that's called in the sense that it inherits from BaseReader 😄 I have a three-layer structure: there is the `XpsReader(BaseReader)`, then I have a layer that I call "mappers" (see [here](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/facb11fce4835c040736db901fc590a9304c1fea/pynxtools_xps/sle/sle_specs.py#L44) for an example) and these mappers themselves actually call "parsers" ([example](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/facb11fce4835c040736db901fc590a9304c1fea/pynxtools_xps/sle/sle_specs.py#L1544)). The mappers are used for one file format (like the sle format from SPECS) and then I have a logic that calls a parser for a specific subsets of such files, e.g. depending on the software version that was used for this file. That allows me to keep functionality across different, yet very similar versions of a format by inheritance and abstract base classes.

Now that I think about it, maybe all of those sub-classes could be readers themselvers, inheriting from our `BaseReader` class (or the `MultiFormatReader`). In any case, having the option to instantiate a "reading"/"mapping"/"parsing" class and invoke its top-level (aka `read`) function through the extension dict would be nice. We just have to make sure that these functions are always called the same.

I additionally have the problem that the extension is often not unique, i.e., many vendors have a .txt export, but all the files are actually different. But this logic could probably be handled by passing a function in the extensions dict that does this. I am actually doing that [already](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/facb11fce4835c040736db901fc590a9304c1fea/pynxtools_xps/file_parser.py#L107). And finally, I also have a check that the file comes from the list of supported vendors.

sherjeelshabih:
There is also an idea to bring a unified json config and mapping to the dataconverter itself. This is still under discussion but I'm leaving this here as a reminder.

domna:
@sherjeelshabih and @lukaspie, I finished the first draft of the MultiFormatReader now. I did not test it yet but I think we can use this as a basis to discuss about the structure.

The general mechanism is as detailed above, that we use the `extensions` dict to figure out which extension is handled by which function and the reader just glues everything together. I guess we could also introduce a testing function in this dict to distinguish different formats if you find this helpful (e.g., when `.txt` files with different syntax are used).

There are also two special functions now: `setup_template` just fills in keys before anything is done (for fixed info like reader name and version and `handle_objects`, which handles the objects passed to the reader function.

What this reader supports is a config dict with some special syntax which is noted as, e.g., `@attrs:...` in the values of the dict. The config dict itself is always reduced to a flat dict, hence users can group it any way they want. We support the following special keywords:

- `@attrs`: Load an attribute from the input data, i.e., this is coming from somewhere else in the complete data passed to the reader. We use this in the mpes reader to load data from metadata in an hdf5 file. It takes a path, i.e., something like `@attrs:/my_metadata/entry`. The MultiFormatReader has a function `get_attr`, which takes a path and returns the data for that field. The sub-reader would then implement this function to provide the actual data to write.
- `@data`: Works similarly to `@attrs`, but for data entries. This is split in the mpes reader because we retrieve data from an xarray. It is resolved by the `get_data` function in the reader. Maybe this can also be aligned with the `@attrs` keyword.
- `@link`: This constructs a link to the provided path. It is the replacement for the `{'link': '/path'}` syntax we cannot use anymore because of the flattening
- There is also some special syntax with wildcard (`*`) replacements, this is what the `get_data_dims` function is for. It gives a list of all data axes available to replace the wildcard. We use this to fill axisnames automatically from xarray dims in the mpes reader (see [here](https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/24eaed611f8f3a80d5bd721b6fa1ff7f94ff24dc/tests/data/config_file.json#L392) for an usage example)
- We also support a wildcard with a name to repeat entries with different names (e.g., `*{my, name, etc}` replaces to three keys with `my`, `name`, `etc` replaced, respectively). [Here](https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/24eaed611f8f3a80d5bd721b6fa1ff7f94ff24dc/tests/data/config_file.json#L114) you can see it in action for the lens voltage readout in mpes (same mapping but different lens name).

[Here](https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/main/tests/data/config_file.json) is the mpes example config file as an implementation of the config file syntax above.

I also had a look at the json map reader and from my view it basically does a `@attrs` on each value in comparison to the notation above. So we could just do an "auto-`@attrs`" for the mapping. The question is whether it would also be feasible to use the `@link` in the mapping and I think we would need to come up with a replacement for the slicing syntax you use since this is embedded in the `shape` key of a nested dict (which the config reader would flatten away).

- @lukaspie Does this roughly align with the config file you use for xps?
- Do you think we should move this as a standard functionality in the BaseReader? People could just override the read function if they don't want the dict behaviour and ignore everything else or they can extend the list with their special parsing for a particular file format.

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    special_key_map = {
+        "@attrs": lambda _, v: ParseJsonCallbacks.attrs_callback(v),
+        "@link": lambda _, v: ParseJsonCallbacks.link_callback(v),
+        "@data": lambda _, v: ParseJsonCallbacks.data_callback(v),
+    }
+    attrs_callback: Callable[[str], Any] = lambda v: v
+    data_callback: Callable[[str], Any] = lambda v: v
+    link_callback: Callable[[str], Any] = lambda v: {"link": v}
```
lukaspie, line 75: This just replaces `@link: v` with `{"link": v}`? So the actual link resolving still occurs in the writer?

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    if callbacks is None:
+        # Use default callbacks if none are explicitly provided
+        callbacks = ParseJsonCallbacks()
+
+    config_file_dict = parse_flatten_json(file_path)
+    fill_wildcard_data_indices(config_file_dict, callbacks.dims)
+
+    for key, value in config_file_dict.items():
+        if not isinstance(value, str) or ":" not in value:
```
lukaspie, line 100: I was wondering if I may be possible to give a list of options in the config file. Something like:
- first check in @attrs
- if the key doesn't exist in @attrs, check the eln
- if it's not there either, write a default or nothing at all

This would probably be achieved by passing a list in the config.

I guess this comes down to how we see the config file. I have been using it in a way that there are default configs for each subreader and that the user ideally doesn't have to change the config file when converting. So then such a list of options would allow to pass an ELN file where a field is missing without breaking the reader, but still have a default config file. This may also come in handy if you have an additional reader for metadata that only comes into place when you upload a metadata file as well (see what I started [here](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/c0788e5a8ba6e8c769d9ca956aded28d83dae512/pynxtools_xps/reader.py#L295)).

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+        data_callback (Callable[[str], Any]):
+            The callback to retrieve the data under the specified key.
+        dims (List[str]):
+            The dimension labels of the data. Defaults to None.
+    """
+
+    special_key_map = {
+        "@attrs": lambda _, v: ParseJsonCallbacks.attrs_callback(v),
+        "@link": lambda _, v: ParseJsonCallbacks.link_callback(v),
+        "@data": lambda _, v: ParseJsonCallbacks.data_callback(v),
```
lukaspie, line 71: I am currently also reading an `@eln` token to read metadata from the ELN file.

In fact, I am first parsing all files and THEN I let the config decide which data to take. As of right now, the ELN always gets priority here.

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        template = Template()
+        self.kwargs = kwargs
+
+        def sort_keys(filename: str) -> str:
+            """
+            Makes sure to read json and yaml files last
+            """
+            ending = os.path.splitext(filename)[1]
+            return ".zzzzz" if ending in (".json", ".yaml", ".yml") else ending
```
lukaspie, line 185: This effectively means the ELN data always get priority over the actual files, even if you have the `@attrs` keyword, right?

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+        self.extensions[".json"] = lambda fn: parse_json_config(
+            fn,
+            ParseJsonCallbacks(
+                attrs_callback=self.get_attr,
+                data_callback=self.get_data,
+                dims=self.get_data_dims,
+            ),
+        )
+
+    def setup_template(self) -> Dict[str, Any]:
```
lukaspie, line 134: This aligns well, I have something similar implemented.

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+                )
+                continue
+            if not os.path.exists(file_path):
+                print(f"WARNING: File {file_path} does not exist, ignoring entry.")
+                continue
+
+            template.update(self.extensions.get(extension, lambda _: {})(file_path))
+
+        template.update(self.setup_template())
+        template.update(self.handle_objects(objects))
```
lukaspie, line 203: I am not sure what you mean by objects here.

pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,208 @@
...
+                print(
+                    f"WARNING: "
+                    f"File {file_path} has an unsupported extension, ignoring file."
+                )
+                continue
+            if not os.path.exists(file_path):
+                print(f"WARNING: File {file_path} does not exist, ignoring entry.")
+                continue
+
+            template.update(self.extensions.get(extension, lambda _: {})(file_path))
```
lukaspie, line 200: Alternative: I have a [function](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/c0788e5a8ba6e8c769d9ca956aded28d83dae512/pynxtools_xps/reader.py#L409) that only updates if there is nothing overwritten that already exists in the template.

lukaspie:
What I don't understand: how are multiple entries handled?

In the XPS reader, I need to always change the template key from `"/ENTRY[entry]/"` (which is in the config) to `"/ENTRY[entry_1]/"`, `"/ENTRY[entry_2]/"`, and so on. The same is actually true for links, where I put `"@link:/entry/instrument/beam_probe"` in the config and then the actual links are linking within the same entry (i.e., to `/entry_1/instrument/beam_probe`. 

Basically I need to iterate the config file and replace the key for each entry. I can do that inside the different functions, e.g. in `get_data`, but maybe there could be a more sophisticated version right at a higher level.

pynxtools/dataconverter/readers/multi/reader.py
domna: Not necessarily, they would have the same sort key so it depends on their initial order (so it's kind of arbitrary from user input). I also thought about separating this and always do the config replacements last.

pynxtools/dataconverter/readers/multi/reader.py
domna: You can pass a python object to the read function. We use this in the mpes reader to directly inject an xarray for the data. Since the object doesn't have a file name it has its own function.

pynxtools/dataconverter/readers/multi/reader.py
domna: Let's discuss this. In my view it might be useful sometimes to allow overwriting, e.g., with the json dict. But I could also imagine giving the reader a flag to switch this behaviour and set one of the options as default.

domna:
> What I don't understand: how are multiple entries handled?
> 
> In the XPS reader, I need to always change the template key from `"/ENTRY[entry]/"` (which is in the config) to `"/ENTRY[entry_1]/"`, `"/ENTRY[entry_2]/"`, and so on. The same is actually true for links, where I put `"@link:/entry/instrument/beam_probe"` in the config and then the actual links are linking within the same entry (i.e., to `/entry_1/instrument/beam_probe`.
> 
> Basically I need to iterate the config file and replace the key for each entry. I can do that inside the different functions, e.g. in `get_data`, but maybe there could be a more sophisticated version right at a higher level.

Yes, this is a good comment. I also thought about this already but did not put it yet, because we just did not have this functionality in the mpes reader/config.
We could also have some kind of wildcard syntax or just let the reader replace `/ENTRY[entry]/` with the appropriate entry name (that way your json dict is a template for a single entry but your reader will create multiple entries based on this template).

lukaspie:
> We could also have some kind of wildcard syntax or just let the reader replace /ENTRY[entry]/ with the appropriate entry name (that way your json dict is a template for a single entry but your reader will create multiple entries based on this template).

This is how I currently have it.

pynxtools/dataconverter/readers/multi/reader.py
domna: Let's discuss this. I was also thinking whether we merge `@attrs` and `@data` and let the reader figure out where to take data from. That would, however, make more logic necessary to be implemented by the person implementing the reader.

pynxtools/dataconverter/readers/multi/reader.py
domna: Correct, we just cannot use the nesting because it will be flattened out by the json reader (to allow for some logical grouping for the users).

pynxtools/dataconverter/readers/multi/reader.py
domna: Yes, I think this also relates to my thoughts to your comment below. Let's discuss it.

lukaspie:
Eventually, bringing in this [functionality](https://github.com/FAIRmat-NFDI/pynxtools-stm/blob/6614b8a3ac78c5dd8958bd037d0b37b14b143cda/pynxtools_stm/reader.py#L186) that Rubel wants to implement anyway (see #255) would be very good as well.

pynxtools/dataconverter/readers/multi/reader.py
lukaspie: Could you please point me to where you are cunrrently doing that in the mpes reader? I am not sure I fully understand it yet, but might be useful for me too.

rettigl:
I have another point to add here: For the new pase classes such as NXcalibration, I typically add nested sections with the required entries, where many of the entries are hard-coded, such as `physical_quantity`, `type`, etc. The only value coming from a data source is typically the `resolution` value, e.g.:
```
    "energy_resolution": {
      "resolution": "@attrs:metadata/instrument/electronanalyser/energy_resolution",
      "resolution/@units": "meV",
      "physical_quantity": "energy",
      "type": "derived"
    },
```
Now, if I don't have any data for `@attrs:metadata/instrument/electronanalyser/energy_resolution` available in a dataset, this will be skipped by the reader, and not available, yet the other entries are present. This results in the writer complaining that the required field `resolution` of the optional group `energy_resolution` is not present, and actually even raises an error at the moment:
```
LookupError: The data entry, /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/energy_resolution/physical_quantity, has an optional parent, /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/energy_resolution, with required children set. Either provide no children for /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/energy_resolution or provide all required ones.
```

To avoid this, it would be good to introduce another flag/mechanism, that allows omitting the whole parent group, if a particular value is not present.

pynxtools/dataconverter/readers/multi/reader.py
domna: Yes, we read the object here: https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/c1ecb70fe80c8f44cd4d6bc01bf5c2429fa98462/pynxtools_mpes/reader.py#L250 (it get's just passed down into this function from the read function itself)

In this example we pass an xarray to the convert function (in the second to last section called "Create a NeXus file from a xarray"): https://gitlab.mpcdf.mpg.de/nomad-lab/north/mpes/-/blob/main/example/E2%20Binning%20of%20WSe2.ipynb?ref_type=heads

domna: And actually we mistreat it a bit as we just pass an xarray object and not a tuple with one object (as the signature indicates). Maybe we can also discuss whether we wan't to allow type `Any` for the objects so it's completely up to the reader what to expect.

We have, however, also a mechanism for passing `**kwargs` down to the read function. This might even be a better option for readers to specialize their input to certain objects.

domna:
I implemented most of the functions. I decided to implement tests first and then do the formula, slicing and string parsing parts (maybe in a separate PR) as these mainly add new special keywords.

There are some points which I did not tick, @lukaspie. Below you find my thoughts on them.

> * [ ]  **Default config files**. Either let the reader (or a sub-reader) set the config file and parse it directly from the reader (like in pynxtools-xps) or allow a user to store their default experiment configuration somewhere.

I added it that it can be passed as CLI argument but can also be set in the reader as `self.config_file` (this will be overwritten when a config file is supplied via CLI).
A default config file I would handle in a separate PR if we want this. We might also collect whether we want to add additional info into it (i.e. loading a nexus repo from another location).

> * [ ]  Optional: **automatically write default  attribute** to immediate child groups ([Adding default attribute to immediate child group #255](https://github.com/FAIRmat-NFDI/pynxtools/issues/255)) or at least have a function that _can_ do this (like [this](https://github.com/FAIRmat-NFDI/pynxtools-stm/blob/6614b8a3ac78c5dd8958bd037d0b37b14b143cda/pynxtools_stm/reader.py#L186)).

I would deal with this in a separate PR. I feel this one is already getting pretty large and the `@default` attribute does not really belong to this changes.

> Maybe it could also be possible that a (sub-)reader has a default config file (essentially covering the basics of what is always in an expected file format) and then if you provide another config file, you can specify which fields to overwrite or additionally read? An example would be the Vamas format, which in general always has the same structure (ISO standard), but you can (and vendors/analysis software do) write additional data/processing into the comment lines.

I thought about this, too. I decided against it because I felt it overcomplicates the situation right now and it might also be simpler for users to supply one config file and that's it. I think it's also better for maintenance in the long run (i.e. having one file as configuration vs. having a chain of config file which depends on the order of input even if this means duplicating some lines). In principle it should still be possible for a sub-reader if you re-use the json parsing function for all json files but there would still be one main json config.

> I have another point to add here: For the new pase classes such as NXcalibration, I typically add nested sections with the required entries, where many of the entries are hard-coded, such as physical_quantity, type, etc. The only value coming from a data source is typically the resolution value, e.g.:

I like the idea but I'm still thinking how to implement this properly (i.e. how does the reader know which fields to omit). Maybe I will introduce a prefix like `?resolution:`, which means to write the entry only if there exists a field named `resolution` at the same level.

lukaspie:
> > * [ ]   **Default config files**. Either let the reader (or a sub-reader) set the config file and parse it directly from the reader (like in pynxtools-xps) or allow a user to store their default experiment configuration somewhere.
> 
> I added it that it can be passed as CLI argument but can also be set in the reader as `self.config_file` (this will be overwritten when a config file is supplied via CLI). A default config file I would handle in a separate PR if we want this. We might also collect whether we want to add additional info into it (i.e. loading a nexus repo from another location).

This is what I wanted for now. I would just have a folder with all the config files and point the reader to the specific file.

> > * [ ]   Optional: **automatically write default  attribute** to immediate child groups ([Adding default attribute to immediate child group #255](https://github.com/FAIRmat-NFDI/pynxtools/issues/255)) or at least have a function that _can_ do this (like [this](https://github.com/FAIRmat-NFDI/pynxtools-stm/blob/6614b8a3ac78c5dd8958bd037d0b37b14b143cda/pynxtools_stm/reader.py#L186)).
> 
> I would deal with this in a separate PR. I feel this one is already getting pretty large and the `@default` attribute does not really belong to this changes.

Fine for me.

> > Maybe it could also be possible that a (sub-)reader has a default config file (essentially covering the basics of what is always in an expected file format) and then if you provide another config file, you can specify which fields to overwrite or additionally read? An example would be the Vamas format, which in general always has the same structure (ISO standard), but you can (and vendors/analysis software do) write additional data/processing into the comment lines.
> 
> I thought about this, too. I decided against it because I felt it overcomplicates the situation right now and it might also be simpler for users to supply one config file and that's it. I think it's also better for maintenance in the long run (i.e. having one file as configuration vs. having a chain of config file which depends on the order of input even if this means duplicating some lines). In principle it should still be possible for a sub-reader if you re-use the json parsing function for all json files but there would still be one main json config.

Yeah, it could get pretty tricky very soon. You can always just copy the default config and make changes as you wish.

> Link renaming if there are multiple. On the value side, the same syntax should work for links, i.e, replace @link: entry*/my_field in the config by {@link: entry_1/my_field} in the template.

Thanks also for implementing this. However, I still see the problem with having multiple collection columns, detectors, etc. Those would not get replaced, right? This is why I was advocating for this `entry*`, `detector*` notation, so one knows which sets of multiple instances need to be provided. The default can still be to write just one (like you do it for entry now).

domna:
> Thanks also for implementing this. However, I still see the problem with having multiple collection columns, detectors, etc. Those would not get replaced, right? This is why I was advocating for this entry*, detector* notation, so one knows which sets of multiple instances need to be provided. The default can still be to write just one (like you do it for entry now).

Yes, I opted for just the entry. My feeling was that the general algorithm would be fairly big and I thought it might be easier to leave this to the reader implementation, because they know already which multiple groups will be present. To do this at config level we'd need yet another callback. I thought we can provide a convenience function to provide renaming in the reader classes easily.

domna:
Hey @lukaspie,

I finished and tested this. There is a working implementation for mpes: https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/6

lukaspie:
> Hey @lukaspie,
> 
> I finished and tested this. There is a working implementation for mpes: [FAIRmat-NFDI/pynxtools-mpes#6](https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/6)

Cool, thanks for bringing this forwards. Can you rebase this branch? There seem to be some old commits mixed in.

domna:
> > Hey @lukaspie,
> > I finished and tested this. There is a working implementation for mpes: [FAIRmat-NFDI/pynxtools-mpes#6](https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/6)
> 
> Cool, thanks for bringing this forwards. Can you rebase this branch? There seem to be some old commits mixed in.

Yes, sorry for that. There was a main merge with somehow did strange stuff. Now it should be fine

lukaspie (commented):
@domna I am currently trying to implement this for the XPS reader, but I am having problems with the `parse_json_config` part. 

This is what the outputs of one of my sub-readers would look like:
image

Essentially, one huge dict, with the metadata for each entry in "/ENTRY[<my-entry>]/<some-string>. My initial idea was to use e.g. "/ENTRY[entry]/method":"@attrs:region/analysis-method", but I am not sure what to implement in the `get_attr` function then.

For the data, it is all stored in the `data` key, which is a dict of xarrays (one for each entry). They contain scan data like this:
```
<xarray.Dataset> Size: 88kB
Dimensions:             (energy: 1001)
Coordinates:
  * energy              (energy) float64 8kB 175.0 174.9 174.8 ... 75.1 75.0
Data variables:
    cycle1              (energy) float64 8kB 191.0 177.8 143.1 ... 83.77 76.35
    cycle1_scan1        (energy) float64 8kB 179.6 194.8 152.7 ... 86.71 85.19
    cycle1_scan1_chan0  (energy) float64 8kB 179.6 194.8 152.7 ... 86.71 85.19
    cycle1_scan2        (energy) float64 8kB 208.0 152.3 128.7 ... 79.35 63.09
    cycle1_scan2_chan0  (energy) float64 8kB 208.0 152.3 128.7 ... 79.35 63.09
    cycle1_scan3        (energy) float64 8kB 120.7 156.5 163.8 ... 74.6 49.62
    cycle1_scan3_chan0  (energy) float64 8kB 120.7 156.5 163.8 ... 74.6 49.62
    cycle2              (energy) float64 8kB 140.6 258.7 283.5 ... 65.11 62.47
    cycle2_scan1        (energy) float64 8kB 140.6 258.7 283.5 ... 65.11 62.47
    cycle2_scan1_chan0  (energy) float64 8kB 140.6 258.7 283.5 ... 65.11 62.47
 ```
 I want all scans which have `_chan` in them to go to detector/raw and the rest to go to data(NXdata). I also want to write the average and the errors for the default view. Any hints how to implement this with the `get_data` function? Thanks!

**Comment by domna** [comment]
> Essentially, one huge dict, with the metadata for each entry in "/ENTRY[]/. My initial idea was to use e.g. "/ENTRY[entry]/method":"https://github.com/attrs:region/analysis-method", but I am not sure what to implement in the get_attr function then.

You can just prefix with the `self.callbacks.entry_name`:

```python
def get_attr(self, path: str) -> Any:
  return self.data_dict.get(f"/ENTRY[{self.callbacks.entry_name}/{path}")
```
(assuming `self.data_dict` is your dictionary)

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+class MultiFormatReader(BaseReader):
+    """
+    A reader that takes a mapping json file and a data file/object to return a template.
+    """
+
+    # Whitelist for the NXDLs that the reader supports and can process
+    supported_nxdls: List[str] = []
+    extensions: Dict[str, Callable[[Any], dict]] = {}
```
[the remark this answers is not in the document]
domna: Yes, this is to be populated by the reader. It should work in the `__init__` _and_ just set as a class parameter. For mpes I do it in the __init__ (https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/8c0f685b92bd548c891042eb4d1e10ecd79c0a9c/pynxtools_mpes/reader.py#L177). But mainly I also do other stuff there.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,333 @@
...
+            return to_bool(value)
+
+        return value
+
+    if callbacks is None:
+        # Use default callbacks if none are explicitly provided
+        callbacks = ParseJsonCallbacks()
+
+    config_file_dict = parse_flatten_json(file_path)
+    fill_wildcard_data_indices(config_file_dict, callbacks.dims)
```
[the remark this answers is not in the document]
domna: It generally goes through all the keys. But I think the part I forgot is to actually pass an entry name to the `callback.dims` to resolve the dimensions of the current entry.

domna:
> I want all scans which have _chan in them to go to detector/raw and the rest to go to data(NXdata). I also want to write the average and the errors for the default view. Any hints how to implement this with the get_data function? Thanks!

You can basically return anything from the `get_data` function what you want. `self.callbacks.entry_name` will give you the current entry name you can use to resolve your data for the entry. Other than that you can just map the path how you want and just reference it in the config dict, i.e., `"my_key": "@data:raw0"` could just map `cycle1_scan1_chan0`, `@data:raw1` to `cycle1_scan1_chan1` etc. You basically can invent your own path notation. Don't know if this helps or if you want to additionally have something which expands all of your raw entries, i.e., writing `@data:raw` populate all `cycle1_scan1_chanX` to the fields.

src/pynxtools/dataconverter/template.py
```diff
@@ -38,14 +43,16 @@ def __init__(self, template=None, **kwargs):
             self.optional_parents: list = copy.deepcopy(template["optional_parents"])
             self.lone_groups: dict = copy.deepcopy(template["lone_groups"])
         else:
-            self.optional: dict = {}
-            self.recommended: dict = {}
-            self.required: dict = {}
-            self.undocumented: dict = {}
-            self.optional_parents: list = []
-            self.lone_groups: list = []
+            self.optional: dict = {}  # type: ignore[no-redef]
```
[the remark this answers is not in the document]
domna: This makes sense. Will change it. Thanks!

src/pynxtools/dataconverter/readers/multi/reader.py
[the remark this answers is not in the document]
domna: I updated the `get_data_dims` process. It now receives a path which contains the value of the json dict, e.g., `@data:*.index`. This includes the special keyword (because we only resolve it after filling the wildcards).

Also the dims get now called inside the loop, such that `self.callbacks.entry_name` is properly set and can be used inside the `get_data_dims` function.

rettigl:
The "!" flag still appears to be broken (entries are not being removed).

domna:
> The "!" flag still appears to be broken (entries are not being removed).

It should be fixed now

lukaspie:
> > The "!" flag still appears to be broken (entries are not being removed).
> 
> It should be fixed now

It worked for me, thanks for the fix. But, there was no output like 
```
logger.info(
    f"Main element {key} not provided. "
    f"Removing the parent group {group_to_delete}."
)
```
I guess this is related to what @rettigl was saying here: https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/22#issuecomment-2240405266

domna:
It's strange. I get the output when I use the dataconverter via cli. It could be that the logging level is set too low in some tests? Because I use the info level here. In tests we normally only catch the `WARNING` level, so this would pass by these. We can, however, also make it a warning but I thought this is not a real problem of the reader but rather something intended.

domna:
I think we have the other readers failing, because they all use `parse_yml` and expect a proper dict entry being written (because there is not the dict post processing we do). We could do a replacement of `ENTRY` to `ENTRY[entry]` in the normal base reader.

lukaspie:
> It's strange. I get the output when I use the dataconverter via cli. It could be that the logging level is set too low in some tests? Because I use the info level here. In tests we normally only catch the `WARNING` level, so this would pass by these. We can, however, also make it a warning but I thought this is not a real problem of the reader but rather something intended.

Hm weird, I don't actually get this if I dataconverter via cli, but the group does get removed..

domna:
> Hm weird, I don't actually get this if I dataconverter via cli, but the group does get removed..
This is really strange. THis is what I get with the mpes example:

```
❯ dataconverter --reader mpes --nxdl NXmpes xarray_saved_small_calibration.h5 -c config_file.json --output example.nxs
Using mpes reader to convert the given files:  
• xarray_saved_small_calibration.h5
INFO:pynxtools.dataconverter.readers.multi.reader:Main element /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy not provided. Removing the parent group /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe].
WARNING:pynxtools.dataconverter.helpers:The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy is required and hasn't been supplied by the reader.
The output file generated: example.nxs.
INFO:pynxtools.dataconverter.convert:The output file generated: example.nxs.
```

and when I set the logging level to warning in the multi reader it vanishes:
```
❯ dataconverter --reader mpes --nxdl NXmpes xarray_saved_small_calibration.h5 -c config_file.json --output example.nxs
Using mpes reader to convert the given files:  
• xarray_saved_small_calibration.h5
WARNING:pynxtools.dataconverter.helpers:The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy is required and hasn't been supplied by the reader.
The output file generated: example.nxs.
INFO:pynxtools.dataconverter.convert:The output file generated: example.nxs.
```

Does it work for you when you change the `logger.info` to `logger.warning`?

rettigl:
> > Hm weird, I don't actually get this if I dataconverter via cli, but the group does get removed..
> 
> This is really strange. THis is what I get with the mpes example (with `/ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy` set with `!` to non-existing path):
> 
> ```
> ❯ dataconverter --reader mpes --nxdl NXmpes xarray_saved_small_calibration.h5 -c config_file.json --output example.nxs
> Using mpes reader to convert the given files:  
> • xarray_saved_small_calibration.h5
> INFO:pynxtools.dataconverter.readers.multi.reader:Main element /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy not provided. Removing the parent group /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe].
> WARNING:pynxtools.dataconverter.helpers:The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy is required and hasn't been supplied by the reader.
> The output file generated: example.nxs.
> INFO:pynxtools.dataconverter.convert:The output file generated: example.nxs.
> ```
> 
> and when I set the logging level to warning in the multi reader it vanishes:
> 
> ```
> ❯ dataconverter --reader mpes --nxdl NXmpes xarray_saved_small_calibration.h5 -c config_file.json --output example.nxs
> Using mpes reader to convert the given files:  
> • xarray_saved_small_calibration.h5
> WARNING:pynxtools.dataconverter.helpers:The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/beamTYPE[beam_probe]/incident_energy is required and hasn't been supplied by the reader.
> The output file generated: example.nxs.
> INFO:pynxtools.dataconverter.convert:The output file generated: example.nxs.
> ```
> 
> Does it work for you when you change the `logger.info` to `logger.warning`?

Can you check if these come from stdout or stderr? If there is no handler defined, and you issue a warning, a default handler to stderr is installed to the root logger. This does Apparently not output the first warning that installs it, but only the ones after. There normally should also be a "key not found" info, no?

domna:
> Can you check if these come from stdout or stderr?

All messages go to stderr. If I append `2> /dev/null` to the dataconverter call the output is empty, for `> /dev/null` I get everything.

> There normally should also be a "key not found" info, no?

No, the optional group removal skips the report of the missing key. Which is fine I guess since the message _is_ telling that the key could not be found and for that reason the whole group is removed.

> You also can tell from the last line, which is being repeated. It is output once from the logger in convert, and once from the last resort root logger

I see we only add the handler in `convert`. I added a proper configuration of the root logger now. I think with this it should work (everything goes to stderr still).

lukaspie:
> I think we have the other readers failing, because they all use `parse_yml` and expect a proper dict entry being written (because there is not the dict post processing we do). We could do a replacement of `ENTRY` to `ENTRY[entry]` in the normal base reader.

I am not sure what you mean here. The base reader does not call the `parse_yml` function, right? I guess we have to do this somewhere in that function.

domna:
> I am not sure what you mean here. The base reader does not call the `parse_yml` function, right? I guess we have to do this somewhere in that function.

It doesn't. But the readers which are failing use the parse_yml function to parse the eln yamls without the MultiFormatReader.
And they assume that it returns a valid path (i.e., including the lowercase entry).

lukaspie:
@domna https://github.com/FAIRmat-NFDI/pynxtools-xps/pull/74 is now fully compatible with this branch 

> > I am not sure what you mean here. The base reader does not call the `parse_yml` function, right? I guess we have to do this somewhere in that function.
> 
> It doesn't. But the readers which are failing use the parse_yml function to parse the eln yamls without the MultiFormatReader. And they assume that it returns a valid path (i.e., including the lowercase entry).

Yeah, I get that. Shall we just allow that one can pass a parent key to the `parse_yml` function and make it `ENTRY[entry]` again by default? Then we can just set it to `ENTRY` in the MultiFormatReader.

lukaspie:
@domna https://github.com/FAIRmat-NFDI/pynxtools-xps/pull/74 is now fully compatible with this branch, so from the XPS side we are fine 👍 just need the other plugins to pass

rettigl:
> > The "!" flag still appears to be broken (entries are not being removed).
> 
> It should be fixed now

It still does not work properly. If I remove "beam_pump/incident_energy" from the dict in specsanalyzer: Tutorial 3, All elements except for beam_pump/distance are being removed. I try to dig in to understand why this one is remaining.

rettigl:
> > > The "!" flag still appears to be broken (entries are not being removed).
> > 
> > 
> > It should be fixed now
> 
> It still does not work properly. If I remove "beam_pump/incident_energy" from the dict in specsanalyzer: Tutorial 3, All elements except for beam_pump/distance are being removed. I try to dig in to understand why this one is remaining.

The issue is that this parsing is happening sequentially. So all keys of a group_to_remove that occurred before the missing key are not being removed. 
The removal has to happen after filling the whole dict, as was the case in the pynxtools-mpes solution.

domna:
> The issue is that this parsing is happening sequentially. So all keys of a group_to_remove that occurred before the missing key are not being removed.
The removal has to happen after filling the whole dict, as was the case in the pynxtools-mpes solution.

I see. That's why I sort keys beginning with `!` before other keys before I go into the for loop. But obviously there is something wrong with it

rettigl:
> I see we only add the handler in `convert`. I added a proper configuration of the root logger now. I think with this it should work (everything goes to stderr still).

I don't like this solution with the root logger. This affects other packages as well (e.g. sed-processor), where loggers for individual modules are created. Now, here everything gets printed twice.

IMHO a proper logger configuration should be:
Each module defines a named logger, removes all existing handlers, and adds one (or more) specific handlers for this logger. Then no double logging should occur even if files are being reloaded, and no interdependence to other modules is created.

domna:
> I see. That's why I sort keys beginning with ! before other keys before I go into the for loop. But obviously there is something wrong with it

Indeed, I accidentally sorted for the key. This is fixed now and the group is entirely removed in the example (when I remove the `incident_energy` entry). Also the dict is sorted correctly now.

domna:
> IMHO a proper logger configuration should be:
Each module defines a named logger, removes all existing handlers, and adds one (or more) specific handlers for this logger. Then no double logging should occur even if files are being reloaded, and no interdependence to other modules is created.

Ok, there is something weird going on. Even when I disable _all_ handlers for the root logger _and_ my custom module logger it still would output some of the logging. Probably, one of the modules we import does change the root logger settings. However, I now do `propagate = False` and do not touch root logger settings. That way pynxtools is not affected by any other package altering the root logger _and_ is not changing the behaviour for other packages. That did shut down the erratic behaviour for me. Can you check if it's fine for your packages now, too?

lukaspie (approved):
From my side, all looks good. Just waiting to see if the logging works properly now.

domna:
From my side as well. I set you as reviewer @rettigl. When you are happy we can merge this + the respective changes in xps and mpes.

rettigl:
> > IMHO a proper logger configuration should be:
> > Each module defines a named logger, removes all existing handlers, and adds one (or more) specific handlers for this logger. Then no double logging should occur even if files are being reloaded, and no interdependence to other modules is created.
> 
> Ok, there is something weird going on. Even when I disable _all_ handlers for the root logger _and_ my custom module logger it still would output some of the logging. Probably, one of the modules we import does change the root logger settings. However, I now do `propagate = False` and do not touch root logger settings. That way pynxtools is not affected by any other package altering the root logger _and_ is not changing the behaviour for other packages. That did shut down the erratic behaviour for me. Can you check if it's fine for your packages now, too?

Most things work now, however tests in sed are still failing. This is because with propagate=False, no log messages can be read by caplog, so the test for expected log warnings fails, see also here: https://github.com/pytest-dev/pytest/issues/3697

The culprit for the eratic behavior seems to have been several calls to logging.info (rather than to a specific logger), see e.g. here: https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/740fc35ce5ccfbd2fb98599989405b5960da49ea/pynxtools_mpes/reader.py#L231

If I replace these, re-add the propagate option, and use suppress_warning, this part seems to work.

However, ELN entries (from .yaml files) don't seem to end up in the dictionary propperly at the moment (this is also tested in the sed write tests, which fails at that point).

rettigl:
I tried to understand how this logic works with the callbacks, but I don't quite get it. The eln values are written into self.eln_data, but never read from there, it appears to me. 

Also, how do I set priority levels between eln data from a file, and the xarray provided as object? As far as I understand, priorities are only possible to set for files (?)

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+            logger.warning(
+                f"Could not find value for key {key} with value {value}.\n"
+                f"Tried prefixes: {prefixes}."
+            )
+
+    # after filling, resolve links again:
+    if isinstance(new_entry_dict.get(key), str) and new_entry_dict[key].startswith(
+        "@link:"
+    ):
+        new_entry_dict[key] = {"link": new_entry_dict[key][6:]}
```
rettigl, line 227: There is this create_link_dict option in the json parser now. I don't understand how it works and what it's good for: Does it have to be considered here as well?

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+    ) -> dict:
+        """
+        Reads data from multiple files and passes them to the appropriate functions
+        in the extensions dict.
+        """
+        self.kwargs = kwargs
+        self.config_file = self.kwargs.get("config_file", self.config_file)
+        self.overwrite_keys = self.kwargs.get("overwrite_keys", self.overwrite_keys)
+
+        template = Template(overwrite_keys=self.overwrite_keys)
```
rettigl, line 398: For my tests, this created an empty template. Not sure if this is what it's supposed to do.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+
+        def get_processing_order(path: str) -> Tuple[int, Union[str, int]]:
+            """
+            Returns the processing order of the file.
+            """
+            ext = os.path.splitext(path)[1]
+            if self.processing_order is None or ext not in self.processing_order:
+                return (1, ext)
+            return (0, self.processing_order.index(ext))
```
rettigl, line 407: How do I influence processing between objects and file types?

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+            if extension not in self.extensions:
+                logger.warning(
+                    f"File {file_path} has an unsupported extension, ignoring file."
+                )
+                continue
+            if not os.path.exists(file_path):
+                logger.warning(f"File {file_path} does not exist, ignoring entry.")
+                continue
+
+            template.update(self.extensions.get(extension, lambda _: {})(file_path))
```
rettigl, line 421: What does this line do? For my tests, nothing.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+
+        template.update(self.setup_template())
```
rettigl, line 423: What does this line do? For my tests, nothing. (apparently). template is still empty afterwards.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+        template.update(self.handle_objects(objects))
```
rettigl, line 424: After this, the template is still empty (handle objects returns {}, I think).

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if self.config_file is not None:
+            self.config_dict = parse_flatten_json(
+                self.config_file, create_link_dict=False
+            )
```
rettigl, line 429: I don't understand the part with the links.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+
+        self.post_process()
```
rettigl, line 431: What is it supposed to do?

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if self.config_dict:
+            template.update(
+                fill_from_config(
```
rettigl, line 435: This only fills the template for me, but only with the values from the xarray, not from the eln file.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+                    self.config_dict,
+                    self.get_entry_names(),
+                    self.callbacks,
+                    **{k: v for k, v in kwargs.items() if k == "suppress_warning"},
```
rettigl, line 439: Why not pop "suppress_warning" from the kwargs before, set it as variable and pass it here? this seems awkwardly complicated to me.

src/pynxtools/dataconverter/readers/multi/reader.py
rettigl, line 1: This thing is seriously sophisticated, and definately needs some detailed documentation and explanation of options and use cases.

rettigl (changes requested):
I tried to understand how this works, but there are various places I don't understand the logic or intended operation. This might be to a good part to still missing documentation.

As for the mpes reader, which I took as test example, the new version does not take the provided ELN data into account as the previous version did. Maybe a different config file would be required for this, but this is not clear to me entirely.

src/pynxtools/dataconverter/readers/multi/reader.py
domna: Currently, you cannot. The objects are always parsed after _all_ files. See L425 where the parsing of objects happen. We could also introduce an `objects` part here to denote when to parse the objects. However, for mpes it does not really matter because we load all files and then process the json file afterwards (with getting all the infos from the `@` keywords)

src/pynxtools/dataconverter/readers/multi/reader.py
domna: That's correct. This is the initialisation of the template and should just create an empty template

src/pynxtools/dataconverter/readers/multi/reader.py
domna: Yes, because in the mpes reader we collect the data and then retrieve it from the callbacks while going through the json file. Therefore for mpes it just returns an empty dict and updates nothing. However, this reader also supports that reading a file can also fill the template directly, then it would return a non-empty dict.

src/pynxtools/dataconverter/readers/multi/reader.py
domna: Same as above

src/pynxtools/dataconverter/readers/multi/reader.py
domna: Same as above

src/pynxtools/dataconverter/readers/multi/reader.py
domna: This is basically a compatibility feature. For some readers we just use the `parse_flatten_json` function to read in json files. There we need to replace `@link` with a dict of the form `{"link": "..."}`. For this reader we want to use the get_link callback function, so we don't want to do this replacement beforehand (otherwise any special treatmant in the callback will not be applied).
In case we use the default get_link callback this doesn't matter because they both do the same thing.

src/pynxtools/dataconverter/readers/multi/reader.py
domna: It's a post processing step after all files are read in and before the actual json parsing happens. As an example, we use it in the igor reader to collect a list of files and `post_process` actually reads them in and creates a data array to be read out by the `get_data` callback.
This is also a function which depends a bit on the design of the respective sub-reader (some might not use them at all).

src/pynxtools/dataconverter/readers/multi/reader.py
domna: I agree. This might be a bit overcomplicated

src/pynxtools/dataconverter/readers/multi/reader.py
domna: Yes, here finally the json file is parsed which calls all the different callbacks to retrieve the infos.
Acutally, yes the eln handling is now a bit different. Previously, we just overwrote with what we found in the template but now we have to explicity tell it, e.g., we need to write a key `"/ENTRY/my_value": "@eln"` to tell that we actually want to retrieve it from an eln. But the old overwriting behaviour can be implemented as well. I guess it's your choice what you prefer for the mpes reader.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+            f"Main element {key} not provided. "
+            f"Removing the parent group {group_to_delete}."
+        )
+        optional_groups_to_remove.append(group_to_delete)
+        return
+
+    if prefixes and new_entry_dict[key] is None:
+        del new_entry_dict[key]
+        if not suppress_warning:
+            logger.warning(
```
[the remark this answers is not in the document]
domna: Updated

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+            break
+        new_entry_dict[key] = callbacks.apply_special_key(prefix, key, path)
+
+        # We found a match. Stop resolving other prefixes.
+        if new_entry_dict[key] is not None:
+            break
+
+    if value.startswith("!") and new_entry_dict[key] is None:
+        group_to_delete = key.rsplit("/", 1)[0]
+        logger.info(
```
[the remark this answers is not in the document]
domna: Updated

src/pynxtools/dataconverter/readers/multi/reader.py
domna: No, it's just backword compatibility (see the other comment on the issue)

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+        list_value = ast.literal_eval(value)
+        prefixes = [parse_config_value(v) for v in list_value]
+    except (SyntaxError, ValueError):
+        prefixes = [parse_config_value(value)]
+
+    for prefix, path in prefixes:
+        if not prefix.startswith("@"):
+            new_entry_dict[key] = try_convert(path)
+            break
+        new_entry_dict[key] = callbacks.apply_special_key(prefix, key, path)
```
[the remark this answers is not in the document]
domna: You can use the `@eln` keyword as explained in my response to the other comment here

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+        prefixes = re.findall(pattern, value)
+        if not prefixes:
+            return ("", value)
+        return prefixes[0]
+
+    # Handle non-keyword values
+    if not isinstance(value, str) or "@" not in str(value):
+        if isinstance(value, str):
+            # Handle "!" notation for required fields in optional groups
+            value = value.lstrip("!")
```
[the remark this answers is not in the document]
domna: I think we will never end up in this code part with just a single literal value. Even if we did I think it would be fine to remove the `!` and do no extra check for literal values (would prevent to write literals beginning with `!` though)

src/pynxtools/dataconverter/readers/multi/reader.py
domna: I agree. It is not well documented and it should even have a dedicated page in the documentation on how to use it for different use cases eventually (I think actually using this is much simpler than understanding this code here).

rettigl:
I adopted a config file for specsscan now, and this seems to work: https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/25

The ! mechanism did not work properly together with lists in entries, which I fixed now, please have a look. In particular, lists did not get parsed if a ! was in front, and adding the ! into the list elements did not trigger the removal machanism...

src/pynxtools/dataconverter/readers/multi/reader.py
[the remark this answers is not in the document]
domna: Ok, now while having a closer look to this it makes sense to remove this check.

domna:
> The ! mechanism did not work properly together with lists in entries, which I fixed now, please have a look. In particular, lists did not get parsed if a ! was in front, and adding the ! into the list elements did not trigger the removal machanism...

Yes, indeed the previous notation was implemented as `[!@eln,....]` but the below check checked the start of the whole value string so it couldn't work. I'm removing also the `!` in the regex pattern as we don't really use it there.

src/pynxtools/dataconverter/readers/multi/reader.py
rettigl: I'd suggest to either add documentation in this PR, or add an issue to add documentation later on.

src/pynxtools/dataconverter/readers/multi/reader.py
rettigl: There is a working solution now for the MPES reader.

rettigl (approved):
Together with https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/6 this should work now at least for the mpes reader.

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -0,0 +1,446 @@
...
+        ):
+            dim_dict[new_key] = new_val
+
+    return dim_dict
+
+
+class ParseJsonCallbacks:
+    """
+    Callbacks for dealing special keys in the json file.
+
```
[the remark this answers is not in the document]
lukaspie: I added a bit more description. Should however also be explained within #398

src/pynxtools/dataconverter/readers/multi/reader.py
lukaspie: I changed this as suggested.

tests/dataconverter/test_validation.py
```diff
@@ -24,6 +24,24 @@
...
+                restore += [(logger, logger.propagate)]
+                logger.propagate = True
+        except AttributeError:
+            pass
+    yield caplog
+
+    for logger, value in restore:
+        logger.propagate = value
+
+
```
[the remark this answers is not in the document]
domna: Removed

tests/dataconverter/test_helpers.py
```diff
@@ -30,6 +30,24 @@
...
+                restore += [(logger, logger.propagate)]
+                logger.propagate = True
+        except AttributeError:
+            pass
+    yield caplog
+
+    for logger, value in restore:
+        logger.propagate = value
+
+
```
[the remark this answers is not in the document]
domna: Removed

tests/dataconverter/test_convert.py
```diff
@@ -28,6 +28,24 @@
...
+                restore += [(logger, logger.propagate)]
+                logger.propagate = True
+        except AttributeError:
+            pass
+    yield caplog
+
+    for logger, value in restore:
+        logger.propagate = value
+
+
```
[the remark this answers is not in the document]
domna: Removed

src/pynxtools/__init__.py
```diff
@@ -16,13 +16,18 @@
...
 import re
 from datetime import datetime
 
 from pynxtools._build_wrapper import get_vcs_version
 from pynxtools.definitions.dev_tools.globals.nxdl import get_nxdl_version
 
+logger = logging.getLogger("pynxtools")
+logger.addHandler(logging.StreamHandler())
+logger.setLevel(logging.INFO)
+
```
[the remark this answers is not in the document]
lukaspie: added through #399. We decided to only print the log level if it is warning or higher in order to keep the output relatively small.

PR #253, opened 2024-02-21
Pluginise XPS reader

lukaspie:


domna:
I updated the regression file and added a small fix for the read functions test that we only test if there is a data folder with them (was failing for me locally because I had `pynxtools-mpes` installed).

For whatever reason the regression file changed now to the one we always generated locally (so hopefully this mystery is gone now)

domna:
Do you also want to add xps as an extra install to `pyproject.toml`?

domna (approved):
LGTM

domna:
> > Do you also want to add xps as an extra install to `pyproject.toml`?
> 
> I still have some problems with packaging for the XPS plugin, so maybe we could just add it in a separate PR (or when we add the extra install for all readers)? Alternatively, I can also wait with this PR here until I have sorted out the kinks there, but I don't want to have parallel code in two places for much longer.

I think a separate PR is fine. Just wanted to make sure that you did not forget to add it.

PR #256, opened 2024-02-22
Fixes child retrieval for fields

domna:
This fixes the child retrievel which resulted in an error for non-existing fields, because parts of the nxdl paths were removed

### DIFF
--- a/pynxtools/dataconverter/helpers.py
+++ b/pynxtools/dataconverter/helpers.py
@@ -112,7 +112,7 @@ def get_all_defined_required_children(nxdl_path, nxdl_name):
 def add_inherited_children(list_of_children_to_add, path, nxdl_root, template):
     """Takes a list of child names and appends them to template for a given path."""
     for child in list_of_children_to_add:
-        child_path = f"{path.rsplit('/', 1)[0]}/{child}"
+        child_path = f"{path}/{child}"
         if child_path not in template.keys():
             optional_parent = check_for_optional_parent(child_path, nxdl_root)
             optionality = (
@@ -164,9 +164,9 @@ def generate_template_from_nxdl(
         ):
             template[optionality][f"{path}/@units"] = None
 
-        parent_path = convert_data_converter_dict_to_nxdl_path(path.rsplit("/", 1)[0])
+        nxdl_path = convert_data_converter_dict_to_nxdl_path(path)
         list_of_children_to_add = get_all_defined_required_children(
-            parent_path, nxdl_name
+            nxdl_path, nxdl_name
         )
         add_inherited_children(list_of_children_to_add, path, nxdl_root, template)
 
--- a/pynxtools/dataconverter/helpers.py
+++ b/pynxtools/dataconverter/helpers.py
@@ -482,6 +482,7 @@ def check_optionality_based_on_parent_group(path, nxdl_path, nxdl_root, data, te
 
 def is_group_part_of_path(path_to_group: str, path_of_entry: str) -> bool:
     """Returns true if a group is contained in a path"""
+
     tokens_group = path_to_group.split("/")
     tokens_entry = convert_data_converter_dict_to_nxdl_path(path_of_entry).split("/")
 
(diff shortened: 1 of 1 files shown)


pynxtools/dataconverter/helpers.py
```diff
@@ -164,9 +164,9 @@ def generate_template_from_nxdl(
         ):
             template[optionality][f"{path}/@units"] = None
 
-        parent_path = convert_data_converter_dict_to_nxdl_path(path.rsplit("/", 1)[0])
+        nxdl_path = convert_data_converter_dict_to_nxdl_path(path)
         list_of_children_to_add = get_all_defined_required_children(
-            parent_path, nxdl_name
+            nxdl_path, nxdl_name
```
sherjeelshabih, line 169: I'm a bit lost. Can you provide me a way to quickly test this or describe what the prob was. 

The changes above make sense. But then I would've kept the parent_path here and passed that to add_inherited_children because we want to add all other inherited children of the parent of the current field/attribute being processed. Not of the field/attribute itself, no?

domna: The problem here is that this was passing the field `/ENTRY[entry]/definition` as path, which resulted in a `parent_path = /ENTRY` and then I got all of the required children of the entry but I just want to have all required children of `definition`.

That's also how I would expect it, no? I want to have the required children of the actual field not of its parent (this is also how it is done for the group below). Or do I misunderstand how this part of the code is working?

domna: I also sent you some data on how to test this on slack

rettigl:
This fix solves the original issue (Exception being raised), but does not seem to do the right thing either. With this fix, I get these error reports from the convert function:
```
The data entry corresponding to /ENTRY[entry]/@version is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/geometries/arpes_geometry/TRANSFORMATIONS[transformations] is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/transformations/@vector is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/transformations/@transformation_type is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/transformations/@depends_on is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/SAMPLE[sample]/transformations/@vector is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/SAMPLE[sample]/transformations/@transformation_type is required and hasn't been supplied by the reader.
The data entry corresponding to /ENTRY[entry]/SAMPLE[sample]/transformations/@depends_on is required and hasn't been supplied by the reader.
```
It seems like for several other fields now, the field name is removed while the attributes are directly added to the parent group name.

rettigl:
Do we have/can we write unit tests that test for the correct behavior of inheritance in different scenarios?

domna:
Yes, I thought that there is a misconception on my side as I just debugged this based on one key. But this makes sense from the change I introduced. Are you able to provide your files you use for conversion? It might be easier to test on the actual problem.

> Do we have/can we write unit tests that test for the correct behavior of inheritance in different scenarios?

We have unit tests but obviously not for this particular case. Also coverage for it can be improved. I hope to add some additional tests with or after the verification is finished #133.

domna:
I went through it again and I think the fix is fine. However, I forgot to change the path when writing the required keys to the template. I updated it and these errors are gone for me (at least for the NXmpes example) and the NxdlAttributeError is still gone.

@rettigl Could you give it a try once more?

rettigl:
Works now for me as well

rettigl (approved):
While I don't understand the details, it fixes my issue, and does not seem to introduce other errors

PR #257, opened 2024-02-22
Add a --get-required flag to the converter

### LINKED ISSUE #264: Make `generate-template` its own subcommand of `dataconverter`
It would be nice to be able to run `dataconverter generate-template` to separate the normal conversion routine from the generation of the template. First suggested here: https://github.com/FAIRmat-NFDI/pynxtools/pull/257#issuecomment-1959545221

If we have `generate-template` as a subcommand, we can possibly have many more flags, like `required`, `recommended`, `required-in-optional-groups`. Maybe even a `to-json` flag.

sherjeelshabih:
I changed the flag to --required. I think we can use it to enforce more of the conversion routine to just focus on required fields in later versions.

This also has to be used with --generate-template right now. So if you do:

```console
dataconverter --nxdl NXmpes --generate-template --required
```

you will get the required fields.

I'm open to suggestions here. Let me know what can be improved and I'll add to it. 

- [x] Check if optional groups are not slipping through when the --required flag is used.
They are slipped through. Needs another MR as other verification routines can be affected by this.

_Partially_ fixes #264 as well.

### DIFF
--- a/tests/dataconverter/test_convert.py
+++ b/tests/dataconverter/test_convert.py
@@ -117,7 +117,7 @@ def test_cli(caplog, cli_inputs):
     result = runner.invoke(dataconverter.main_cli, cli_inputs)
     if "generate-template" in cli_inputs:
         assert result.exit_code == 0
-        assert '"/ENTRY[entry]/NXODD_name/int_value": "None",' in result.stdout
+        assert '"/ENTRY[entry]/NXODD_name/int_value": null,' in result.stdout
     elif "--input-file" in cli_inputs:
         assert "test_input" in caplog.text
     elif result.exit_code == 2:
--- a/tests/dataconverter/test_convert.py
+++ b/tests/dataconverter/test_convert.py
@@ -104,9 +104,7 @@ def test_get_names_of_all_readers():
 @pytest.mark.parametrize(
     "cli_inputs",
     [
-        pytest.param(
-            ["--nxdl", "NXtest", "--generate-template"], id="generate-template"
-        ),
+        pytest.param(["generate-template", "--nxdl", "NXtest"], id="generate-template"),
         pytest.param([], id="nxdl-not-provided"),
         pytest.param(
             ["--nxdl", "NXtest", "--input-file", "test_input"], id="input-file"
@@ -116,8 +114,8 @@ def test_get_names_of_all_readers():
 def test_cli(caplog, cli_inputs):
     """A test for the convert CLI."""
     runner = CliRunner()
-    result = runner.invoke(dataconverter.convert_cli, cli_inputs)
-    if "--generate-template" in cli_inputs:
+    result = runner.invoke(dataconverter.main_cli, cli_inputs)
+    if "generate-template" in cli_inputs:
         assert result.exit_code == 0
         assert '"/ENTRY[entry]/NXODD_name/int_value": "None",' in result.stdout
     elif "--input-file" in cli_inputs:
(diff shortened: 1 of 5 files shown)


domna:
Thanks. I'll test it

I'm also thinking whether it would be a good idea to introduce subcommands to the dataconverter, i.e., doing this as `dataconverter generate-template --required`. That way it's a bit easier to specify which parts allow which flags and it's clear that it's doing something else than the normal conversion routine. What do you think?

sherjeelshabih:
> Thanks. I'll test it
> 
> I'm also thinking whether it would be a good idea to introduce subcommands to the dataconverter, i.e., doing this as `dataconverter generate-template --required`. That way it's a bit easier to specify which parts allow which flags and it's clear that it's doing something else than the normal conversion routine. What do you think?

That actually makes sense. I'll let the other's vote on it. If it's agreed, I'll make a PR for it.

domna:
When I use the `--required` flag it also reports fields which are under optional groups, which we should not output here. I would say that we should only require those which are strictly required.

I think it would be also nice if we would just output the template fields as a list of paths (or have a switch for it / allow directly writing it to json, so people can use it to start a config file).

Do you think it's also possible to print the requiredness alongside the path (which only makes sense when it's printed w/o the required flag of course)?

lukaspie:
> > Thanks. I'll test it
> > I'm also thinking whether it would be a good idea to introduce subcommands to the dataconverter, i.e., doing this as `dataconverter generate-template --required`. That way it's a bit easier to specify which parts allow which flags and it's clear that it's doing something else than the normal conversion routine. What do you think?
> 
> That actually makes sense. I'll let the other's vote on it. If it's agreed, I'll make a PR for it.

+1 on this one.

domna:
Looks very good already. It seems to report the correct fields for mpes if I ignore the required field under an optional group.

Some things I found during testing (which you probably already are aware of but I just wanted to note them here):
- There is still `--generate-template` under the convert subcommand
- `--nxdl` can be set to required for the subcommand as we don't need the params trick there. That way the reader does not error out.

lukaspie:
My suggestion:
- make the json print (what is now called `--json` the default) -> the only difference is the distinction between "None" and null
- add a flag for outputting this json to a json file (so that one can provide a filepath where to store the config template)

tests/dataconverter/test_convert.py
```diff
@@ -116,10 +114,10 @@ def test_get_names_of_all_readers():
 def test_cli(caplog, cli_inputs):
     """A test for the convert CLI."""
     runner = CliRunner()
-    result = runner.invoke(dataconverter.convert_cli, cli_inputs)
-    if "--generate-template" in cli_inputs:
+    result = runner.invoke(dataconverter.main_cli, cli_inputs)
+    if "generate-template" in cli_inputs:
         assert result.exit_code == 0
-        assert '"/ENTRY[entry]/NXODD_name/int_value": "None",' in result.stdout
+        assert '"/ENTRY[entry]/NXODD_name/int_value": null,' in result.stdout
```
sanbrock, line 120: Are all the readers prepared for this change?

sherjeelshabih: These changes only affect the CLI. It doesn't change the objects sent to any reader. So they should all work the same as before.

lukaspie (approved):
LGTM, thanks for addressing all requested ideas. Before merging, could you please have a look at the README changes that they reflect what the dataconverter can do right now?

PR #263, opened 2024-02-26
Updates ASE package in pyproject.toml

sherjeelshabih:
Nomad wants to update the ASE package now. This is to help build the dependency file in the Nomad repo and to merge when everything there works.

- [ ] Check if this branch still works with the updated Nomad branch


sherjeelshabih:
The error we see here would most likely be gone with @mkuehbach's changes for EM and APT. We can wait for the relevant PRs to be merged or we can disable the imports temporarily, if there is an urgency from @ladinesa.

pyproject.toml
mkuehbach, line 1: The issue is related to the fact that the ifes library was heavily refactored and updated, given that >=0.0.9 it will pull since today the newest version, with function names that are no longer the same as they are by the still not updated apm reader. My proposal is:
set ifes_apt_tc_data_modeling==0.0.9 and merge this, then the parsers will still work as they are as both refactorings for em and apm have not yet been merged.

The merge will bump ifes_apt_tc_data_modeling
Immediately after the merge 
The em and apm readers will be made into plugins and with this all remaining em_* readers and the apm reader in the pynxtools main repo and with this also the ifes dependency will be removed

mkuehbach (changes requested):

pyproject.toml
sherjeelshabih: Thank you. This helps. I'll switch to ==0.0.9 and let the CI do the testing.

sherjeelshabih: I guess now ifes_apt_... at 0.0.9 is limited to Python3.11.  Let me ask Alvin how quick he needs this resolved.

mkuehbach: Wait I recall this was an issue some time ago lemme check the version history I think there is also one for 3.12

sherjeelshabih: Alright. Thanks. Also, Alvin says this can wait. So no rush.

mkuehbach: v0.1 released that constraint and was used at the time I think by @domna in a deployment,
so use ==0.1 refactoring function names was done after that release

sherjeelshabih:
Before this gets merged we should check we can install this with Nomad.

lukaspie (approved):
NOMAD CI passes as far as I can see, so this should be good to go.

lukaspie:
Could you update the Nomad MR (https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1700) to the latest commit in this PR here before merging here? Just to see that the installation still works.

mkuehbach (approved):
For now setting ifes==0.1 hard is okay, so this PR is LGTM to me.
With the pynxtools-apm plugin addition we have to go to >=0.2...

PR #265, opened 2024-02-26
Add pynxtools-xps as extra install

lukaspie:
Following the initial release of pynxtools at https://pypi.org/project/pynxtools-xps/, allow for installing with pynxtools as an extra install.


lukaspie:
Currently breaking tests because of `ifes_act_tc_data_modeling==0.9` in pyproject.toml, see #263.

Issue #221, opened 2024-01-25
Always add definitions version


domna:
Currently, we add a version string function but actually writing the nexus version has to be done by each reader.
We should add a general way to add the definition and its version to a file independent of the reader.

This is how it's done for ellipsometry:
https://github.com/FAIRmat-NFDI/pynxtools/blob/ac19484528443129ae947b492141c15918395365/pynxtools/dataconverter/readers/ellips/reader.py#L467-L472

This should be agnostic of the reader and should be done for every entry.

sherjeelshabih:
We are not locking down the NXDL version with the reader here, right? We'll just provide a convenient function for readers to use to get the current definition version being used. If that is the case, why shouldn't we just use the conversion routine to verify it passes with the current version being used and just add it there? It means less code in the reader that has to be updated if we change the versioning function, etc.

sherjeelshabih:
Then your point is to allow reader's to have the freedom to place this arbitary version in any spot in the template they like. I can see why this helps. 

What do you think about having readers not fill the version but just do it auto, like we do with a warning rn, in the converter? If the version will always be the one used by dataconverter, then we're good. I don't want to add a burden of responsibility for people dealing with reader's for something trivial.

PR #267, opened 2024-03-07
Fix version retrieval for Ellipsometry reader

domna:
This fixes the version retrieval for the ellipsometry reader.
Writing of the program and version retrieval should however eventually be handled by pynxtools itself outside the readers.
The only problem is that it has to be attached to each entry.


sanbrock (approved):
LGTM

pynxtools/dataconverter/readers/ellips/reader.py
```diff
@@ -466,7 +468,10 @@ def read(
         )
         template["/ENTRY[entry]/definition/@version"] = get_nexus_version()
         template["/ENTRY[entry]/program_name"] = "pynxtools"
-        template["/ENTRY[entry]/program_name/@version"] = version("pynxtools")
+        try:
```
sanbrock, line 471: what about pynxtools.get_version() which would always give back something sensible

domna: Yes, this would be preferred and I also thought to generalise it here. However, we wanted to write these metadata outside the specific readers anyways, because pynx already knows everything from its own settings and CLI arguments. That's why I kept this out of here for now but eventually the whole version handling should be moved out of the reader. #221 is the related issue.

Issue #283, opened 2024-03-14
Pluginization of EM, APM, Ellips, XRD


mkuehbach:

domna:
Duplicate of #251

PR #284, opened 2024-03-14
Remove apm dataconverter as its functionality are superseeded by the …

mkuehbach:
…pynxtools-apm plugin, implementing a contribution for #283


mkuehbach:
Working on the pluginization of the apm dataconverter it turned out that here ifes_apt_tc_data_modeling library in version >=0.2 is needed while here ifes_apt_tc_data_modeling<0.2 which generates a conflict when installing pynxtools as the dependency for the pynxtools-apm locally. Therefore, and as it is needed anyway soon, we should remove the apm dataconverter here.

End users should consult pynxtools<=0.1.1 if they still want to use the apm parser for now.
Developers and users should follow on pynxtools-apm and comment and raise issues there for atom probe specific parsing into NeXus and NOMAD Oasis.

pyproject.toml
```diff
@@ -33,7 +33,6 @@ dependencies = [
     "ase>=3.19.0",
     "flatdict>=4.0.1",
     "hyperspy>=1.7.4",
-    "ifes_apt_tc_data_modeling<0.2,>=0.0.9",
```
domna, line 36: Is this the only dependency which was _only_ used by the apm reader?
All of the specific dependencies can be removed here now and should be added to `pyproject.toml` of the apm reader.

pyproject.toml
domna, line 1: Please also add apm as an extra install, see the mpes, xps, stm extras for an example

domna (changes requested):
Please add `apm` as an extra in `pyproject.toml` and check if there are no dependencies left which are only needed by apm.

Otherwise, everything looks good

sanbrock (approved):
LGTM

pyproject.toml
mkuehbach: Thank you @domna, yeah this also my understanding. Out of the top of my head this was indeed the only specifically-apm-used deps. If I find in during the pluginization there is one more I will remove them also.
Right now all other deps are mainly required by the em readers but with pynxtools-em also many of these will be removed

pyproject.toml
mkuehbach: Implemented

domna (approved):

PR #285, opened 2024-03-18
Removes transmission and hall readers

domna:
This removes the transmission and hall readers as they are managed by nomad plugins now.

Fyi: @aalbino2, @ka-sarthak (could you post the links of the respective plugins here so the development can be followed?)


aalbino2:
LGTM

sherjeelshabih (approved):

PR #286, opened 2024-03-18
Remove em-specific dataconverters as their functionality is superseed…

mkuehbach:
…ed by the pynxtools-em plugin, implementing a contribution for #283


mkuehbach:
Associated with this PR we should check all of us if the following hefty deps are really required for pynxtools:
- scipy
- pandas

Yeah they come in with many python packages right now but if somebody uses a bare installation of pynxtools with just one tool why would one need a complete scipy and or pandas installation? Point is the dependencies of the base pynxtools should be as small as possible.

mkuehbach:
When merged, I create another PR for merging the mmov_em_apm_announcement and with this closing the issue pluginization of apm and em as far as the base pynxtools is concerned

domna:
Could you merge this to your other branch and then create a new PR on the main branch? That way the CI/CD runs (it only runs for PRs targeting the main branch). We can then review the other PR, as far as I see it currently only contains some doc changes.

PR #289, opened 2024-03-19
Remove em-specific code that is becoming replaced by pynxtools-em

mkuehbach:


domna (changes requested):
The removal of files look all good to me.
I went through the file and checked which imports we actually use and I found that we don't use these packages anymore:
- flatdict
- gitpython
- zipfile37
- tzlocal
- scipy

I don't know whether these packages were used by you or if they were removed previously. However, I think we can savely remove them from the dependencies now.

There is also a `shared` folder in `dataconverter/readers` which is not used anymore. Shall we remove this as well. Otherwise I would suggest we move the functions in `dataconverter/utils.py`. If this will be removed we can also remove the `pytz` dependency in #285

mkuehbach:
Getting consistent code for managing time stamps and zone rigorously may demand using the pytz library, in one case of em it even required tzlocal but we should give it a test, so please remove pytz via #285, right the shared directory I missed I will also remove it now

domna (approved):
Thanks LGTM now

PR #291, opened 2024-03-20
Including xrd as a plugin.

RubelMozumder:


domna (changes requested):
Please also add `xrd` to the `convert` extra

RubelMozumder:
> Please also add `xrd` to the `convert` extra

Sorry for missing that. I will add it.

domna (approved):
LGTM

PR #294, opened 2024-03-25
Count readers onee if readers from reader package and pynxtools are c…

RubelMozumder:
Count readers once.

### DIFF
--- a/pynxtools/dataconverter/convert.py
+++ b/pynxtools/dataconverter/convert.py
@@ -94,7 +94,7 @@ def get_names_of_all_readers() -> List[str]:
             all_readers.append(
                 file[index_of_readers_folder_name:index_of_last_path_sep]
             )
-    plugins = list(map(lambda ep: ep.name, entry_points(group="pynxtools.reader")))
+    plugins = list(set(map(lambda ep: ep.name, entry_points(group="pynxtools.reader"))))
     return sorted(all_readers + plugins)
 
 
--- a/pynxtools/dataconverter/convert.py
+++ b/pynxtools/dataconverter/convert.py
@@ -94,8 +94,8 @@ def get_names_of_all_readers() -> List[str]:
             all_readers.append(
                 file[index_of_readers_folder_name:index_of_last_path_sep]
             )
-    plugins = list(set(map(lambda ep: ep.name, entry_points(group="pynxtools.reader"))))
-    return sorted(all_readers + plugins)
+    plugins = list(map(lambda ep: ep.name, entry_points(group="pynxtools.reader")))
+    return list(set(sorted(all_readers + plugins)))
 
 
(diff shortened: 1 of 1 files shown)


sherjeelshabih (commented):
I wouldn't want to have duplicate readers. But I guess in the case where we do have two of the same named ones, showing the name(s) once is alright.

These changes are fine for me. I'll let @domna also have his comment. @domna , feel free to approve when you're happy.

domna:
Actually, I came across the same problem and I decided not to do this fix.
If people have multiple readers it just goes unnoticed and they might not be aware that they are actually using another reader as they expect, which might have unforeseen consequences (wrongly produced data etc). So I would at least do an additional warning here or even error out as multiple readers of the same name should not be allowed as this should not happen too often and can simply be solved by uninstalling one of the reader plugins.

sherjeelshabih:
> Actually, I came across the same problem and I decided not to do this fix. If people have multiple readers it just goes unnoticed and they might not be aware that they are actually using another reader as they expect, which might have unforeseen consequences (wrongly produced data etc). So I would at least do an additional warning here or even error out as multiple readers of the same name should not be allowed as this should not happen too often and can simply be solved by uninstalling one of the reader plugins.

I also have the same feeling for this. The code that picks up our readers has a clear priority that it will take the first one it sees. That can be annoying to debug as well. Then I will say let's throw an error at the very top of any invokation of our pipeline. So top level import/cli run. That way the user knows something is wrong in their setup. A warning might not be enough? What do you guys think?

rettigl:
> I also have the same feeling for this. The code that picks up our readers has a clear priority that it will take the first one it sees. That can be annoying to debug as well. Then I will say let's throw an error at the very top of any invokation of our pipeline. So top level import/cli run. That way the user knows something is wrong in their setup. A warning might not be enough? What do you guys think?

I agree, but it would be good to include some hints on how to fix the issue in the error message.

domna:
I agree with both of your points, I think we should

- Emit an error message when we have readers with the same name
- Explain that this is due to incompatible plugins and also display their package names. We can then tell the user that they need to uninstall one of them for pynx to work again

RubelMozumder:
> I agree with both of your points, I think we should
> 
> * Emit an error message when we have readers with the same name
> * Explain that this is due to incompatible plugins and also display their package names. We can then tell the user that they need to uninstall one of them for pynx to work again


I also see the value of the error message. Actually, the duplication was coming from the third part plugin which has another version of `nomad-lab` as a dependency. And that `nomad-lab` is the older one that brings the `pynxtools` reader in a subdirectory. So, in the error message, we can add this as a sort of possible reason for error.

RubelMozumder:
It is resolved by @domna in another PR.

PR #299, opened 2024-03-27
Update manifest file

domna:
As @fekad pointed out we include `docs` and `tests` directory in our wheel file. This should not be the case.

This PR updates the `MANIFEST.in` file to explicitly include only the correct folder and files.


lukaspie (approved):
LGTM

PR #301, opened 2024-03-28
Apply and pin ruff==0.3.4

domna:
This pins ruff so we don't introduce arbitrary reformatting in other PRs depending on the ruff version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.1.7
+    rev: v0.3.4
     hooks:
       # Run the linter.
       - id: ruff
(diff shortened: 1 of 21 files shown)


domna:
Considerations:
- nomad uses `ruff==0.1.8`. Should we also use this version?
- nomad has some additional formatting settings. Should we align?

Since we are not submoduling or anything it's fine if we are not fully aligned with nomad. My suggestion would be: We pin `ruff==0.3.4` and copy the additional formatting settings from nomad. I would maybe just change to double quote autoformatting. I don't know why nomad uses single quotes as typically python autoformatting does double quotes (also black).

This would be the relevant changes from nomads `pyproject.toml`:
```
# Same as Black.
line-length = 88
indent-width = 4

[tool.ruff.format]
# use single quotes for strings.
quote-style = "single"

# indent with spaces, rather than tabs.
indent-style = "space"

# Like Black, respect magic trailing commas.
skip-magic-trailing-comma = false

# Like Black, automatically detect the appropriate line ending.
line-ending = "auto"
```

My suggestion would just take it as is but remove the `quote-style = single`

sherjeelshabih (approved):
:pray:

PR #303, opened 2024-03-30
Nexus file parsing refactoring

domna:
A general note on the code: It is _really_ hard to understand what is happening in this code part and I'm not sure whether there are other problems because it's hard to get a general overview. I have a few recommendations on general refactoring:
- Introduce an nxdl object which holds its own inheritance chain and properties (doc, requiredness, etc). From it docstrings and parent elements can be deduced at any time. This can also be easily and more consistently translated to nomad metainfo.
- The nxdl_path then just is a list of these objects
- Docstring generation is then a separate process of traversing these objects. This avoids the logging/path generation concurrency

### DIFF
--- a/pynxtools/nexus/nexus.py
+++ b/pynxtools/nexus/nexus.py
@@ -2,8 +2,8 @@
 """Read files from different format and print it in a standard NeXus format"""
 
 import logging
-import os
 import sys
+from typing import Union
 
 import click
 import h5py
@@ -216,12 +216,14 @@ def get_nxdl_doc(hdf_info, logger, doc, attr=False):
     """Get nxdl documentation for an HDF5 node (or its attribute)"""
     hdf_node = hdf_info["hdf_node"]
     # new way: retrieve multiple inherited base classes
-    (class_path, nxdl_path, elist) = get_inherited_hdf_nodes(
+    class_path, nxdl_path, elist = get_inherited_hdf_nodes(
         nx_name=get_nxdl_entry(hdf_info),
         hdf_node=hdf_node,
         hdf_path=hdf_info["hdf_path"] if "hdf_path" in hdf_info else None,
         hdf_root=hdf_info["hdf_root"] if "hdf_root" in hdf_info else None,
     )
+    # Copy the nxdl_path, otherwise the cached object is altered
+    nxdl_path = nxdl_path.copy()
     elem = elist[0] if class_path and elist else None
     if doc:
         logger.debug("classpath: " + str(class_path))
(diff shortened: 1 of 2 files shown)


pynxtools/nexus/nexus.py
```diff
@@ -216,12 +216,14 @@ def get_nxdl_doc(hdf_info, logger, doc, attr=False):
...
     # new way: retrieve multiple inherited base classes
-    (class_path, nxdl_path, elist) = get_inherited_hdf_nodes(
+    class_path, nxdl_path, elist = get_inherited_hdf_nodes(
         nx_name=get_nxdl_entry(hdf_info),
         hdf_node=hdf_node,
         hdf_path=hdf_info["hdf_path"] if "hdf_path" in hdf_info else None,
         hdf_root=hdf_info["hdf_root"] if "hdf_root" in hdf_info else None,
     )
+    # Copy the nxdl_path, otherwise the cached object is altered
+    nxdl_path = nxdl_path.copy()
```
sanbrock, line 226: also covert by another PR

pynxtools/nexus/nexus.py
```diff
@@ -625,122 +628,99 @@ def hdf_node_to_self_concept_path(hdf_info, logger):
...
-        self, logger, nexus_file, d_inq_nd=None, c_inq_nd=None, is_in_memory_file=False
-    ):
+    def __init__(self, logger, nexus_file, document=None, concept_path=None):
         self.logger = logger
-        local_dir = os.path.abspath(os.path.dirname(__file__))
 
-        self.input_file_name = (
-            nexus_file
-            if nexus_file is not None
-            else os.path.join(local_dir, "../../tests/data/nexus/201805_WSe2_arpes.nxs")
```
sanbrock, line 641: default option is to run it as an example on the provided test file

pynxtools/nexus/nexus.py
```diff
@@ -625,122 +628,99 @@ def hdf_node_to_self_concept_path(hdf_info, logger):
...
+        self.concept_path = concept_path
         # Aggregating hdf path corresponds to concept query node
-        self.hdf_path_list_for_c_inq_nd = []
+        self.hdf_path_list_for_concept = []
 
     def visit_node(self, hdf_name, hdf_node):
         """Function called by h5py that iterates on each node of hdf5file.
         It allows h5py visititems function to visit nodes."""
-        if self.d_inq_nd is None and self.c_inq_nd is None:
-            process_node(hdf_node, "/" + hdf_name, self.parser, self.logger)
```
sanbrock, line 655: default option is to annotate the whole file

PR #304, opened 2024-03-30
Fix nexus parsing for NOMAD

domna:
See related issue https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1731

- One problem was related to a caching issue, i.e., `lru_cache` always returns the same object if it has been cached already. If this object is altered, the altered object gets returned for subsequent calls of the lru_cached function. See this stackoverflow question for more information: https://stackoverflow.com/questions/54909357/how-to-get-functools-lru-cache-to-return-new-instances. I simply avoid this now by making a copy after the function call. Maybe we'd need to investigate for other objects of this function call and/or in other places if we have the same problem there
- A second problem originated in the `get_nxdl_attr_doc` function which added attribute nodes twice. This was due to the fact that _all_ attributes which are part of the inheritance chain were added instead of only the lowest child. Maybe this problem also exists for fields.

I made some remarks on the general code quality in #303


sanbrock (approved):
LGTM

sanbrock (approved):
LGTM

PR #307, opened 2024-04-08
Update definitions

domna:


lukaspie (approved):
LGTM, XPS example are also still working.

PR #310, opened 2024-04-17
Adds partial capital names as renamable in template generation

mkuehbach:
- [x] The culprit path_... function was also used in a test that call I changed but there are now two tests failing
- [x] Two tests are failing

### DIFF
--- a/tests/dataconverter/test_helpers.py
+++ b/tests/dataconverter/test_helpers.py
@@ -272,7 +272,7 @@ TEMPLATE["optional"]["/@default"] = "Some NXroot attribute"
                 TEMPLATE, "/ENTRY[my_entry]/NXODD_name/bool_value", "required"
             ),
             (
-                "The data entry corresponding to /ENTRY[entry]/NXODD_name/bool_value is"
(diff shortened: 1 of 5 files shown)


domna:
I think it would be good if we update the example template here, too: https://github.com/FAIRmat-NFDI/pynxtools/blob/3faa83523c924f339b4542b219b28a7c021ce570/tests/dataconverter/test_helpers.py#L150

This means changing `NXODD_name` to `NXODD_name[nxodd_name]` in the template keys _and_ in the respective tests.

tests/dataconverter/test_helpers.py
```diff
@@ -272,7 +272,7 @@ def fixture_filled_test_data(template, tmp_path):
                 TEMPLATE, "/ENTRY[my_entry]/NXODD_name/bool_value", "required"
```
sanbrock, line 272: Shouldn't we change this and all the above Template entries with NXODDxxx?

sherjeelshabih: That's how the template generation produces it for us, right? Then we should have this as NXODD_name[nxodd_name] everywhere.

sherjeelshabih (approved):

PR #312, opened 2024-04-17
Write definition and version directly independently from the reader

### LINKED ISSUE #221: Always add definitions version

### LINKED ISSUE #311: The written definition should be checked against the `--nxdl`
Currently, we can write a different `/entry/definition` into the file as we supply via the `--nxdl` attribute. This should be checked and ideally this should also be handled directly independently from the reader, so that they don't have to write the definition and version themself.

This also relates to #221

domna:
Fixes #311
Fixes #221

### DIFF
--- a/pynxtools/dataconverter/convert.py
+++ b/pynxtools/dataconverter/convert.py
@@ -423,15 +423,20 @@ def convert_cli(
             "The --input-file option is deprecated. Please use the positional arguments instead."
         )
 
-    convert(
-        tuple(file_list) + input_file,
-        reader,
-        nxdl,
-        output,
-        fair,
-        undocumented,
-        skip_verify,
-    )
+    try:
+        convert(
+            tuple(file_list) + input_file,
+            reader,
+            nxdl,
+            output,
+            fair,
+            undocumented,
+            skip_verify,
+        )
+    except FileNotFoundError as exc:
+        raise click.BadParameter(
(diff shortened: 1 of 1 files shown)


pynxtools/dataconverter/helpers.py
```diff
@@ -203,6 +205,15 @@ def generate_template_from_nxdl(
         generate_template_from_nxdl(child, template, path, nxdl_root, nxdl_name)
 
 
+def get_all_entry_names(template: Template) -> Set[str]:
+    entry_names = set()
+    for key in template:
+        entry_name_match = re.search(r"\/ENTRY\[([a-zA-Z0-9_\.]+)\]", key)
+        if entry_name_match is not None:
+            entry_names.add(entry_name_match.group(1))
+    return entry_names
```
sherjeelshabih, line 214: Can we move this to the Template class instead? Makes sense to have it there.

domna: Sure. good suggestion, I'll move it

PR #313, opened 2024-04-18
Fixes enumeration check

domna:


pynxtools/dataconverter/helpers.py
```diff
@@ -629,7 +627,7 @@ def validate_data_dict(template, data, nxdl_root: ET.Element):
                 data[path] = is_valid_data_field(data[path], nxdl_type, path)
                 is_valid_enum, enums = is_value_valid_element_of_enum(data[path], elem)
```
sanbrock, line 628: This seems to check only the given nxdl elem which concept the data element directly belongs to. Note that the enum list may have been inherited. In case of enums, one should check the first elem in the returned elist which contains an enum list rather than checking the absolut first elem.

pynxtools/dataconverter/helpers.py
```diff
@@ -264,14 +264,12 @@ def convert_data_dict_path_to_hdf5_path(path) -> str:
...
-def is_value_valid_element_of_enum(value, elem) -> Tuple[bool, list]:
+def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
     """Checks whether a value has to be specific from the NXDL enumeration and returns options."""
-    if elem is not None:
-        has_enums, enums = nexus.get_enums(elem)
-        if has_enums and (
-            isinstance(value, list) or value not in enums[0:-1] or value == ""
-        ):
-            return False, enums
+    for elem in elist:
```
sanbrock, line 269: We may want to check if elist is None or an empty list

sanbrock (changes requested):

pynxtools/dataconverter/helpers.py
domna: I checked this already and elist is always a list. If there are not elements is just the empty list.

We also use it like this here:
https://github.com/FAIRmat-NFDI/pynxtools/blob/48314117452e21fd3e8003f220a0d9556fd350ab/pynxtools/dataconverter/helpers.py#L110

sanbrock (approved):
LGTM

PR #314, opened 2024-04-18
Correct validation for required fields in variadic groups

domna:
# Fixes

- Correct validation of repeating groups inside the template
- Correct validation of variadic groups with an optional parent

# Additions

This adds a `Collector` class for the validation to collect and log validation problems in one central place. This can be used for a verifier.

## Initial problem
Since merging of #310 I get the following error with the mpes example:

```
LookupError: The data entry, /ENTRY[entry]/INSTRUMENT[instrument]/sourceTYPE[source_probe]/probe,
has an optional parent, /ENTRY[entry]/INSTRUMENT[instrument]/sourceTYPE[sourcetype],
with required children set.
Either provide no children for /ENTRY[entry]/INSTRUMENT[instrument]/sourceTYPE[sourcetype] 
or provide all required ones.
```

This PR attempts to fix the issue and adds tests for the particular issue


domna:
@sherjeelshabih The actual problem here is fixed. However, the test do strange things but it's probably that I did something wrong. When I test it with the mpes example it all works as expected now.

domna:
We also have a second problem here:
https://github.com/FAIRmat-NFDI/pynxtools/blob/10b1c442d97647bf082d143bfca80d587ecd9118/pynxtools/dataconverter/helpers.py#L514-L516

The `renamed_path` only contains the generic path. So if we only write a variation of this in the dict and the generic one is still contained in the dict it reports it as required but missing even though a variation is present.

sherjeelshabih:
I like this idea. Please give it a try. We also have this func: https://github.com/FAIRmat-NFDI/pynxtools/blob/10b1c442d97647bf082d143bfca80d587ecd9118/pynxtools/dataconverter/helpers.py#L541

If somehow we need a complete Template to correctly place such a field, this func can help. But it's another loop so it will be better to avoid this.

sherjeelshabih:
I like this idea. Please give it a try. We also have this func: https://github.com/FAIRmat-NFDI/pynxtools/blob/10b1c442d97647bf082d143bfca80d587ecd9118/pynxtools/dataconverter/helpers.py#L541

If somehow we need a complete Template to correctly place such a field, this func can help. But it's another loop so it will be better to avoid this.

tests/dataconverter/test_helpers.py
```diff
@@ -365,23 +468,23 @@ def fixture_filled_test_data(template, tmp_path):
...
             alter_dict(
                 remove_from_dict(
                     TEMPLATE, "/ENTRY[my_entry]/required_group/description"
                 ),
-                "/ENTRY[my_entry]/required_group",
+                "/ENTRY[entry]/required_group",
                 None,
             ),
-            "The required group, /ENTRY[entry]/required_group, hasn't been supplied.",
+            "The required group, /ENTRY/required_group, hasn't been supplied.",
```
sanbrock, line 487: Should not this report follow consequently either 
- Template style (so we keep [entry]) - I would vote for that, or  
- the actually missing hdf group (so /entry/...) - problem is that the missing item may have a variadic concept name, or 
- the NeXus concept (so we add the name of the AppDef, too, like /NXtest/ENTRY/...)

sanbrock (changes requested):

tests/dataconverter/test_helpers.py
domna: > Should not this report follow consequently either
> 
> * Template style (so we keep [entry]) - I would vote for that, or

Yes, but it seemed a bit odd to me that we report the generic names if the full concept is missing. But we can certainly change it back to `/ENTRY[entry]/...`. I just did it this way because it was easier from the variadic groups where I only have a concept name (or multiple entries). It also seemed to make more sense to me.
> * the actually missing hdf group (so /entry/...) - problem is that the missing item may have a variadic concept name, or

In case we know in which group we are operating we are reporting this specific group. This results in different error logs here, containing either `/ENTRY[my_entry]/...` or `/ENTRY[entry]/...` based on that.
> * the NeXus concept (so we add the name of the AppDef, too, like /NXtest/ENTRY/...)

Possible as well but normally users know that they used the reader with this particular appdef (as specified by `--nxdl`) so it seems a bit redundant to me.


Generally, all three options are possible from a technical perspective. We just have to decide which one it shall be.

domna: I changed it now such that we use `/ENTRY[entry]/...` generic paths everywhere if the concept is variadic or unnamed (i.e. not present).

sanbrock (approved):
LGTM

PR #323, opened 2024-05-03
Refactor internal representation of nxdl's to tree structure

domna:
The idea is to keep a dict/Template and use the tree as a generic nxdl verification to check against

Here is a testing snippet of the tree generation function:

```python
from pynxtools.dataconverter.helpers import generate_tree_from_nxdl
from pynxtools.dataconverter.convert import get_nxdl_root_and_path

from anytree import RenderTree, Resolver

root, _ = get_nxdl_root_and_path("NXmpes")
tree = generate_tree_from_nxdl(root)
print(RenderTree(tree))
```
This prints a tree structure of `NXmpes` to the console:
```
ROOT[nxmpes] (req)
└── ENTRY[entry] (req)
    ├── definition (req)
    │   └── @version (req)
    ├── title (req)
    ├── start_time (req)
    ├── end_time (rec)
    ├── method (rec)
    ├── COORDINATE_SYSTEM_SET[geometries] (opt)
    ├── USER[user] (rec)
    │   ├── name (req)
    │   └── affiliation (req)
    ├── INSTRUMENT[instrument] (req)
    │   ├── RESOLUTION[energy_resolution] (opt)
    │   │   ├── physical_quantity (req)
    │   │   ├── type (rec)
...
```

We can also resolve concept paths:
```python
resolver = Resolver("name")
print(resolver.get(tree, "ENTRY/INSTRUMENT/MANIPULATOR/temperature_sensor/name"))
```

## Example verification

This is a working example with NXtest. You can add, remove or alter fields and should see some validation problems being reported _if_ the change is invalid for NXtest. This can be done with any appdef.
```python
from pynxtools.dataconverter.validation import validate_dict_against

valid_dataset = {
    "/entry/program_name": "My program",
    "/entry/definition": "NXtest",
    "/entry/definition/@version": 10,
    "/entry/odd_name/bool_value": True,
    "/entry/odd_name/bool_value/@units": "",
    "/entry/odd_name/int_value": 5,
    "/entry/odd_name/int_value/@units": "m",
    "/entry/odd_name/posint_value": 5,
    "/entry/odd_name/posint_value/@units": "m",
    "/entry/odd_name/char_value": "My char value",
    "/entry/odd_name/char_value/@units": "",
    "/entry/odd_name/date_value": "2024-05-07T15:57:00+01:00",
    "/entry/odd_name/date_value/@units": "",
    "/entry/odd_name/type": "1st type",
    "/entry/required_group/description": "Some description",
    "/entry/required_group2/description": "Some description",
}
validate_dict_against("NXtest", valid_dataset)
```

## TODOs
- [x] Add support for dimensions
- [x] Add support for enumerations
- [x] Add support for choices
- [x] Add dict path traversal and verification function (split, namefit, verify subtree)
- [x] Test unit literals against nxdlTypes.xsd
- [x] Add required traversal function (aka find if anything is missing in the provided dict)
- [x] Check if fields or attributes are present in one of the parent nodes
- [x] Support for `CLASSNAME[path]` notation
- [x] Complete testing
- [x] Complete docstrings for new functions and methods (especially `NexusNode` and subclasses!)
- [x] Replace template verification with tree validation in the actual conversion code
- [x] Add `CLASSNAME[path]` resolution to only write to the respective class to resolve ambiguity
- [x] Follow links and check if it is compatible with the current node

### Will likely not be included in this PR but additional feature PR(s)
- Verification against an h5py tree
- Support for symbols
- Checking of dimensions
- Checking of occurences
- Documentation of the tree structure and the general nxdl rules (two separate pages)
- Use `NexusNode` for the nexus parsing routines
- Build metainfo from `NexusNode` (dynamically!? or add feature for adding all sub-fields)
- ~Extend support for `CLASSNAME[path]` notation!? I would argue we don't need it anymore (because it's better to namefit the file -> This is also what is done when the file is read, e.g., `sourceTYPE` vs. `SOURCE` can only be resolved by namefitting) but it might be useful to speed up things or to resolve ambiguities~

### Ideas
- From the tree, documentation can be easily generated (also with dynamic parent/child resolution). We could easily build a documentation based on mkdocs, which includes the features we are missing (docstrings of parents, list of inherited fields)


domna:
Python 3.8 is failing because `typing.Annotated` was only added in python 3.9:
https://docs.python.org/3.9/library/typing.html#relevant-peps

It's used in pydantic to check the range of int values. Here I use it to constrain the allowed values of the min/max occurs to $\ge0$. We can either remove the check or support for py3.8

domna:
@FAIRmat-NFDI/areab There are still a few things needed but it's already in a reviewable state.

Missing things:
- Complete testing (this I would consider necessary for merging)
- Currently, it is a separate code from the template verification. This needs to be switched before merging
- Support for new features (dimensions, symbols, occurrences, ...) -> Probably best to handle this in a separate PR to progress with a new pynx release

lukaspie (approved):
LGTM

PR #328, opened 2024-05-06
Make `get_names_of_all_readers` a set

domna:
I think this was proposed before by @RubelMozumder and we decided against.
However, it is problematic again but I'm still not sure whether this will solve the problem.

See https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/2000


domna:
Will be handled by a guard clause inside nomad https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1825. So I'm closing this again

PR #329, opened 2024-05-06
Add plugins and links to individual docs to the main docs

lukaspie:
- [x] Add plugins and links to individual docs to the main/reference
- [x] Remove plugins from the main readme to keep single source principle


mkuehbach (approved):

mkuehbach:
Pipeline errors have to be fixed, otherwise agree with the content in this PR

PR #330, opened 2024-05-13
Update NeXus definitions

lukaspie:
Should only be merged if there are no immediate PRs on nexus-definitions that we want to merge as well.


domna (approved):
From my side fine. If there are other PRs we also can update it once more

PR #335, opened 2024-05-28
Small changes to plugin tests

lukaspie:


pynxtools/testing/nexus_conversion.py
```diff
@@ -94,15 +95,15 @@ def convert_to_nexus(self, ignore_undocumented=False):
             template=Template(template), file_paths=tuple(input_files)
         )
 
+        entry_names = read_data.get_all_entry_names()  # type: ignore[attr-defined]
```
lukaspie, line 98: @RubelMozumder I added this here because this is [part of the regular nexus converter](https://github.com/FAIRmat-NFDI/pynxtools/blob/c8e3c8b45e3cb5a738ea54b38a821cd62a667916/pynxtools/dataconverter/convert.py#L169) (so that every nexus file file gets the version by default) and thus it makes sense to add this also during the testing.

pynxtools/testing/nexus_conversion.py
```diff
@@ -94,15 +95,15 @@ def convert_to_nexus(self, ignore_undocumented=False):
...
-        with self.caplog.at_level("ERROR", "WARNING"):
-            _ = validate_dict_against(
-                self.nxdl, read_data, ignore_undocumented=ignore_undocumented
-            )
-        assert not self.caplog.records, "Validation is not successful. Check logs."
-        for record in self.caplog.records:
-            if record.levelname == "ERROR":
-                assert False, record.message
+
+        with self.caplog.at_level(logging.WARNING):
```
lukaspie, line 104: This introduces some minor changes to the assert, so that the plugin test is in line with how we test the core pynxtools in https://github.com/FAIRmat-NFDI/pynxtools/blob/71020b564b88e33a8e50770315b6cfa63ffea31e/tests/dataconverter/test_helpers.py#L529

pynxtools/testing/nexus_conversion.py
RubelMozumder: I would say
to replace lines 92-100 add the code 
```python
    data = transfer_data_into_template(
        input_file=input_file,
        reader=reader,
        nxdl_name=nxdl,
        nxdl_root=nxdl_root,
        skip_verify=skip_verify,
        **kwargs,
    )
``` 

 from https://github.com/FAIRmat-NFDI/pynxtools/blob/c8e3c8b45e3cb5a738ea54b38a821cd62a667916/pynxtools/dataconverter/convert.py#L221C1-L228C6. Because, we should not right the same code different place, but need to reuse them . The benefit is in the future we can change in a single place and automatically reused everywhere.

RubelMozumder (changes requested):

pynxtools/testing/nexus_conversion.py
RubelMozumder: It does not grep any error only ẁarning` not the `error`. Isn't it good to replace the code on the `test_helper.py` not to skip any errors, or?

RubelMozumder (changes requested):

pynxtools/testing/nexus_conversion.py
lukaspie: It will also catch any errors. Any log messages with a severity level of logging.WARNING or higher (i.e., WARNING, ERROR, CRITICAL) will be captured by caplog if you use `with caplog.at_level(logging.WARNING)`. So, errors are already included at this level.

pynxtools/testing/nexus_conversion.py
lukaspie: I agree, this makes the most sense. I have implemented it here. In that case, the `reader ` passed to the ReaderTest object must however not be the class, but the name of the reader. This is because within `transfer_data_into_template`, the `get_reader` function is called. See my changes to the documentation how this influences the implementation.

RubelMozumder (changes requested):
A few modifications are needed. LGTM!

.github/workflows/plugin_test.yaml
```diff
@@ -11,104 +11,61 @@ on:
...
+          #   tests_to_run: tests/.
+          # - plugin: pynxtools-em
+          #   branch: main
+          #   tests_to_run: tests/.
+          # - plugin: pynxtools-mpes
+          #   branch: main
+          #   tests_to_run: tests/.
+          - plugin: pynxtools-stm
+            branch: use-new-testing
+            tests_to_run: tests/test_reader.py::test_stm_reader
```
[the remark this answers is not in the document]
lukaspie: Ok

RubelMozumder (approved):

PR #338, opened 2024-06-03
Adds newlines to the list of steps

sherjeelshabih:
Just a quick fix to add two new lines. 

It used to be like this: 
image

It looks like this now: 
image


lukaspie (approved):
LGTM

PR #340, opened 2024-06-05
Align with NOMAD docs

lukaspie:


domna (approved):
LGTM. I'm however not sure if we want to keep the vanilla NOMAD logo. It looks a bit like that this is the central documentation for NOMAD. This might confuse people and it might be good if we had a small `pynxtools` below or beside it or so. That way it would be clear that it belongs to nomad but is actually a different documentation.

Also it always confuses me that clicking on the logo brings me to the official nomad page. I'm used to landing on the entry page of the site I'm currently at. But it's the same for NOMAD itself so it's no real issue arising from this PR.

lukaspie:
> LGTM. I'm however not sure if we want to keep the vanilla NOMAD logo. It looks a bit like that this is the central documentation for NOMAD. This might confuse people and it might be good if we had a small `pynxtools` below or beside it or so. That way it would be clear that it belongs to nomad but is actually a different documentation.
> 
> Also it always confuses me that clicking on the logo brings me to the official nomad page. I'm used to landing on the entry page of the site I'm currently at. But it's the same for NOMAD itself so it's no real issue arising from this PR.

100% agree on both points. I guess the idea with the vanilla NOMAD logo is that it all points back to NOMAD, but pynx is in principle also standalone. Let's just keep it for now, I saw some discussion in discord about the particular UX when clicking on the button, so maybe this will be changed anyway.

PR #344, opened 2024-06-06
Switch to src layout, add default data to package

### LINKED ISSUE #300: Switch to `src` layout
There is a discussion going on that using a `src` layout is better for enabling proper testing. See this article: https://hynek.me/articles/testing-packaging/

I think it also might make sense to switch to a `src` layout in pynxtools and maybe also do this for our plugin template.

lukaspie:
Switching to a src layout is better for enabling proper testing.
The default data for the example reader is now shipped with the package instead of being parsed from the tests/data folder.

This also updates the definitions since I had to re-init the submodule.

Fixes #300


tests/nexus/test_nexus.py
```diff
@@ -146,7 +146,7 @@ def test_get_node_at_nxdl_path():
 
     nxdl_file_path = os.path.join(
         local_dir,
-        "../../pynxtools/definitions/contributed_definitions/NXiv_temp.nxdl.xml",
+        "../../src/pynxtools/definitions/contributed_definitions/NXiv_temp.nxdl.xml",
```
sanbrock, line 149: Some test data files are in tests/data which are now also copied to src/pynxtools/data/
Should not we just have one place for them and pick from up from there?

domna:
I checked locally, too and everything was fine. I tried both editable and "normal" install.

domna (approved):

RubelMozumder (commented):
LGTM!

tests/nexus/test_nexus.py
lukaspie: The reason why I moved them into the repo is that if they are not there, the example reader cannot be instantiated when you pip install pynxtools (since we don't include the tests/data in the build). Thus, with the new src layout, it cannot find it when running the tests because we actually test the repo as it would be installed. However, in the test function defined in tests/, we also load these files. 

I removed these files from the tests/dat and now point the test_functions to src/pynxtools/data/.

PR #345, opened 2024-06-06
Add workflow for checking against NOMAD dependencies

### LINKED ISSUE #85: Check against nomad dependencies
When we change dependencies it would be nice to check compatibility with nomad with ci/cd. We could just download the requirements files from nomad and check with them the current pyproject dependencies.

lukaspie:
This is a draft for how we can make sure that the dependencies in the pyproject.toml file are in line with the NOMAD dependencies.

Strategy:
~~- [x] pip-compile pyproject.toml to a requirements file~~
~~- checkout NOMAD requirements from GitLab~~
~~- Combine the two requirements file and try to install the combined file~~
EDIT:
- check out NOMAD@develop and init all its submodules
- in nomad/pyproject.toml, replace the pynxtools[convert]==<version> by pynxtools[convert]@git+https://github.com/FAIRmat-NFDI/pynxtools.git@<this-current-branch> to check if the current branch makes any problematic changes
- use NOMADS `uv pip compile` to generate the requirements-(dev).txt
- try to install these

Possible improvements:
- [x] only run this workflow if the pyproject.toml has actually changed
- [x] return possibly resolving dependencies

Fixes #85

### DIFF
--- a/.github/workflows/nomad-requirements.yml
+++ b/.github/workflows/nomad-requirements.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Checkout NOMAD from GitLab
       run: |
         git clone https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git nomad
-        cd nomad-FAIR
+        cd nomad
         git checkout develop 
 
     - name: Set up Python
--- a/.github/workflows/nomad-requirements.yml
+++ b/.github/workflows/nomad-requirements.yml
@@ -17,12 +17,11 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Checkout NOMAD from GitLab
-      with:
-        submodules: 'recursive'
       run: |
         git clone https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git nomad
         cd nomad
         git checkout develop
+        git submodule update --init --recursive
     
     - name: Replace pynxtools dependency in NOMAD pyproject.toml
       run: |
--- a/.github/workflows/nomad-requirements.yml
+++ b/.github/workflows/nomad-requirements.yml
@@ -1,0 +1,58 @@
+name: validate NOMAD dependencies compatibility
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  validate_nomad_dependencies:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout pynxtools
+      uses: actions/checkout@v2
+
+    - name: Checkout NOMAD from GitLab
+      run: |
+        git clone https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git nomad
+        cd nomad-FAIR
+        git checkout develop 
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'  # or any other version you need
+
+    - name: Parse pynxtools pyproject.toml
+      id: parse_pyproject
+      run: |
+        python -c "
+          import toml
+          import json# Load the pyproject.toml
+        
+          with open('pyproject.toml', 'r') as file:
+              pyproject = toml.load(file)
+        
+          dependencies = pyproject.get('project', {}).get('dependencies', [])
+        
+          # Convert to requirements format and save
+          with open('pyproject-requirements.txt', 'w') as f:
+              for dep in dependencies:
+                  f.write(f'{dep}\n')"
+    
+    - name: Combine and Check Dependencies
+      run: |
+        # Combine repo1's requirements with repo2's dependencies from pyproject.toml
+        cat nomad/requirements-dev > combined-requirements.txt
+        echo "" >> combined-requirements.txt
+        cat pyproject-requirements.txt >> combined-requirements.txt
+
+        # Upgrade pip and install the combined dependencies
+        python -m pip install --upgrade pip
+        pip install -r combined-requirements.txt
+      env:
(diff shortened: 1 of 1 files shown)


domna:
It might make sense to use [uv](https://astral.sh/blog/uv) instead of `pip-compile`. I think nomad is switching to it and it is way faster. This might benefit this ci/cd as well.

.github/workflows/nomad-requirements.yml
```diff
@@ -0,0 +1,50 @@
...
+        
+    - name: Combine and Check Dependencies
+      run: |
+        # Combine NOMADS requirements with pynxtools dependencies from pyproject.toml
+        cat nomad/requirements-dev.txt > combined-requirements.txt
+        echo "" >> combined-requirements.txt
+        cat pyproject-requirements.txt >> combined-requirements.txt
+
+        # Install the combined dependencies
+        pip install -r combined-requirements.txt
```
domna, line 48: I think it would make more sense to run nomads pip-compile (or uv) here. Because this is the part which might be failing. Just installing the combined requirements can install other dependencies by packages which are installed later

lukaspie:
> It might make sense to use [uv](https://astral.sh/blog/uv) instead of `pip-compile`. I think nomad is switching to it and it is way faster. This might benefit this ci/cd as well.

I have a PR (https://github.com/FAIRmat-NFDI/pynxtools/pull/347) to use uv across the whole repository for compiling and installing requirements.

.github/workflows/nomad-requirements.yml
lukaspie: Please have a look at the updated workflow. I switched it such that
- it checks out NOMAD@develop and inits all its submodules
- in nomad/pyproject.toml, it replaces the `pynxtools[convert]==<version>` by `pynxtools[convert]@git+https://github.com/FAIRmat-NFDI/pynxtools.git@<this-current-branch>` to check if the current branch makes any problematic changes
- it then uses NOMADS `uv pip compile` to generate the requirements-(dev).txt
- and then it tries to install them

I checked and the pynxtools version actually does get replaced properly.

.github/workflows/nomad-requirements.yml
```diff
@@ -0,0 +1,56 @@
...
+  validate_dependencies:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout pynxtools
+      uses: actions/checkout@v2
+
+    - name: Checkout NOMAD from GitLab
+      run: |
+        git clone https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git nomad
```
domna, line 25: You could also add `--depth 1` here, that way the ci/cd would only checkout nomad shallowly (faster).

.github/workflows/nomad-requirements.yml
```diff
@@ -0,0 +1,56 @@
...
+        cd nomad
+        git checkout develop
+        git submodule update --init --recursive
```
domna, line 28: Also `--depth 1`

.github/workflows/nomad-requirements.yml
domna: Also you could use `--branch develop` here and `--recurse-submodules` then you don't have to add the lines below

.github/workflows/nomad-requirements.yml
```diff
@@ -0,0 +1,56 @@
...
+      run: |
+        curl -LsSf https://astral.sh/uv/install.sh | sh 
+
+    - name: Generate (dev-)requirements.txt from modified pyproject.toml
+      working-directory: ./nomad
+      run: |
+        uv pip compile -p 3.9 --annotation-style=line --extra=infrastructure --extra=parsing --output-file=requirements.txt dependencies/nomad-dos-fingerprints/pyproject.toml dependencies/parsers/eelsdb/pyproject.toml pyproject.toml
+        uv pip compile -p 3.9 --annotation-style=line --extra=dev --extra=infrastructure --extra=parsing --output-file=requirements-dev.txt requirements.txt pyproject.toml
+
+    - name: Install NOMAD dependencies with pynxtools from current branch
```
domna, line 50: I think strictly we don't really need to install it in the end. The most important part is that we can generate the dependencies and we might not have much benefit installing it without testing it.

.github/workflows/nomad-requirements.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+    # Run workflow only when there are changes in pyproject.toml or dev-requirements.txt
+    # paths:
+    # - 'pyproject.toml'
```
domna, line 12: From my point of view it would make sense to activate these. Or are there still some problems when using these lines?

domna (approved):
LGTM

I left a few comments for cleanup

.github/workflows/nomad-requirements.yml
lukaspie: changed

.github/workflows/nomad-requirements.yml
lukaspie: changed

.github/workflows/nomad-requirements.yml
lukaspie: I left it here because maybe we will want to do something else in this nomad workflow (like checking that nomad parses nexus data properly with any changes to pynxtools).

.github/workflows/nomad-requirements.yml
lukaspie: I intentionally kept this open because it wouldn't run the CI/CD in this branch. Now that everything is fixed, I have reactivated them.

PR #347, opened 2024-06-08
Switch to uv compile and install

lukaspie:
- [x] Replace pip-tools by [uv](https://github.com/astral-sh/uv) like NOMAD did [here](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1843#f4fb69164b97e1b6f17984369c9c382573a3f919)
- [X] Use `uv pip install` to speed up the CI/CD (note that you need to pass the `--system` flag because the default of uv is that you work in a virtual environment) -> this reduces the runtime of the pytest.yaml CI/CD by 25%

As of python 3.12, `setuptools` is not anymore installed by default with python. If you don't use pip (which still has setuptools), you need to specify `setuptools` as a dependency in the pyproject.toml file. Since uv is used in the CI/CD workflows now (which [doesn't come with a build backend](https://github.com/astral-sh/uv/issues/3957)), `setuptools` was added to the dependencies.


.github/workflows/build_docs.yml
```diff
@@ -34,8 +34,8 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install ".[docs]"
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          uv pip install --system ".[docs]"
```
domna, line 38: I think it might be easier to just use `pip install`. I think `uv pip install` is for the cases in which you want to use the venv created by uv. But in ci we use just the native env anyways (which is also totally fine). Or was there another reason for you to do this?

lukaspie: It's a bit faster if you use `uv pip install` (also with the download of uv with `curl`), but the difference is not that big. If you use the `--system` flag, it installs into the native env (because by default it expects a venv). For me, both ways is fine.

domna: I see. Then let us just keep it the way it is. I was just a bit surprised by the redundancy (as uv is also in the dev dependencies)

PR #348, opened 2024-06-12
Fix requirements workflow

domna:
This should fix the failing ci in master.


lukaspie (approved):
LGTM, I did indeed not check this. Thanks for the fix!

PR #350, opened 2024-06-18
Validation fixes

domna:
# Fixes

- An issue when entries are search by naming convention (i.e., `ENTRY` for `NXentry`). Previously, it ignored if the found group has an actual name. This lead to wrong groups being returned.
- Attaching sub-groups correctly in their parent concepts, i.e., `my_data(NXdata)` is contains `DATA(NXdata)` if the latter is part of a super-concept of the current appdef
- Validation support if a specialised concept is given for for a required variadic group
- Pins `numpy<2.0.0`

### DIFF
--- a/src/pynxtools/dataconverter/nexus_tree.py
+++ b/src/pynxtools/dataconverter/nexus_tree.py
@@ -145,11 +145,21 @@ class NexusNode(NodeMixin):
     parent_of: List["NexusNode"]
 
     def _set_optionality(self):
+        """
+        Sets the optionality of the current node
+        if `recommended`, `required` or `optional` is set.
+        Also sets the field to optional if `maxOccurs == 0` or to required
+        if `maxOccurs > 0`.
+        """
         if not self.inheritance:
             return
         if self.inheritance[0].attrib.get("recommended"):
             self.optionality = "recommended"
-        elif self.inheritance[0].attrib.get("required"):
+        elif self.inheritance[0].attrib.get("required") or (
+            isinstance(self, NexusGroup)
+            and self.occurrence_limits[0] is not None
+            and self.occurrence_limits[0] > 0
+        ):
             self.optionality = "required"
         elif self.inheritance[0].attrib.get("optional") or (
             isinstance(self, NexusGroup) and self.occurrence_limits[0] == 0
@@ -181,6 +191,9 @@ class NexusNode(NodeMixin):
         self.parent_of = []
 
     def _construct_inheritance_chain_from_parent(self):
+        """
+        Builds the inheritance chain of the current node based on the parent node.
+        """
         if self.parent is None:
             return
         for xml_elem in self.parent.inheritance:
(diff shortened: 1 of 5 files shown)


domna (commented):
@RonHildebrandt: Can you check if the warning is gone with this fix?

domna:
@RonHildebrandt Everything should be fixed now. If you like you can give it another test. But I checked with your ellipsometry example files.

tests/dataconverter/test_validation.py
```diff
@@ -21,46 +21,49 @@
...
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/int_value": 2,
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/int_value/@units": "eV",
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/posint_value": np.array(
+            [1, 2, 3], dtype=np.int8
+        ),
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/posint_value/@units": "kg",
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/char_value": "just chars",
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/char_value/@units": "",
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/type": "2nd type",
+        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/date_value": "2022-01-22T12:14:12.05018+00:00",
```
RubelMozumder, line 44: The key with `NXODD_name[nxodd_name]` should give an error and the error can be catch in test. Because, only accepted convention is `NXODD_NAME[nxodd_name]`, the given name should  be inside the square bracket. Or I am missing something.

RubelMozumder (changes requested):
Some of the function does not have any doc. Having a doc (even without the explanation of the function parameters) would be very helpful for future modification and development. Other than that I do not see any other big issue with this PR.

src/pynxtools/dataconverter/nexus_tree.py
```diff
@@ -489,6 +546,37 @@ class NexusGroup(NexusNode):
         Optional[int],
     ] = (None, None)
 
+    def _check_sibling_namefit(self):
+        if not self.variadic:
+            return
```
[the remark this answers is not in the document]
domna: I will add it

tests/dataconverter/test_validation.py
domna: I don't understand your comment. They key `/ENTRY[my_entry]/NXODD_name[nxodd_name]/date_value` should be correct as any key which fits `NXODD_name` is (i.e.,`*_name`)

RubelMozumder: I meant we do not write any `/ENTRY[my_entry]/NXODD_name[nxodd_name]/date_value` but rather `/ENTRY[my_entry]/NXODD_NAME[nxodd_name]/date_value`. If we modify or overwrite the name of the group that goes inside the square bracket. Or?

domna: Yes exactly, this notation means that we write a path `/my_entry/nxodd_name/date_value` to the file. The parts outside the brakets are just there to fit to a certain concept. It should also work with `/ENTRY[my_entry]/DATA[nxodd_name]/date_value` as `nxodd_name` will be an `NXdata` eventually.

src/pynxtools/dataconverter/validation.py
```diff
@@ -328,13 +328,24 @@ def check_nxdata():
 
     def handle_group(node: NexusGroup, keys: Mapping[str, Any], prev_path: str):
         variants = get_variations_of(node, keys)
```
RubelMozumder, line 330: If you would a short doc here also that would be great as well. Because, the name `handle_group` has really a broad meaning. Other than that I think all are good.

RubelMozumder (approved):
LGTM!

src/pynxtools/dataconverter/validation.py
domna: In this case I'm opting against this. These are all just inline groups. Their purpose is already to make the code more readable by chunking it into smaller functions and nobody can call this functions from outside. Commenting all of them would make this function even longer + we have to see if we keep it this way or use a validation which works directly on the hdf nodes (I already started to work on this in #333)

PR #351, opened 2024-06-18
Upgrade numpy and ruff

domna:
Recently, numpy 2.0.0 has been released which brings some breaking changes. This adapts to the changes and upgrades ruff to `0.4.8`, which provides correct linting for the new numpy.

See https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#numpy-2-migration-guide for reference


lukaspie (approved):

lukaspie:
CI/CD is working again, so we can merge this. I guess for the repos where we are using `pint`, we still have the version conflict, right? So we cannot just put `"numpy>=1.21.2"` there.

domna:
> CI/CD is working again, so we can merge this. I guess for the repos where we are using `pint`, we still have the version conflict, right? So we cannot just put `"numpy>=1.21.2"` there.

Yes, I think the problem with pint is that it theoretically allows for numpy 2 (in pint 0.17 I think) but it doesn't work with it, so we have to constrain pint in this cases. Since we don't use pint here I would not enforce the version here.

PR #358, opened 2024-07-01
Updates sibling namefittings

domna:
The sibling namefitting was still unreliable. This PR should fix it.

### DIFF
--- a/src/pynxtools/dataconverter/nexus_tree.py
+++ b/src/pynxtools/dataconverter/nexus_tree.py
@@ -406,7 +406,13 @@ class NexusNode(NodeMixin):
         """
         name = xml_elem.attrib.get("name")
         inheritance_chain = [xml_elem]
-        for elem in self.inheritance:
+        inheritance = iter(self.inheritance)
+        for elem in inheritance:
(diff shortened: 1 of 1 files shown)


src/pynxtools/dataconverter/nexus_tree.py
```diff
@@ -245,6 +245,9 @@ def search_child_with_name(
                 return self.add_inherited_node(name)
         return None
 
+    def get_children_for(self, xml_elem: ET._Element) -> Optional[ET._Element]:
+        return next((x for x in self.children if x.inheritance[0] == xml_elem), None)
```
lukaspie, line 249: I guess this could benefit from a docstring.

lukaspie (approved):
LGTM, maybe some docs might be helpful for future maintenance.

src/pynxtools/dataconverter/nexus_tree.py
domna: Ah yes, I forgot to write it. Thanks I'll update it

domna: Added

PR #359, opened 2024-07-01
Pynxtools as nomad plugin

### LINKED ISSUE #354: Move all pynxtools related tools from NOMAD to pynxtools
- [ ] Move `nexus.py`
- [ ] Move `metainfo/nexus.py`
- [ ] Update in NOMAD
- [ ] Check if everything works

### LINKED ISSUE #355: Make pynxtools a NOMAD parser plugin
https://nomad-lab.eu/prod/v1/staging/docs/howto/plugins/parsers.html

- [x] Adapt to type system first
- [ ] Move code out of NOMAD first
- [ ] Follow the above documentation to create a parser plugin for NOMAD

domna:
This PR enables pynxtools to work as nomad plugin for

- [x] The metainfo schema
- [x] The parsing of nexus files

Fixes #355
Fixes #354


sanbrock (approved):

lukaspie:
We should probaby update the README / docs to reflect that this is now also a full NOMAD plugin.

.github/workflows/pytest.yml
```diff
@@ -30,9 +30,15 @@ jobs:
         run: |
           curl -LsSf https://astral.sh/uv/install.sh | sh
           uv pip install --system coverage coveralls
-      - name: Install package
+      - name: Install pynx without nomad plugin
+        if: "${{ matrix.python_version == '3.8' }}"
```
lukaspie, line 34: We could also drop support for py3.8. but it's fine for now.

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -123,12 +126,12 @@ def check_reproducibility_of_nexus(self):
...
-                )
+                # skip ignored lines (mainly version conflicts)
+                for ignore_line in IGNORE_LINES:
+                    if gen_l.startswith(ignore_line) and ref_l.startswith(ignore_line):
+                        break
+                else:
+                    assert False, (
+                        f"Log files are different at line {ind}"
+                        f" generated: {gen_l} \n referenced : {ref_l}"
+                    )
```
lukaspie, line 137: Just as an idea, it would be also cool to have the nomad parsing of the individual test files tested with this `testing` module.

lukaspie (approved):
LGTM

.github/workflows/pytest.yml
domna: Yes, I would only drop it when it's end of life or we have some explicit code in pynx which renders it incompatible. I was hesitant to generally drop it because of it being a plugin.

src/pynxtools/testing/nexus_conversion.py
domna: That's a nice idea. Let me see if I can quickly come up with something for this PR

domna: I added something pretty rudimentary. But it can be improved in the future in another PR

PR #360, opened 2024-07-01
Small typo and docstring fixes

domna:
- Adds missing docstrings for `is_a` and `parent_of` relationships
- Fixes a typo and residual code in sibling retrieval
- Renames a function to correctly represent what it does

### DIFF
--- a/src/pynxtools/dataconverter/nexus_tree.py
+++ b/src/pynxtools/dataconverter/nexus_tree.py
@@ -228,6 +228,20 @@ class NexusNode(NodeMixin):
     def search_add_child_for_multiple(
         self, names: Tuple[str, ...]
     ) -> Optional["NexusNode"]:
+        """
+        Searchs and adds a child with one of the names in `names` to the current node.
+        This calls `search_add_child_for` repeatedly until a child is found.
+        The found child is then returned.
+
+        Args:
+            name (Tuple[str, ...]):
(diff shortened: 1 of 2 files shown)


lukaspie:
@domna shall I test this somehow with one of my own files? Or are you sure it's (still) working as intended?

domna:
> @domna shall I test this somehow with one of my own files? Or are you sure it's (still) working as intended?

For the part changed here I'm sure that it does what it should. However, while testing this I found another bug and I would also bring the fix here (which only appears sometimes... so I wasn't quite aware from the start)

lukaspie:
> > @domna shall I test this somehow with one of my own files? Or are you sure it's (still) working as intended?
> 
> For the part changed here I'm sure that it does what it should. However, while testing this I found another bug and I would also bring the fix here (which only appears sometimes... so I wasn't quite aware from the start)



> > @domna shall I test this somehow with one of my own files? Or are you sure it's (still) working as intended?
> 
> For the part changed here I'm sure that it does what it should. However, while testing this I found another bug and I would also bring the fix here (which only appears sometimes... so I wasn't quite aware from the start)

Alright, then just let me know once you have that fix and then I can test it.

domna:
> Alright, then just let me know once you have that fix and then I can test it.

Should be fixed now

lukaspie (approved):
LGTM

PR #362, opened 2024-07-02
Bugfix: Nomad parsing, atom_types mapping on chemical_formula and use of chemical_symbols

mkuehbach:


src/pynxtools/dataconverter/helpers.py
```diff
@@ -810,13 +810,13 @@ def check_for_valid_atom_types(atoms: Union[str, list]):
 
     if isinstance(atoms, list):
         for elm in atoms:
-            if elm not in chemical_symbols:
+            if elm not in chemical_symbols[1::]:
```
domna, line 813:
```suggestion
            if elm not in chemical_symbols[1:]:
```

I also suggest to remove the exception raising and use logging to report that the element was not found.

src/pynxtools/nomad/parser.py
```diff
@@ -383,13 +385,21 @@ def _get_chemical_formulas(self) -> Set[str]:
             if sample.get("atom_types__field") is not None:
                 atom_types = sample.atom_types__field
                 if isinstance(atom_types, list):
-                    atomlist = atom_types
+                    for symbol in atom_types:
+                        if symbol in chemical_symbols[1::]:
```
domna, line 389: I think it would also make sense to emit a warning here that an element was ignored during parsing

src/pynxtools/nomad/parser.py
```diff
@@ -383,13 +385,21 @@ def _get_chemical_formulas(self) -> Set[str]:
...
+                            element_set.add(symbol)
+                    if len(element_set) == 1:
+                        chemical_formulas.add("".join(list(element_set)))
                 # Caution: The element list will be overwritten
                 # in case a single chemical formula is found
-                material.elements = list(set(material.elements) | set(atomlist))
+                material.elements = list(set(material.elements) | element_set)
             if sample.get("chemical_formula__field") is not None:
                 chemical_formulas.add(sample.chemical_formula__field)
```
sanbrock, line 404: The periodic table in NOMAD should be populated from material.elements, so we should not normalise element_set to checmical_formulas because element set does not deal with the ratio of the elements in the formula and can be misinterpreted.

sanbrock (approved):

src/pynxtools/nomad/parser.py
mkuehbach: Agree, implemented

src/pynxtools/nomad/parser.py
mkuehbach: This is what 6d18d6d implemented, in the case that there is really only one element the mapping is possible.
However, I agree with you that even using this special case can be misleading because:
I somebody knows in advance (compared to ones upload to NOMAD) that I populate a field of concept atom_types one agrees that one only indeed lists the elements and thus also accepts that composition information is not recoverable.
Populating the chemical_formula even with a single element can still be factually wrong because NOMAD will pick up on the fact that there is a chemical_formula with in this case one element and concludes that the material is made fully of that element but is not necessarily a correct conclusion that should be NOMADs authority to make.

sanbrock: Exactly. Conclusions, and (respective implementation of) normalisation and visualisation functions in NOMAD shall be fixed, too.

src/pynxtools/dataconverter/helpers.py
mkuehbach: I agree to that in should be captured by the logger, however that is a trickier refactoring that goes beyond the scope of this specific fixing branch. I support that this point should be brought up as an issue that is about "refactor pynxtools nomad nexus parser for using consistent logging and exception handling" key point is: the here mentioned helper function is isolated from the context of the nexus parser class that has a well-defined own logger. Sure one could make a quick solution and pass the logger to the function but I think this is dirty for two reasons: i) this function discussed here has a well-defined task it extracts a string of atom types from a chemical formula if that creates an issue that function should have an own logger but then we have many loggers, the function is used as far as I know only in the pynxtools nexus parser, i.e. therefore it could also well be a class member function of a utility class that gets its own logger that is then connected to the parser logger function,
ii) also there are repetitive checks in several places whether the element is a valid one, I see a point that given the limited usage of the function that it should not at all be required to check if the chemical formula is build from only correct elements, if one could assume that these checks here are even redundant, still of course other issues can happen during the function call and there indeed that function should get a logger, but as I said that function should become a member function of a utility class UNLESS this function is planned to be used outside the pynxtools project for doing this utility task but that execution can again be in a different context with again own different loggers, long short: this is an own issue unrelated to the issue fixed in this pr

domna: You don't need to pass the logger to the function. You can just import `pynxtools.logger` (I'm not sure if this is the exact location) and just use it from there.

We already discussed that we should not raise in pynxtools, because this breaks the parsing of the file completely in nomad, so we should adapt to this here accordingly.

domna: Ah no sorry, we just use normal python logging, see here for an example https://github.com/FAIRmat-NFDI/pynxtools/blob/aec8ab9f070295d38411b666417500bcd2d25a49/src/pynxtools/dataconverter/helpers.py#L45

mkuehbach: implemented

pyproject.toml
```diff
@@ -62,6 +62,10 @@ dev = [
     "uv",
     "pre-commit",
 ]
+nomad_plugin = [
+    "pydantic",
+    "nomad-lab",
+]
```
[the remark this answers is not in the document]
mkuehbach: implemented

PR #363, opened 2024-07-03
Fix bug metainfo.domain should not be dft

mkuehbach:
Currently all parsed content is assigned a nomad Parser's default domain which is 'dft'

For this solution to work we need also a fix in NOMAD specifically in nomad/datamodel/datamodel.py line 866 
domain = Quantity(
which is an enum currently listing 'dft' and 'ems'

'ems' is only used in the EELSDB database and that should be changed to 'em' i.e. electron microscopy, surplus all keywords from the NXDEF_TO_METAINFO_DOMAIN dictionary from this PR need to be added in that Quantities definition enum as well.

### DIFF
--- a/src/pynxtools/nomad/parser.py
+++ b/src/pynxtools/nomad/parser.py
@@ -16,21 +16,21 @@
 # limitations under the License.
 #
 
-from typing import Optional, Set
+from typing import Dict, Optional, Set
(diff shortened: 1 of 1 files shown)


mkuehbach:
Closed for #364

PR #364, opened 2024-07-03
Suggestion for bugfix metainfo.domain no longer only dft and ems

mkuehbach:
Currently all parsed content is assigned a nomad Parser's default domain which is 'dft'

For this solution to work we need also a fix in NOMAD specifically in nomad/datamodel/datamodel.py line 866 
domain = Quantity(
which is an enum currently listing 'dft' and 'ems'

'ems' is only used in the EELSDB database and that should be changed to 'em' i.e. electron microscopy, surplus all keywords from the NXDEF_TO_METAINFO_DOMAIN dictionary from this PR need to be added in that Quantities definition enum as well.


mkuehbach:
Replaces https://github.com/FAIRmat-NFDI/pynxtools/pull/363

src/pynxtools/nomad/parser.py
```diff
@@ -472,6 +484,23 @@ def parse(
...
+        if len(scientific_domain) == 1:
+            # NeXusParser(Parser), domain is member of parser, therein initialized to 'dft'
+            archive.metainfo.domain = scientific_domain[0]
+        else:
+            archive.metainfo.domain = "exp"
+            if len(scientific_domain) > 1:
+                self._logger.warn(
+                    f"cannot resolve domain for multi-domain {scientific_domain}"
+                )
+
```
domna, line 503: I suggest to switch the key and value in the lookup table and directly check it. Also I think it makes sense to explicitly spell out all sub-appdefs as we agreed against the prefixing naming convention (with `NXmpes_arpes` being the exception because of the existence of `NXarpes`).

```suggestion
        if appdef in NXDEF_TO_METAINFO_DOMAIN:
           archive.metainfo.domain = NXDEF_TO_METAINFO_DOMAIN[appdef]
        else:
          archive.metainfo.domain = "exp"
          ... do logging (but I think it's not really necessary as we fall back to a reasonable default)
```

src/pynxtools/nomad/parser.py
```diff
@@ -33,6 +33,18 @@
...
+# a parsed NeXus entry into a specific metainfo.domain other than dft
+NXDEF_TO_METAINFO_DOMAIN = {
+    "em": ["NXem"],
+    "mpes": ["NXmpes", "NXarpes"],
+    "xps": ["NXxps"],
+    "opt": ["NXopt", "NXellipsometry", "NXoptical_spectroscopy", "NXraman"],
+    "apm": ["NXapm"],
+    "spm": ["NXspm", "NXsensor_scan"],
+    "xrd": ["NXxrd"],
+}
```
domna, line 46:
```suggestion
NXDEF_TO_METAINFO_DOMAIN = {
   "NXem": "em",
   "NXmpes": "mpes",
   "NXmpes_arpes":  "mpes",
   "NXarpes": "mpes",
   ...
}
```

domna (changes requested):
I suggest to simplify the lookup a bit. I'm also not sure if we really want to add such a detailed domain, as we also need to always update this in the `domain` property and I'm asking myself if we get any benefit from it. It would be searchable by it but we actually can just search for the appdef name. Maybe it would sense to just set `exp` or `nexus` to denote that the data is coming from a nexus file.

src/pynxtools/nomad/parser.py
mkuehbach: Agree to your proposal of simplifying the lookup, implemented this.

I have a different opinion about the metainfo concepts "entry_type" and "domain".
Currently, the nexus parser has populated entry_type already with the name of the appdef which is useful as it quickly allows to filter via ES for specific measurements following specific appdefs. "Domain" indeed if the mapping table is always each key has only one value then there is not much of an advantage. However, currently we have e.g. the EELSDB whose domain is 'ems' (likely a shorthand for electron microscopy spectroscopy) it is a poor enum value but it has a benefit namely to group entries related to electron microscopy which is not only NXem, domain can be one such tag-like mechanism whereby different entries within the same scientific domain can be filtered for irrespective whether they have been generated using the NeXus parser or e.g. EELS or something maybe from custom base sections within customized parsing. I agree that this is just one tag and thus limited and like with many discussions about suitable tags there will be then people asking for more fine granularizable tags, well its a tradeoff, indeed in what NOMAD was so far (heavily biased towards dft) why to have at all a domain? Now with reaching out to experiments and many different methods I think this is really useful, not only because in fairmat we have not yet been able to agree that even for em there is at least one mechanism where all em related entries can be found by. Domain is one such in my opinion, I agree maybe not an ideal one for everyones particularity interests and expectation how they are using nomad

src/pynxtools/nomad/parser.py
mkuehbach: Implemented as suggested

domna (changes requested):
I would not overcomplicate the situation. The `domain` attribute is mainly deprecated so I would avoid such a large lookup table for better future maintenance.

Let's just write `exp` or `nexus` in the domain regardless of a certain appdef (it will anyways be a nexus domain if it is written by pynxtools)

mkuehbach:
The comments have been addressed, it is mapping to nexus and exp now, I learned that "domain" is mainly deprecated because I happen to discuss with you not because it was somewhere documented transparently.
Key is that the enum in nomad-FAIR gets fixed for either of the suggested solution to work, sth that needs to happen in nomad.

The mapping dictionary is no overcomplication it is a clear expression of my point and a necessity to have things written out when the lookup should be clean as it was suggested.

So, let's forward with this: Feel free to remove the mapping table, I cannot merge this anyway yet as the nomad side is not fixed thus causing the CI to break for well-understood reasons.

domna (approved):
Should pass now as I added the `nexus` domain to the nomad PR

PR #370, opened 2024-07-04
Remove setuptools from dependencies

lukaspie:
- [x] Remove `setuptools` from dependencies
- [x] Replace `setuptools.distutils` with `shutil `in tests
- [x] Remove `_build_wrapper.py` from built package

This currently uses a copy of the `get_vcs_version` function in `_build_wrapper.py` and in `src/pynxtools/__init__.py`, which is probably not ideal.


domna (approved):
LGTM

PR #371, opened 2024-07-04
Update definitions

domna:


domna:
This needs an update for the regression files in xps @lukaspie

lukaspie (approved):
LGTM

PR #373, opened 2024-07-08
Removes unused CLI flags

domna:
This removes the `--documented` and `--fair` flag from the dataconverter CLI. They are not used in the new validation and create wrong warnings _if_ they are used. They are replaced by the new flag `--ignore-undocumented` for the new validation, which simply ignores all undocumented fields.

cc @rettigl

### DIFF
--- a/.github/workflows/plugin_test.yaml
+++ b/.github/workflows/plugin_test.yaml
@@ -61,7 +61,6 @@ jobs:
           repository: FAIRmat-NFDI/${{ matrix.plugin }}
           path: ${{ matrix.plugin }}
           ref: ${{ matrix.branch }}
-          lfs: true
       - name: Install nomad
         run: |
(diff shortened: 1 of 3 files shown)


domna:
The plugin tests for xps and mpes will probably fail because we exceeded our lfs quota.

sherjeelshabih (approved):

domna:
I also added the `--fail` keyword now. If the flag is given it fails raises an error and does not produce a file if validation problems are found. It will also exit with exit code 1 on the shell.

@rettigl: Is it fine like this?

rettigl:
> I also added the `--fail` keyword now. If the flag is given it fails raises an error and does not produce a file if validation problems are found. It will also exit with exit code 1 on the shell.
> 
> @rettigl: Is it fine like this?

Yes, excellent, thanks.

PR #374, opened 2024-07-08
add tests of ellipsometry plugin

RonHildebrandt:
Activating ellipsometry reader tests as the plugin is published.
Idea from @domna 
(lets see if this works out) I just thaught that if this passes cicd, then it everything should be fine?


domna:
> I just thaught that if this passes cicd, then it everything should be fine?

It is. This is just an additional test so we see directly if changes to pynxtools affect any of our readers. Otherwise, we might break something without knowing

RonHildebrandt:
So, this works fine. If the PR is approved, then it can be merged in direclty.

lukaspie (approved):
LGTM

PR #375, opened 2024-07-08
Use log files in testing

### LINKED ISSUE #424: NOMAD tests in testing framework
We should add a test for nomad parsing to the `pynxtools.testing` framework so that the individual plugins can run the `nomad parse` functionality themselves.

lukaspie:
This PR changes the `pynxtools.testing` framework such that it compares the `read_nexus logs` directly during the tests. This avoids having to store binary NeXus files in the repos. Instead, we only commit reference logs, which will lead to much smaller diffs and keep the repositories lighter.

For pynxtools-xps, I have a corresponding branch where this change in the testing is already implemented, including a script to autogenerate these log file consistently: https://github.com/FAIRmat-NFDI/pynxtools-xps/pull/69/files#diff-1ac04a8e7b33c27f6b8eea0d6f44313388de34a688758042f1ee9582f5b34ca2. This branch is also installed in the test_plugins pipeline here.


src/pynxtools/testing/nexus_conversion.py
```diff
@@ -51,8 +53,8 @@ def __init__(self, nxdl, reader_name, files_or_dir, tmp_path, caplog) -> None:
         files_or_dir : str
             List of input files or full path string to the example data directory that contains all the files
             required for running the data conversion through the reader.
-        ref_nexus_file : str
-            Full path string to the reference NeXus file generated from the same
+        ref_log_file : str
+            Full path string to the reference log file generated from the same
```
RubelMozumder, line 57: This is redundant. The log files will come in place of `nexus_file` in `files_or_dir` from the plugins.

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -64,7 +66,7 @@ def __init__(self, nxdl, reader_name, files_or_dir, tmp_path, caplog) -> None:
         self.reader_name = reader_name
         self.reader = get_reader(self.reader_name)
         self.files_or_dir = files_or_dir
-        self.ref_nexus_file = ""
+        self.ref_log_file = ref_log_file
```
RubelMozumder, line 69:
```suggestion
        self.ref_log_file = ""
```

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -86,11 +88,15 @@ def convert_to_nexus(
...
-            0
-        ]
+
+        if not self.ref_log_file:
+            self.ref_log_file = [
+                file for file in example_files if file.endswith(".log")
+            ][0]
+        assert self.ref_log_file, "Reference nexus .log file not found"
+
         input_files = [file for file in example_files if not file.endswith(".nxs")]
```
RubelMozumder, line 98:
```suggestion
        input_files = [file for file in example_files if not file.endswith(".log")]
```

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -86,11 +88,15 @@ def convert_to_nexus(
...
         else:
             example_files = sorted(glob(os.path.join(self.files_or_dir, "*")))
-        self.ref_nexus_file = [file for file in example_files if file.endswith(".nxs")][
-            0
-        ]
+
+        if not self.ref_log_file:
+            self.ref_log_file = [
+                file for file in example_files if file.endswith(".log")
+            ][0]
```
RubelMozumder, line 95:
```suggestion
          self.ref_log_file = [
              file for file in example_files if file.endswith(".log")
          ][0]
```

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -134,10 +140,9 @@ def check_reproducibility_of_nexus(self):
             "DEBUG - value: v",
             "DEBUG - value: https://github.com/FAIRmat-NFDI/nexus_definitions/blob/",
         ]
-        ref_log = get_log_file(self.ref_nexus_file, "ref_nexus.log", self.tmp_path)
         gen_log = get_log_file(self.created_nexus, "gen_nexus.log", self.tmp_path)
```
RubelMozumder, line 143:
```suggestion
        gen_log_file = get_log_file(self.created_nexus, "gen_nexus.log", self.tmp_path)
```
To having the same name convention.

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -134,10 +140,9 @@ def check_reproducibility_of_nexus(self):
...
         with open(gen_log, "r", encoding="utf-8") as gen, open(
```
RubelMozumder, line 144:
```suggestion
        with open(gen_log_file, "r", encoding="utf-8") as gen, open(
```

RubelMozumder (changes requested):

RubelMozumder:
This is a good idea to prevent our plugin from unnecessarily weighing. We also have to replace nexus files by its log files.
Readme file also needs to updated.

src/pynxtools/testing/nexus_conversion.py
lukaspie: I disagree. What if the reader can support two different appdefs, then I need two reference log files for the same data set. I would not like to store the same data twice in two folders, but rather just two log files in the same `files_or_dir`. See [here ](https://github.com/FAIRmat-NFDI/pynxtools-xps/pull/69/files#diff-030d39f7d85e7dd25282dad362350a7a06cd685af9fa186d8ecf4e83c8f6028b)for an example in pynxtools-xps (supporting NXmpes and NXxps).

I guess we could just pass the `ref_log_file ` as part of a list of files to `files_or_dir`, but I would prefer to have the option to pass it directly to the function.

src/pynxtools/testing/nexus_conversion.py
lukaspie: You are right. It should also probably not end with ".nxs" anyway, so I just implemented a check against both.

src/pynxtools/testing/nexus_conversion.py
lukaspie: see above

src/pynxtools/testing/nexus_conversion.py
lukaspie: ok

src/pynxtools/testing/nexus_conversion.py
lukaspie: implemented

src/pynxtools/testing/nexus_conversion.py
RubelMozumder: I can not think such `generic_reader` that will be fit for multiple appdefs. Even it supports multiple `appdef` then `eln`, `config_file` or even raw must be different. 

However, `ref_log_file` can also be added in the list of the files, not in the `dir`. In that case, using multiple `appdef` for the same reader will also be applicable.

The reason, I wanted to reduce the unnecessary function `arg`, which keeps the `test framework` easy to use and intuitive to external users.
But, if you are optimistic about keeping the `ref_log_file` no problem it can be kept.

lukaspie: Look at the link, the data file, `eln`, `config`, and so on are the same for `NXxps` and `NXmpes` (because `NXxps` extends `NXmpes`). For me, having it as an argument is fine, if not needed, one can simply set it to None. As a compromise, I set to None by default, so if you don't give any value for `ref_log_file`, you can simply omit it and the log file is directly parsed from `files_or_dir`.

.github/workflows/plugin_test.yaml
```diff
@@ -39,13 +39,13 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-spm
-            branch: main
+            branch: RepalceNexusFile
             tests_to_run: tests/.
           - plugin: pynxtools-xps
-            branch: main
+            branch: adapt-testing
```
mkuehbach: Still up-to-date

src/pynxtools/nexus/nexus.py
```diff
@@ -425,6 +425,91 @@ def get_inherited_hdf_nodes(
...
+    """
+    # Normalize NumPy scalar and 0D array types
+    if isinstance(value, np.generic):
+        value = value.item()
+    elif isinstance(value, np.ndarray) and value.shape == ():
+        value = value.item()
+
+    def format_float(value: float) -> str:
+        """Format a float deterministically."""
+        if value == 0.0:
```
mkuehbach: Almost always epsilon comparison https://stackoverflow.com/questions/19837576/comparing-floating-point-number-to-zero

But what is the aim to achieve here, why not report all digits?
Indeed often floats hav a lower fractional parts that make two values different but that is a practical irrelevant difference.

Is this the rationale behind this formatting change?
Again then why not render all digits?

src/pynxtools/nexus/nexus.py
```diff
@@ -425,6 +425,91 @@ def get_inherited_hdf_nodes(
...
+    # --- Arrays ---
+    if isinstance(value, np.ndarray):
+        flat = value.flatten()
+        formatted = []
+        for v in flat:
+            if isinstance(v, (np.generic, np.ndarray)):
+                v = v.item()
+            if isinstance(v, float):
+                formatted.append(format_float(v))
+            elif isinstance(v, (int, bool)):
```
mkuehbach: combine branches 475 and 481

tests/nexus/test_nexus.py
```diff
@@ -139,6 +139,74 @@ def test_decode_if_string(string_obj, decode, expected):
         assert result == expected, f"Failed for {string_obj} with decode={decode}"
 
 
+@pytest.mark.parametrize(
+    "value,precision,expected_contains",
+    [
+        # --- Floats ---
+        (0.0, 8, "0.0"),
```
mkuehbach: +0. and -0. would add both

mkuehbach: 0. +- eps would make sense as well

mkuehbach (changes requested):
Reasonable overall to move forward, but treatment of numeric edge cases is discussion-worthy, branch in need of sync with main, CI fixes

.github/workflows/plugin_test.yaml
lukaspie: Because we are testing here if this feature branch can work together with the associated branch in `pynxtools-xps`. Will be rolled back as soon as tests are passing.

mkuehbach: Makes sense, remind rolling back then, ok

tests/nexus/test_nexus.py
mkuehbach: Good

mkuehbach (approved):
Thanks for this addition, can be merged once updated by `main` and CI made passing.

PR #376, opened 2024-07-08
Fix nomad parsing and dataconverter

domna:
This fixes the nomad parsing and some imports in the dataconverter.
The `NexusParser` had a `is_mainfile` function which just returned `True`. This lead to all files being parsed by the `NexusParser`.

### DIFF
--- a/src/pynxtools/nomad/parser.py
+++ b/src/pynxtools/nomad/parser.py
@@ -473,12 +473,9 @@ class NexusParser(MatchingParser):
         self._logger = logger if logger else get_logger(__name__)
         self._clear_class_refs()
 
-        if mainfile.endswith("nxs"):
-            *_, self.nxs_fname = mainfile.rsplit("/", 1)
-            nexus_helper = HandleNexus(logger, mainfile)
-            nexus_helper.process_nexus_master_file(self.__nexus_populate)
-        else:
-            raise ValueError("No Nexus file is found to populate the nexus definition.")
+        *_, self.nxs_fname = mainfile.rsplit("/", 1)
+        nexus_helper = HandleNexus(logger, mainfile)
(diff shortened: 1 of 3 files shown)


src/pynxtools/nomad/parser.py
```diff
@@ -473,12 +473,9 @@ def parse(
         self._logger = logger if logger else get_logger(__name__)
         self._clear_class_refs()
 
-        if mainfile.endswith("nxs"):
-            *_, self.nxs_fname = mainfile.rsplit("/", 1)
-            nexus_helper = HandleNexus(logger, mainfile)
-            nexus_helper.process_nexus_master_file(self.__nexus_populate)
-        else:
-            raise ValueError("No Nexus file is found to populate the nexus definition.")
+        *_, self.nxs_fname = mainfile.rsplit("/", 1)
```
mkuehbach, line 476: I agree that it makes sense to lift the unsubstantiated restriction that only files with NXS ending are accepted, as NeXus content can also be shipped using different formats or any valid HDF5 ending e.g. hdf, hdf5, h5 etc.
In this way now one could even pass here a text file, just the place where this would throw is when the h5py library faces a file that is then not a proper HDF5 file is fine from my perspective

mkuehbach (approved):
One comment apart from this nothing really warranting delays

src/pynxtools/nomad/parser.py
domna: No, we are using a `MatchingParser` now, which only passes the file to this parser if it has a `*.nxs` file ending (we could also have a file type check on hdf5). Basically, this is implemented in the `is_mainfile` function of `MatchingParser` and this is also the reason why I needed to remove this function in this PR.

We could still pass another file if we would manually call this parser, but since this is not for manual use I thought it made sense to remove this check here.

PR #377, opened 2024-07-09
Add raman plugin to `convert` extra and ci/cd testing

RonHildebrandt:
As the Raman Reader as plugin is published, this now shall include it in pynxtools with CI/CD.

Updated pyproject.toml, dev-requirements.txt and activated github workflow for CI/CD

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -45,13 +45,7 @@ identify==2.5.36
 idna==3.7
     # via requests
 importlib-metadata==8.0.0
-    # via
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
domna, line 1: Please revert as this doesn't need to be updated. The plugin dependencies are not included in this requirements file.

RonHildebrandt: arg. okay

domna (approved):
LGTM

PR #378, opened 2024-07-09
Update nomad link in ci/cd

domna:
This changes the link from the old branch to the nomad package to run the tests against.


lukaspie (approved):
LGTM

domna:
@lukaspie The tests are passing now. I also updated the actions to the latest versions and changed the publishing action to use github as a trusted publisher. That way we don't need tokens anymore and it's set in the pynxtools package settings instead in the user settings.

lukaspie (approved):
Looks good. We can do similar changes to the workflows in the plugin at some point.

PR #379, opened 2024-07-11
Update definitions

lukaspie:


.github/workflows/plugin_test.yaml
```diff
@@ -28,13 +28,13 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-mpes
-            branch: main
+            branch: update-to-new-nxmpes
```
domna, line 31: Just a reminder to update this before merging

.github/workflows/plugin_test.yaml
```diff
@@ -28,13 +28,13 @@ jobs:
...
             tests_to_run: tests/.
           - plugin: pynxtools-stm
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-xps
-            branch: main
+            branch: update-configs-after-mpes-changes
```
domna, line 37: This needs to be updated before merging as well

domna (approved):

domna:
Hey @lukaspie, I was thinking whether we could allow multiple branches in the CI/CD. Then it should check whether they exist and if not proceed to the main/next branch. That way it's possible to merge this without merging the failing ci in the readers or in pynx.

lukaspie:
> Hey @lukaspie, I was thinking whether we could allow multiple branches in the CI/CD. Then it should check whether they exist and if not proceed to the main/next branch. That way it's possible to merge this without merging the failing ci in the readers or in pynx.

Yeah, sounds good, I also don't like how we have to do it currently. I will try to cook something up.

.github/workflows/plugin_test.yaml
lukaspie: done

.github/workflows/plugin_test.yaml
lukaspie: done

PR #380, opened 2024-07-15
Add docs for rules in NeXus

lukaspie:


docs/learn/nexus-rules.md
```diff
@@ -0,0 +1,33 @@
+# Rules for storing data in NeXus
+
+!!! danger "Work in progress"
+
+There are several rules which apply for storing single data items in NeXus. There exists a [summary](https://manual.nexusformat.org/datarules.html) in the NeXus documentation outlining most of these rules. However, this explanation is not exhaustive and thus, we have compiled here additional information.
+
+
+## Namefitting
+
+In general, the names of NeXus group and field items are validated according to these boundaries:
```
sanbrock, line 10: L10-22 looks like a copy from the above referenced NeXus documentation page. That would make sense to include references in these cases to the give subsection.
E.g.: NeXus documentation chapter "1.2.3.2. Rules for Storing Data Items in NeXus Files" section "NXDL group and field names" (https://manual.nexusformat.org/datarules.html)

docs/learn/nexus-rules.md
```diff
@@ -0,0 +1,33 @@
...
+- Allowed names
+  - any combination of upper and lower case letter, numbers, underscores and periods, except that periods cannot be at the start or end of the string
+  - This statement is equivalent to matching  this regular expression (named `validItemName` in the [nxdl.xsd](https://github.com/nexusformat/definitions/blob/main/nxdl.xsd)) XML Schema file:
+  ```
+  ^[a-zA-Z0-9_]([a-zA-Z0-9_.]*[a-zA-Z0-9_])?$
+  ```
+- Invalid names:
+  - any name not matching this `validItemName` regular expression
+
+Note that this explicitly also means that it is not allowed to have whitestrings " " in NeXus names.
```
sanbrock, line 23: probably you mean white space, but note that whitespaces include not only space, but also other characters, like tab, new line, etc.

docs/learn/nexus-rules.md
```diff
@@ -0,0 +1,33 @@
...
+
+In NeXus base classes and application definitions, there are two options for defining a concept name. If the group or field in the definition is lowercase, that means that any instance must have the exact same (**fixed**) name. As an example, if there is a field called `my_field` in an application definition, the only allowed name in a file would be `my_field`.
```
sanbrock, line 25: Very important! I think the biggest problem with NeXus documentation that it does not make it clear if the names (or titles) of the concepts defined in base classes and appdef using NXDL can be handled on the same way as the names (or labels) of the data items used in (hdf5) data files, or if there are different rules to be applied to them.

docs/learn/nexus-rules.md
lukaspie: Thanks for noticing. I guess none of these are allowed, so it's fine to just change to white space.

docs/learn/nexus-rules.md
lukaspie: The idea was to state this here explicitly, so people don't have to search across multiple manuals. I think we can rewrite this a bit and make it even more clear, then we can have an individual statement. Note that the page is anyway linked.

docs/learn/nexus-rules.md
```diff
@@ -0,0 +1,33 @@
...
+
+The validation of names is performed by **namefitting**, i.e., fitting the name that is used by the data provider to the name given in the base class / application definitions.
+
+A python implementation of this process can be found in [this function](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/34aed4a74b8d2a682eb0b9292055dc00e5e0220e/dev_tools/utils/nxdl_utils.py#L112). This function returns twice the length for an exact match, otherwise the number of matching characters (case insensitive) or zero, if `name_any` is set to True, is returned. This is also the function that is used in the [validation](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/validation.py) implemented in pynxtools.
```
sanbrock, line 33: Also note some other points:
- name fitting shall not consider UPPERCASE reserved words (like BLUESKY_, DECTRINS_, IDF_, etc.) listed in section "Reserved prefixes".  
- additionally to namefitting, data annotation can use further information, like in case of NXdata where axes listed  among @axes shall fit to AXISNAME and data objects listed in @signal or @auxiliary_signals shall fit to DATA

lukaspie: I added both + the reserved suffixes

sherjeelshabih (approved):
Text is clear and understandable. I like it. :+1:

PR #388, opened 2024-07-23
parse config before postprocess

lukaspie:
This is just to demonstrate how to post process the config dict given multiple entities.


domna (approved):
I think it's a good approach to expose the dict and being able to do post processing on it.

The changes being made for this look all good.

lukaspie:
> I think it's a good approach to expose the dict and being able to do post processing on it.
> 
> The changes being made for this look all good.

Great, I merged it in already now.

PR #389, opened 2024-07-24
Parse a Nexus file as a entry or as a uploaded file.

RubelMozumder:
Prevent overwriting the schema type from NeXus if nexus will be included a subsection into the archive entry.

This PR add new features:
1. If other areas want the nexus file as file and use the file as a reference to their data path, then nexus file will be under the uploads but not as an entry (E.G. XRD use case by area A).
2. If Nexus subsection is added under the other schema, the new modification will prevent to overwrite the schema name from from the nexus app def (e.g. it would not overwrite the ELNXRDdiffraction schema name.).


src/pynxtools/nomad/parser.py
```diff
@@ -487,8 +488,10 @@ def parse(
             if getattr(archive.nexus, var, None) is not None:
                 app_def = var
                 break
-        archive.metadata.entry_type = app_def
-        archive.metadata.domain = "nexus"
+        # If NeXus subsection written under the other schema
+        if not extend_archive and not archive.metadata.entry_type:
```
lukaspie, line 492: Do we need the need argument here? I guess if any other schema sets the `entry_type` by itself, then simply checking for `if not archive.metadata.entry_type` should be sufficient, no? Not sure if there is any case where that's not applicable.

lukaspie (commented):
I honestly don't understand why the nexus subsection shall not be filled. You don't loose anything if it's filled and you can still reference the nexus file as a file. Is there any reason why populating the nexus section is necessarily bad?

RubelMozumder:
> I honestly don't understand why the nexus subsection shall not be filled. You don't loose anything if it's filled and you can still reference the nexus file as a file. Is there any reason why populating the nexus section is necessarily bad?

Nexus subsection will be populated under the other schema whether the nexus file is a nomad entry or just a file, it up to users who wants nexus file as an entry or not. Just giving an option to the users (other areas) what they want.

src/pynxtools/nomad/parser.py
RubelMozumder: I also thought I could check if `entry_type` was filled or not. 

But, I added the argument so that it does not break any in any cases which may be out of my knowledge.

lukaspie: I suggest (in the interest of code maintainability) to **not** add the `extend_archive` parameter and instead just try if the  `not archive.metadata.entry_type` is sufficient. We can always add the parameter in case there is an actual use case for it, but I don't like having it for the sake that it might be needed.

lukaspie:
> > I honestly don't understand why the nexus subsection shall not be filled. You don't loose anything if it's filled and you can still reference the nexus file as a file. Is there any reason why populating the nexus section is necessarily bad?
> 
> Nexus subsection will be populated under the other schema whether the nexus file is a nomad entry or just a file, it up to users who wants nexus file as an entry or not. Just giving an option to the users (other areas) what they want.
> 
> Some say I would refer the data in the nexus file from the archive entry, why I need extra entry there having the same data.

My opinion would be exactly the other way around, you don't gain anything from not having it filled. But okay, we can add the parameter if necessary.

lukaspie (approved):
LGTM

PR #390, opened 2024-07-24
Allow lists of possible values in config file

lukaspie:
This allows to write `"['@attrs:<text>', '@eln', 'no title']"` in the config file, which gets parse into the prefixes as a list of tuples `[('@attrs', 'text'), ('@eln', ''), ('', 'no title')]`. For each of these, it is checked if a value can be returned (whatever that means for the reader solution) and if so, the value is written. If not, "no title" is written. I think this is much cleaner implementation of having these multiple options than before.

I also changed the `parent_key` in `FlattenSettings` to "/ENTRY" to match our previous discussion. Finally, I removed the link handling in the `flatten_json` as this shall be handled by the link callback. Are there any issues with this?

@domna there were some problems with the handling of `!` for the required fields in optional groups while I was implementing this solution. I think it works now, but I didn't have a good example for this, can you check with your example if it still works?

### DIFF
--- a/src/pynxtools/dataconverter/readers/multi/reader.py
+++ b/src/pynxtools/dataconverter/readers/multi/reader.py
@@ -20,6 +20,7 @@
 import logging
 import os
 import re
+import ast
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from pynxtools.dataconverter.readers.base.reader import BaseReader
@@ -121,6 +122,7 @@ def resolve_special_keys(
     value: Any,
     optional_groups_to_remove: List[str],
     callbacks: ParseJsonCallbacks,
+    suppress_warning: bool = False,
 ) -> None:
     def try_convert(value: str) -> Union[str, float, int, bool]:
         """
@@ -137,12 +139,42 @@ def resolve_special_keys(
 
         return value
 
-    if not isinstance(value, str) or ":" not in value:
+    def parse_config_value(value: str) -> Tuple[str, Any]:
+        # Regex pattern to match @prefix:some_string
+        pattern = r"!?(@\w+)(?::(.*))?"
+        match = re.match(pattern, value)
+        if match:
+            prefix, suffix = match.groups()
+            if suffix is None:
+                suffix = ""
+            return (prefix, suffix)
+        else:
+            return ("", value)
+
+    if not isinstance(value, str) or "@" not in str(value):
+        if isinstance(value, str):
+            value = value.lstrip("!")
         new_entry_dict[key] = value
         return
 
-    prefixes = re.findall(r"(@\w+)(\:([^,]+))?", value)
-    for prefix, _, path in prefixes:
+    prefixes: List[Union[Tuple[str, str], str]] = []
+
+    # Regular expression to check if the string contains a list
+    if re.match(r"^\[.*\]$", value):
+        try:
+            # Safely evaluate the string to a list
+            list_value = ast.literal_eval(value)
+            if isinstance(list_value, list):
+                prefixes = [parse_config_value(v) for v in list_value]
+            else:
+                prefixes = [parse_config_value(value)]
+        except (SyntaxError, ValueError):
+            # If the evaluation fails, treat it as a single string
+            prefixes = [parse_config_value(value)]
+    else:
+        prefixes = [parse_config_value(value)]
+
+    for prefix, path in prefixes:
         if not prefix.startswith("@"):
             new_entry_dict[key] = try_convert(path)
             break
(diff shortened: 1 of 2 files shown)


src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -137,12 +139,42 @@ def try_convert(value: str) -> Union[str, float, int, bool]:
...
+            list_value = ast.literal_eval(value)
+            if isinstance(list_value, list):
+                prefixes = [parse_config_value(v) for v in list_value]
+            else:
+                prefixes = [parse_config_value(value)]
+        except (SyntaxError, ValueError):
+            # If the evaluation fails, treat it as a single string
+            prefixes = [parse_config_value(value)]
+    else:
+        prefixes = [parse_config_value(value)]
```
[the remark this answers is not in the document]
domna: Or ideally only using the `re.match`, but I assume it's not enough to ensure that the list parses properly

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -163,10 +195,11 @@ def try_convert(value: str) -> Union[str, float, int, bool]:
 
     if prefixes and new_entry_dict[key] is None:
         del new_entry_dict[key]
-        logger.warning(
-            f"Could not find value for key {key} with value {value}.\n"
-            f"Tried prefixes: {prefixes}."
-        )
+        if not suppress_warning:
+            logger.warning(
```
domna, line 199: Consider setting a level for the warnings and adapt the `logger.level` appropriately. That way you don't have to pass the `suppress_warning` through every function call. However, this is not to critical and I think it's also fine to merge this with the `suppress_warning`

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -343,11 +387,21 @@ def get_processing_order(path: str) -> Tuple[int, Union[str, int]]:
...
+                        self.get_entry_names(),
+                        self.callbacks,
+                        kwargs["suppress_warning"],
+                    )
+                )
+            else:
+                template.update(
+                    fill_from_config(
+                        self.config_dict, self.get_entry_names(), self.callbacks
+                    )
```
domna, line 403: Can be compacted a bit

```suggestion
            template.update(
                fill_from_config(
                    self.config_dict,
                    self.get_entry_names(),
                    self.callbacks,
                    **({'suppress_warning': suppress_warning} if "suppress_warning" in kwargs else {})
                )
            )
```

src/pynxtools/dataconverter/readers/utils.py
```diff
@@ -46,7 +46,7 @@ class FlattenSettings:
     dic: Mapping
     convert_dict: dict
     replace_nested: dict
-    parent_key: str = "/ENTRY[entry]"
+    parent_key: str = "/ENTRY"
```
domna, line 49: Thank you. I missed this value. Can you also change the `Defaults to "/ENTRY[entry]"` in the docstring?

src/pynxtools/dataconverter/readers/multi/reader.py
```diff
@@ -137,12 +139,42 @@ def try_convert(value: str) -> Union[str, float, int, bool]:
...
+        # Regex pattern to match @prefix:some_string
+        pattern = r"!?(@\w+)(?::(.*))?"
+        match = re.match(pattern, value)
+        if match:
+            prefix, suffix = match.groups()
+            if suffix is None:
+                suffix = ""
+            return (prefix, suffix)
+        else:
+            return ("", value)
```
domna, line 152: Can be a bit simpler
```suggestion
        prefixes = re.findall(pattern, value)
        if not prefixes:
            return ("", value)
        return prefixes[0]
```

domna (approved):
Everything works with the mpes example. I made some suggestions to simplify the code.

src/pynxtools/dataconverter/readers/multi/reader.py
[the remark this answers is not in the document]
lukaspie: Thanks, that's cleaner. `re.match` did indeed not work.

src/pynxtools/dataconverter/readers/multi/reader.py
lukaspie: Went with `**{k: v for k, v in kwargs.items() if k == "suppress_warning"}`, I think that's even more concise.

src/pynxtools/dataconverter/readers/multi/reader.py
lukaspie: done

src/pynxtools/dataconverter/readers/utils.py
lukaspie: done

src/pynxtools/dataconverter/readers/multi/reader.py
lukaspie: Yeah, I thought about that, but the idea was that the reader that inherits MultiFormatReader can set that level itself. Wasn't sure how to pass the logger.level adjustments properly through the functions. For the XpsReader, I just want this particular call suppressed, while for the rest of the function, the level shall be at logging.INFO.

PR #392, opened 2024-07-26
Update nexus defs and bump `ruff=0.5.5`

lukaspie:
- [x] I would wait for https://github.com/FAIRmat-NFDI/nexus_definitions/pull/274 to be merged and then update once again. Would be good to have this update in the new release because I am using `NXenergydispersion/radius` now.

**EDIT**: since pynxtools-xrd is not yet ready, we only update to the point right before the changes introduced by https://github.com/FAIRmat-NFDI/nexus_definitions/pull/274. The XRD changes will come with #389.


RubelMozumder (approved):
LGTM!

PR #396, opened 2024-08-02
hook nexus definition to SPM_domain_app_def

RubelMozumder:
Nexus regenerated because:
space difference in root doc from
`'        A generic positioner such as a motor or piezo-electric transducer.  \n'`
 to 
`'         A generic positioner such as a motor or piezo-electric transducer.\n`

in NXpositioner.


lukaspie (approved):
LGTM

Issue #397, opened 2024-08-07
[Bug]: converter failing with is invalid appdef although file not found is culprit


mkuehbach:
### What happened?

Calling pynxtools-em from its top-level directory (i.e. the directory from which src/pynxtools_em are the next subdirectory layers) using the sprint23_em_v3 commit of pynxtools from today and the following call line
```
dataconverter convert examples/eln_data.yaml examples/em.oasis.specific.yaml --reader em --nxdl NXem --output=../dbg/dbg.nxs
```

throws with "NXem is not a valid application definition" although the NXem application definition is found to be alright and the template filling stage works as expected smoothly.

The reason is line 402 https://github.com/FAIRmat-NFDI/pynxtools/blob/a320595a61341893b821704df07311fb614d5993/src/pynxtools/dataconverter/convert.py#L402
locally I hadnt created the directory ../dbg but instead of detecting that this does not exist multiple file do-not-exist cases are bundled into one and all warning for an invalid appdef.

Suggestion: Check for existence of path for output and appdef separately to allow issuing correct and more instructive error messages.

PR #400, opened 2024-08-12
Reraise error on missing file in cli_convert

lukaspie:
Since we already have handlers for missing input, nxdl, and output files, we can simply raise the error again (as `click.BadParameter`) in order to avoid the problem described in #397


lukaspie:
@mkuehbach do you think this is sufficient to fix the issue you raised?

mkuehbach (approved):
Jip, this removes the possibly incorrect assumption that the exception was from an invalid appdef where in reality it was from a file not found

PR #401, opened 2024-08-12
Add documentation  for multi format reader

lukaspie:


docs/reference/plugins.md
```diff
@@ -1,12 +1,29 @@
 # Plugins
 There are a number of plugins available for pynxtools that are maintained within FAIRmat. These are extensions of pynxtools used for reading data of specific experimental techniques.
 
-- [**pynxtools-mpes**](https://github.com/FAIRmat-NFDI/pynxtools-mpes): A reader for multi-dimensional photoelectron spectroscopy data.
-- [**pynxtools-stm**](https://github.com/FAIRmat-NFDI/pynxtools-stm): A reader for scanning tunneling microscopy (SPM) and spectroscopy (STS) data.
-- [**pynxtools-xps**](https://github.com/FAIRmat-NFDI/pynxtools-xps): A reader for X-ray photoelectron spectroscopy (XPS) data. Documentation can be found [here](https://fairmat-nfdi.github.io/pynxtools-xps/).
+### Photoemission spectroscopy
+- [**pynxtools-mpes**](https://github.com/FAIRmat-NFDI/pynxtools-mpes): A reader for multi-dimensional photoelectron spectroscopy (MPES) data.
+- [**pynxtools-igor**](https://github.com/FAIRmat-NFDI/pynxtools-igor): A reader for MPES data stored in the [Igor Binary Wave Format](https://www.wavemetrics.com/).
```
domna, line 6: This repo does not exist. I have https://github.com/domna/pynxtools-igor but this just dumps some old stuff I worked on. The igor reader is basically included in the mpes and xps readers.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,385 @@
...
+```
+[project.entry-points."pynxtools.reader"]
+mydatareader = "pynxtools_plugin.reader:MyDataReader"
+```
+
+Here, we will focus mostly on the `reader.py` file and how to build a reader. For guidelines on how to build the other parts of your plugin, you can have a look here:
+- [Documentation writing guide](https://nomad-lab.eu/prod/v1/staging/docs/writing_guide.html)
+- [Plugin testing framework](using-pynxtools-test-framework.md)
+
+Note: There is also a [cookiecutter template](https://github.com/FAIRmat-NFDI/pynxtools-plugin-template) available for creating your own pynxtools plugin, but this is currently not well-maintained.
```
domna, line 39: I wouldn't even mention it here + the repo is private now.

domna: It's better to not say anything than to say that there is something which doesn't work well and is not recommended to use (it should rather just be added to the docs as soon as it's usable).

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,385 @@
...
+
+
+## Writing a Reader
+
+The pynxtools converter allows extending support to other data formats by allowing extensions called readers.
+The converter provides a dev platform to build a NeXus compatible reader by providing checking against a chosen NeXus Application Definition.
+
+<img src="media/convert_routine.svg"/>
```
domna, line 47: I find the picture a bit overcomplicated and you can essentially tell it's content by writing. 

```
dataconverter --reader <reader> --nxdl <nxdl> files ...
```

Also it still uses deprecated syntax such as `--input-file`.

domna: I recommend to only state the command here and explain it.

domna: You can also connect it to the reader name `mydatareader` from above. That way it's clear how to call a reader from the entrypoint.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,385 @@
...
+
+        self.extensions = {
+            ".yml": self.handle_eln_file,
+            ".yaml": self.handle_eln_file,
+            ".json": self.set_config_file,
+            # Here, you can add functions for handling any other file extension(s)
+        }
+# This has to be set to allow the convert script to use this reader. Set it to "MyDataReader".
+READER = MyDataReader
+```
```
domna, line 164: I advise against explaining this with the original reader code. It's rather complicated in comparison on how the reader is constructed (i.e., the complicated code is in the multiformatreader such that they can be implemented easier).

I would also add the functions you put into the dict here to explain it with a simple example.

domna: I would maybe give an example and then construct a reader for it around it. Expanding the reader step by step (first adding just an eln handling, then a `get_data` function etc etc.)

domna (commented):
I think it's generally a good approach but I would rather work with an example and build an example reader from ground up rather than showing the original code of the multi format reader. It is not required for users to understand the original code + we try to abstract this complication away with it.

docs/reference/plugins.md
lukaspie: removed

docs/how-tos/build-a-plugin.md
lukaspie: I removed the image and refactored this part a bit.

docs/how-tos/build-a-plugin.md
lukaspie: I decided to split these docs into a learning section (using the existing parts) and a how-to along the step by step idea described here.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
+# Build your own pynxtools plugin
+
+Your current data is not supported yet by the [built-in pynxtools readers](https://github.com/FAIRmat-NFDI/pynxtools/tree/master/src/pynxtools/dataconverter/readers) or any of the officially supported [pynxtools plugins](../reference/plugins.md)? 
+
+Don't worry, the following how-to will guide you how to write a reader your own data.
```
mkuehbach, line 5: "will guide you through the steps of writing a reader for your own data".

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+```
+pynxtools-plugin
+├── .github/workflows
+├── docs
+│   ├── explanation
+│   ├── how-tos
+│   ├── reference
+│   ├── tutorial
+├── src
+│   ├── pynxtools_plugin
```
mkuehbach, line 19: maybe one should make sure that "plugin" serves as a placeholder e.g. for xps
for "myownreader"

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+        for path in file_paths:
+            print(path)
+
+        template["/entry/instrument/scan"] = raw_scan_data
+
+        return template
+
+
+# This has to be set to allow the convert script to use this reader. Set it to "MyDataReader".
+READER = MyDataReader
```
mkuehbach, line 75: then above "pynxtools_mydata"

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+```
+### The reader template dictionary
+
+The read function takes a [`Template`](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/template.py) dictionary, which is used to map from the measurement (meta)data to the concepts defined in the application definition. It contains keys based on the provided NXDL file (you can get an empty template by using `dataconverter generate-template`).
```
mkuehbach, line 82: template

mkuehbach, line 82: NeXus application definition

mkuehbach: Instead of "It" make a more specific reference, the appdef contains keys but these are NeXus concept names and the template contains keys NeXus template dictionary keys

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+
+```python
+template["/entry/instrument/source"] = {"link": "/path/to/source/data"}
+```
+
+### Building off the BaseReader
+When building off the [`BaseReader`](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/readers/base/reader.py), the developer has the most flexibility. Any new reader must implement the `read` function, which must return a filled template object.
+
+
+### Building off the MultiFormatReader
```
mkuehbach, line 140: off from the

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+While building on the ```BaseReader``` allows for the most flexibility, in most cases it is desirable to implement a reader that can read in multiple file formats and then populate the template based on the read data. For this purpose, `pynxtools` has the [**`MultiFormatReader`**](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/readers/multi/reader.py), which can be readily extended for your own data.
+
+You can find an extensive how-to guide to build off the `MultiFormatReader` [here](./use-multi-format-reader.md).
+
+
+## Calling the reader from the command line
+
+The pynxtools converter allows extending support to other data formats by allowing extensions called readers.
+The converter provides a dev platform to build a NeXus compatible reader by providing checking against a chosen NeXus Application Definition.
```
mkuehbach, line 149: "application definition"

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,192 @@
...
+
+The dataconverter can be executed using:
+```console
+user@box:~$ dataconverter --reader mydatareader --nxdl NXmynxdl --output path_to_output.nxs
+```
+Here, the ``--reader`` flag must match the reader name defined in `[project.entry-points."pynxtools.reader"]` in the pyproject.toml file. The NXDL name passed to ``--nxdl``must be a valid NeXus NXDL/XML file in `pynxtools.definitions`.
+
+Aside from this default structure, there are many more flags that can be passed to the dataconverter call. Here is an ouput if its ```help``` call:
+```console
```
mkuehbach, line 158: single source principle, beyond this line because if the click options docstrings change they have to be changed also in this documentation here, instead one could use a python script whereby the current code is fished and then added here programmatically, avoiding a potential source of inconsistence between actual code and its documentation.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+It contains metadata about the user and the sample that was measured.
+
+We now need to write a function to read in this ELN data. Luckily, there exists already a solution within `pynxtools`, using the `parse_yaml` function:
+
+```python
+from pynxtools.dataconverter.readers.utils import parse_yml
+
+CONVERT_DICT = {
```
mkuehbach, line 172: @lukaspie indeed many similarities with concept mapping code in pynxtools-em and pynxtools-apm the tricky parts are cases like
detectorID[detector*], i.e. how to deal with multiple instances of things. I agree that this might be to much to cover for a getting started

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,396 @@
...
+    "/ENTRY[entry]/SAMPLE[sample]/name": "my_sample",
+    "/ENTRY[entry]/SAMPLE[sample]/physical_form": "powder",
+    "/ENTRY[entry]/SAMPLE[sample]/temperature": 300,
+    "/ENTRY[entry]/SAMPLE[sample]/temperature/@units": "K"
+}
+```
+Note that here we are using `parent_key="/ENTRY[entry]"` as well as a `CONVERT_DICT`, meaning that each key in `self.eln_data` will start with `"/ENTRY[entry]"` and some of the paths will be converted to match the template notation. This will be important later.
+
+## Parsing the config file
+Next up, we can make us of the config file, which is a JSON file that tells the reader how to map the concepts from the HDF5 and ELN files in order to populate the template designed to match `NXsimple`. Essentially, the config file should contain all keys that are present in the NXDL. In our case, the config file looks like this:
```
mkuehbach, line 205: I recommend to make the statement even stronger, the choices made in the config file define how semantics from the src and trg (NeXus) side are mapped. In other words, the config.json is ontology mapping in practice.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,396 @@
...
+def get_eln_data(self, key: str, path: str) -> Any:
+    """Returns data from the given eln path."""
+    if self.eln_data is None:
+        return None
+    
+    return self.eln_data.get(key)
+```
+Here, we are making use of the fact that we have used `CONVERT_DICT` in the `parse_yml` function above. Thus, the keys of the `self.eln_data` dictionary are exactly the same as those in the config file. We can therefore just get this data using the `key` coming from the config file.
+
+Finally, we also need to adddress the `@data` prefix, which gets used in the `data_callback` to populate the NXdata group in the template. Note that here we use the same `@data` prefix to fill the `x_values` as well as the `data` (from `y_values`) fields. We achieve this by using the path that follows `@data:` in the config file:
```
mkuehbach, line 274: address

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,396 @@
...
+For the ELN data, we must implement the `get_eln_data` function that gets called from the `eln_callback` when using the `@eln` prefix:
+```python
+def get_eln_data(self, key: str, path: str) -> Any:
+    """Returns data from the given eln path."""
+    if self.eln_data is None:
+        return None
+    
+    return self.eln_data.get(key)
+```
+Here, we are making use of the fact that we have used `CONVERT_DICT` in the `parse_yml` function above. Thus, the keys of the `self.eln_data` dictionary are exactly the same as those in the config file. We can therefore just get this data using the `key` coming from the config file.
```
mkuehbach, line 272: Therefore, we can ..

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,396 @@
...
+    ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.extensions = {
+            ".yml": self.handle_eln_file,
+            ".yaml": self.handle_eln_file,
+           ".json": self.set_config_file,
+            ".hdf5": self.handle_hdf5_file,
```
mkuehbach, line 321: Maybe a note of caution should be added:
along these lines:
"Case selection logics that rely on the filename suffix ! (not even the binary signature, mimetype) can be insufficient when multiple versions of hdf5 files exists, typical example H5OINA is HDF5 but there is v1, v2, v3 - v6... many concepts intersect but many are different

mkuehbach (commented):
@lukaspie some comments to an already good how-to shwoing people how to write their own parser @

docs/how-tos/build-a-plugin.md
lukaspie: It doesn't have to be like that. For example, `pynxtools-mpes` has two readers, `mpes` and `igor-mpes`. In principle, one plugin can have many readers.

docs/how-tos/build-a-plugin.md
lukaspie: see above

docs/how-tos/build-a-plugin.md
lukaspie: used uppercase `Template` here because it is a class.

lukaspie: adjusted the text

docs/how-tos/build-a-plugin.md
lukaspie: changed

docs/how-tos/build-a-plugin.md
lukaspie: 👍

docs/how-tos/build-a-plugin.md
lukaspie: I changed it to include a file that is dynamically created using a python script. Could be cleaner, but works for now.

docs/how-tos/use-multi-format-reader.md
lukaspie: added that to the docs describing the design of the multi-format-reader in `learnß 👍

docs/how-tos/use-multi-format-reader.md
lukaspie: changed

docs/how-tos/use-multi-format-reader.md
lukaspie: changed

docs/how-tos/use-multi-format-reader.md
lukaspie: Yeah, the idea is that the `self.handle_hdf5_file` method would implement such case selection logic. I will add a disclaimer in the `learn` docs.

docs/how-tos/use-multi-format-reader.md
lukaspie: This is exactly what we were discussing in #250 and I even suggested the same notation. We opted to not go for this for now because the PR was already large. I see two options:
- writing `detectorID[detector*]`, thus saying that `*` is in the key, it has to be overwritten
- simply writing `ENTRY/INSTRUMENT/detectorID`, i.e., if there is no default name, simply not write it at all. Then the reader must implement the replacements (but we could use default `entry/instrument/detector_id` for example).

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+    "/ENTRY[entry]/SAMPLE[sample]/name": "my_sample",
+    "/ENTRY[entry]/SAMPLE[sample]/physical_form": "powder",
+    "/ENTRY[entry]/SAMPLE[sample]/temperature": 300,
+    "/ENTRY[entry]/SAMPLE[sample]/temperature/@units": "K"
+}
+```
+Note that here we are using `parent_key="/ENTRY[entry]"` as well as a `CONVERT_DICT`, meaning that each key in `self.eln_data` will start with `"/ENTRY[entry]"` and some of the paths will be converted to match the template notation. This will be important later.
+
+## Parsing the config file
+Next up, we can make us of the config file, which is a JSON file that tells the reader how to map the concepts from the HDF5 and ELN files in order to populate the template designed to match `NXsimple`. Essentially, the config file should contain all keys that are present in the NXDL. In our case, the config file looks like this:
```
rettigl, line 205:
```suggestion
Next up, we can make use of the config file, which is a JSON file that tells the reader how to map the concepts from the HDF5 and ELN files in order to populate the template designed to match `NXsimple`. Essentially, the config file should contain all keys that are present in the NXDL. In our case, the config file looks like this:

```

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+
+I recommend to make the statement even stronger, the choices made in the config file define how semantics from the src and trg (NeXus) side are mapped.
```
rettigl, line 207: Implement and remove comment

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+    "temperature/@units":"@eln"
+  },
+  "/ENTRY/data": {
+    "@axes":["x_values"],
+    "@signal": "data",
+    "data": "@data:y_values",
+    "x_values": "@data:x_values"
+  }
+}
+```
```
rettigl, line 237: Add here at least a brief mentioning of the keyword concepts, and that they will be explained below.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+def get_attr(self, key: str, path: str) -> Any:
+    """
+    Get the metadata that was stored in the main file.
+    """
+    if self.hdf5_data is None:
+        return None
+    
+    return self.hdf5_data.get(path)
+```
+Here, we are making use of the `path`, which is the part of the config value that comes after the `@attrs:` prefix. For example, for the config value `"@attrs:metadata/instrument/version"`, the extracted path is `metadata/instrument/version`, which is also one of the keys of the `self.hdf5_data` dictionary.
```
rettigl, line 264: Also mention keys and what it contains

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+For the ELN data, we must implement the `get_eln_data` function that gets called from the `eln_callback` when using the `@eln` prefix:
+```python
+def get_eln_data(self, key: str, path: str) -> Any:
+    """Returns data from the given eln path."""
+    if self.eln_data is None:
+        return None
+    
+    return self.eln_data.get(key)
+```
+Here, we are making use of the fact that we have used `CONVERT_DICT` in the `parse_yml` function above. Thus, the keys of the `self.eln_data` dictionary are exactly the same as those in the config file. Therefore, we can just get this data using the `key` coming from the config file.
```
rettigl, line 275: Better explain the content of keys and path above (when introducing the special conept keys). Also check for non-empty path, and use rather than key, to allow for the example using eln paths.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+    "name":"@eln",
+    "physical_form":"@eln",
+    "temperature":"@eln",
+    "temperature/@units":"@eln"
+  },
+  "/ENTRY/data": {
+    "@axes":["x_values"],
+    "@signal": "data",
+    "data": "@data:y_values",
+    "x_values": "@data:x_values"
```
rettigl, line 234: aren't the units attributes missing here?

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+        if path.endswith(("x_values", "y_values")):
+            return self.hdf5_data.get(f"data/{path}")
+        else:
+            logger.warning(f"No axis name corresponding to the path {path}.")  
+   
+READER = MyDataReader
+```
+
+## Using the reader
+We can call our reader using the following
```
rettigl, line 391:
```suggestion
We can call our reader using the following command

```

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+        <group name="data" type="NXdata">
+            <doc>
+                 The default NXdata group containing a view on the measured data.
+            </doc>
+        </group>
+    </group>
+</definition>
+```
+
+The NXDL requires a user, some sample information, some instrument metadata, and the measured data to be written. Some groups, fields, and attributes are strictly required, but just recommended.
```
rettigl, line 68: Explain where to copy this file to make it be recognized

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+        Returns the processing order of the file.
+        """
+        ext = os.path.splitext(path)[1]
+        if self.processing_order is None or ext not in self.processing_order:
+            return (1, ext)
+        return (0, self.processing_order.index(ext))
+
+    sorted_paths = sorted(file_paths, key=get_processing_order)
+```
+If the reader has a `self.processing_order`, the input files get sorted in this order. If `self.overwrite_keys` is True, later files get precedent.
```
rettigl, line 69: Maybe give an example for a processing_order entry.

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+            )
+            continue
+        if not os.path.exists(file_path):
+            logger.warning(f"File {file_path} does not exist, ignoring entry.")
+            continue
+
+        template.update(self.extensions.get(extension, lambda _: {})(file_path))
+```
+This parts reads in the data from all data files. The `MultiFormatReader` has an `extensions` property, which is a dictionary that for each file extension calls a function that reads in data from files with that extension. If the reader shall handle e.g. an HDF5 file, a method for handling this type of file should be added, i.e., `self.extensions[".hdf5"] = self.handle_hdf5`. 
+Note that these method should also implement any case selection logic, i.e., it may not be sufficient to rely on the filename suffix, but when may also need to check for multiple versions, binary signature, mimetype, etc.
```
rettigl, line 87:
```suggestion
Note that these methods should also implement any logic depending on the provided data, i.e., it may not be sufficient to rely on the filename suffix, but the reader may also need to check for different file versions, binary signature, mimetype, etc.

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+```python
+    template.update(self.setup_template())
+```
+Next, the `setup_template` method can be implemented, which is used to populate the template with initial data that does not come from the files themselves. This may be used to set fixed information, e.g., about the reader. As an example, `NXentry/program_name` (which is defined as the name of program used to generate the NeXus file) scan be set to `pynxtools-plugin` by making `setup_template` return a dictionary of the form
+```json
+{
+  "/ENTRY[my_entry]/program_name": "pynxtools-plugin",
+  "/ENTRY[my_entry]/program_name/@version": "v0.1.0"
+}
+```
```
rettigl, line 107: Shoudn't this come before reading the files?

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Handling objects
+```python
+    if objects is not None:
+        template.update(self.handle_objects(objects))
+```
+Aside from data file, it is also possible to directly pass any Python objects to the `read` function (e.g., a numpy array with measurement data). In order to exploit this, the `handle_objects` method must implemented, which should return a dictionary that populates the template.
```
rettigl, line 114:
```suggestion
Aside from data files, it is also possible to directly pass any Python objects to the `read` function (e.g., a numpy array with measurement data). In order to exploit this, the `handle_objects` method must implemented, which should return a dictionary that populates the template.

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+    "@data": data_callback if data_callback is not None else self.identity,
+    "@eln": eln_callback if eln_callback is not None else self.identity,
+}
+```
+That means, if the config file has an entry ```{"/ENTRY/title": "@attrs:metadata/title"}```, the `get_attr` method of the reader gets called and should return an attribute from the given path, i.e., in this case from `metadata/title`.
+
+By default, the MultiFormatReader supports the following special prefixes:
+
+- `@attrs`: To get metadata from the read-in experiment file(s). You need to implement the `get_attr` method in the reader.
+- `@data`: To get measurement data from the read-in experiment file(s). You need to implement the `get_data` method in the reader.
```
rettigl, line 205: The destinction between data and metadata is somewhat arbitrary here. The reason to have both of these is for files/object types that require different methods to access different parts of the object, e.g. the xarrays. Maybe woth explaining this here.

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+- `@eln`: To get metadata from addtional ELN files. You need to implement the `get_eln_data` method in the reader.
+- `@link`: To implement a link between two entities in the NeXus file. By default, the link callback returns a dict of the form {"link": value.replace("/entry/", f"/{self.entry_name}/")}, i.e., a generic `/entry/` get replaced by the actual `entry_name`.
+
+In addition, the reader can also implement the `get_data_dims` method, which is used to return a list of the data dimensions (see below for more details).
+
+All of `get_attr`, `get_data`, and `get_eln_data`  (as well as any similar method that might be implemented) should have the same call signature:
+```python
+def get_data(self, key: str, path: str) -> Any:
+```
+Here, `key` is the config dict key (e.g., `"/ENTRY[my-entry]/data/data"`) and path is the path that comes _after_ the prefix in the config file. In this example, `path` would be `mydata`. With these two inputs, the reader should be able to return the correct data for this template key.
```
rettigl, line 215: Not sure which example you refer to here with mydata (?)

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Special rules
+- **Lists as config value**: It is possible to write a list of possible configurations of the sort
+  ```json
+  "/ENTRY/title":"['@attrs:my_title', '@eln', 'no title']"
+  ```
+  The value most be a string of a list, with each item being a string itself. This allows to provide different options depending if the data exists for a given callback. For each list item , it is checked if a value can be returned and if so, the value is written. In this example, the converter would check (in order) the `@attrs` (with path `"my_title"`) and `@eln` (with path `""`) tokens and write the respective value if it exists. If not, it defaults to "no title".
```
rettigl, line 222:
```suggestion
  The value must be a string which can be parsed as a list, with each item being a string itself. This allows to provide different options depending if the data exists for a given callback. For each list item , it is checked if a value can be returned and if so, the value is written. In this example, the converter would check (in order) the `@attrs` (with path `"my_title"`) and `@eln` (with path `""`) tokens and write the respective value if it exists. If not, it defaults to "no title".

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+  This concept can be particularly useful if the same config file is used for multiple measurement configurations, where for some setup, the same metadata may or not be available.
```
rettigl, line 223:
```suggestion
  This concept can be particularly useful if the same config file is used for multiple measurement configurations, where for some setup, the same metadata may or may not be available.

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+
+    Note that if this notation is used, it may be helpful to pass the `suppress_warning` keyword as `True` to the read function. Otherwise, there will be a warning for every non-existent value.
+
+- **Wildcard notation**: There exists a wildcard notation (using `*`)
+  ```json
+  "/ENTRY/data/AXISNAME[*]": "@data:*.data",
```
rettigl, line 229: I remember a discussion, that UPPERCASE notation is actually not valid for field names. What is the current status here? While the converter does not complain, it might just ignore it (?)

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+  ```
+  that allows filling multiple fields of the same type from a list of dimensions. This can be particularly helpful for writing `DATA` and `AXISNAME` fields that are all stored under similar paths in the read-in data.
+  For this, the `get_data_dims` method need to implemented. For a given path, it should return a list of all data axes available to replace the wildcard.
```
rettigl, line 232:
```suggestion
  For this, the `get_data_dims` method needs to be implemented. For a given path, it should return a list of all data axes available to replace the wildcard.

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+    
+    The same wildcard notation can also be used with a name to repeat entries with different names (e.g., *{my, name, etc} replaces to three keys with my, name, etc replaced, respectively). As an example, for multiple lenses and their voltage readouts, one could write:
```
rettigl, line 234:
```suggestion
    The same wildcard notation can also be used within a name to repeat entries with different names (e.g., field_*{my, name, etc} is converted into three keys with * replaced by my, name, etc, respectively). As an example, for multiple lenses and their voltage readouts, one could write:

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+  ```json
+  "LENS_EM[lens_*{A,B,C,Foc}]": {
+    "name": "*",
+    "voltage": "@attrs:metadata/file/Lens:*:V",
+    "voltage/@units": "V"
+  },
+  ```
+  which would write `NXlens_em` instances named `lens_A`, `lens_B`, and `lens_Foc`.
```
rettigl, line 242:
```suggestion
  which would write `NXlens_em` instances named `lens_A`, `lens_B`, `lens_C`, and `lens_Foc`.

```

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+
+    Note that the `get_data_dims` must correspond to all use cases, i.e., for `@data`, `@attrs`, and so on.
```
rettigl, line 244: This I don't understand. Will this not just replace "*" by the entries in the dict? Does this use case really require get_data_dims?

docs/learn/multi-format-reader.md
```diff
@@ -0,0 +1,263 @@
...
+
+- **Required fields in optional groups**: There will sometimes be the situation that there is an optional NeXus group in an application definition, that (if implemented) requires some sub-element. As an example, for the instrument's energy resolution, the only value expected to come from a data source is the `resolution`, whereas other fields are hardcoded.
+  ```json
+  "ENTRY/INSTRUMENT[instrument]/energy_resolution": {
+    "resolution": "@attrs:metadata/instrument/electronanalyser/energy_resolution",
+    "resolution/@units": "meV",
+    "physical_quantity": "energy"
+  }
+  ```
+  Now, if there is no data for `@attrs:metadata/instrument/electronanalyser/energy_resolution` available in a dataset, this will be skipped by the reader, and not available, yet the other entries are present. During validation, this means that the required field resolution of the optional group `energy_resolution` is not present, and thus an error would be raised:
```
rettigl, line 254:
```suggestion
  Now, if there is no data for `@attrs:metadata/instrument/electronanalyser/energy_resolution` available in a dataset, this will be skipped by the reader, and not available, yet the other entries are present. During validation, this means that the required field `resolution` of the optional group `energy_resolution` is not present, and thus a warning or error would be raised:

```

docs/how-tos/use-multi-format-reader.md
lukaspie: I think we are talking about two different things here:
1) (not) using the `CONVERT_DICT`. I agree that your approach of non-replacing is probably better. I went with using the replacement in the example here so that people can see that you can also use the `key` instead of the `path` passed to `get_eln_data` to fill the template.
2) what @mkuehbach was talking about is the option of having multiple instances of e.g. a detector, which only get detected when handling the file. In that case, it might be a good idea to have a wildcard syntax in the config _key_ that tells the reader that these names must be replaced (this is what I outlined above)

docs/how-tos/build-a-plugin.md
lukaspie: We tried this already using cookiecutter: https://github.com/FAIRmat-NFDI/pynxtools-plugin-template (repo is private atm). However, we didn't maintain it over time (because there was no resources for it), so it quickly became outdated. It is on my ToDo list and we may even use [cruft](https://cruft.github.io/cruft/) to automatically update all the plugins once there is a change in the template. But since it is not working, I have not included it here.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+        objects: Tuple[Any] = None
+    ) -> dict:
+        """Reads data from given file and returns a filled template dictionary"""
+        # Fill the template
+        for path in file_paths:
+            print(path)
+
+        template["/entry/instrument/scan"] = raw_scan_data
+
+        return template
```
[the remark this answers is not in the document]
lukaspie: I added some comment lines to indicate which parts a user should change, also added the `supported_nxdls`.

What do you mean by `provided as a template`? Where exactly?

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+
+# This has to be set to allow the convert script to use this reader. Set it to "MyDataReader".
+READER = MyDataReader
+
+
+
+```
+### The reader template dictionary
+
+The read function takes a [`Template`](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/template.py) dictionary, which is used to map from the measurement (meta)data to the concepts defined in the NeXus application definition. The template contains keys that match the concepts in the provided NXDL file (you can get an empty template by using `dataconverter generate-template`).
```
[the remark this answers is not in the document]
lukaspie: I just removed it since the two instances were very close together.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+
+```json
+{
+  "/entry/instrument/source/type": "None"
+}
+```
+
+For a given NXDL schema, you can generate an empty template with the command
+```console
+user@box:~$ dataconverter generate-template` --nxdl NXmynxdl
```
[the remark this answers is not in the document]
lukaspie: implemented

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+#### Naming of groups
+In case the NXDL does not define a `name` for the group the requested data belongs to, the template dictionary will list it as `/NAME_IN_NXDL[name_in_output_nexus]`
+You can choose any name you prefer instead of the suggested name (see [here](../learn/nexus-rules.md) for the naming conventions). This allows the reader function to repeat groups defined in the NXDL to be outputted to the NeXus file.
```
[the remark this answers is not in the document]
lukaspie: changed

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+For attributes defined in the NXDL, the reader template dictionary will have the assosciated key with a "@" prefix to the attributes name at the end of the path:
+
+```json
+{
+  "/entry/instrument/source/@attribute": "None"
+}
+```
+
+#### Units
+If there is a field defined in the NXDL, the converter expects a filled in /data/@units entry in the template dictionary corresponding to the right /data field unless it is specified as NX_UNITLESS in the NXDL. Otherwise, you will get an exception.
```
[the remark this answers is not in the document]
lukaspie: changed. Note that I did not use the suggestion because there was something messing up when I tried, but rather I changed it locally. Same for all other suggestions.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+
+The dataconverter can be executed using:
+```console
+user@box:~$ dataconverter --reader mydatareader --nxdl NXmynxdl --output path_to_output.nxs
+```
+Here, the ``--reader`` flag must match the reader name defined in `[project.entry-points."pynxtools.reader"]` in the pyproject.toml file. The NXDL name passed to ``--nxdl``must be a valid NeXus NXDL/XML file in `pynxtools.definitions`.
+
+Aside from this default structure, there are many more flags that can be passed to the
+dataconverter call. Here is an output of its ```help``` call:
+{!how-tos/dataconverter_help_output.md!}
```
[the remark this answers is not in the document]
lukaspie: This is a bit weird because locally (using `mdocs serve`) it worked. Investigating...

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+### Building off of the MultiFormatReader
+While building on the ```BaseReader``` allows for the most flexibility, in most cases it is desirable to implement a reader that can read in multiple file formats and then populate the template based on the read data. For this purpose, `pynxtools` has the [**`MultiFormatReader`**](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/readers/multi/reader.py), which can be readily extended for your own data.
+
+You can find an extensive how-to guide to build off the `MultiFormatReader` [here](./use-multi-format-reader.md).
+
+
+## Calling the reader from the command line
+
+The pynxtools converter allows extending support to other data formats by allowing extensions called readers.
+The converter provides a dev platform to build a NeXus compatible reader by providing checking against a chosen NeXus application definition.
```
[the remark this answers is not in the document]
lukaspie: Yeah, this was copied from somewhere else. I moved it to the top.

examples/mock-data-reader/config_file.json
```diff
@@ -0,0 +1,26 @@
...
+    "DETECTOR[detector]":{
+      "count_time":"@attrs:metadata/instrument/detector/count_time",
+      "count_time/@units":"@attrs:metadata/instrument/detector/count_time_units"
+    }
+  },
+  "/ENTRY/SAMPLE[sample]": {
+    "name":"@eln",
+    "physical_form":"@eln",
+    "temperature":"@eln",
+    "temperature/@units":"@eln"
```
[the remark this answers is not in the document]
lukaspie: We can really only either do `@eln` using the `CONVERT_DICT` and the config key OR the `@eln:path/in/dict` notation. As outlined above, I opted for the first option because the second one is covered by the `@attrs` examples.

I did not use the `!` here in order not to overcomplicate things. The user can just check the MultiReaderFormat docs in `learn` for these special use cases.

examples/mock-data-reader/create_mock_data.py
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+"""A script to create mock data for the example reader."""
+
+import numpy as np
+import h5py
+
+
+# Function to generate a Lorentzian curve
+def lorentzian(x, x0, gamma):
+    return (gamma**2) / ((x - x0) ** 2 + gamma**2)
```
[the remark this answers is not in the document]
lukaspie: Yes, intially I wanted to have two entries, but it became too complicated. Removed.

scripts/generate_dataconverter_help.py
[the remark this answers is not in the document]
lukaspie: It worked locally, but apparently not through the CI/CD. I need to check what went wrong.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
+# How to use the built-in MultiFormatReader
+While building on the ```BaseReader``` allows for the most flexibility, in most cases it is desirable to implement a reader that can read in multiple file formats and then populate the template based on the read data. For this purpose, `pynxtools` has the [**`MultiFormatReader`**](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/dataconverter/readers/multi/reader.py), which can be readily extended for your own data. In this how-to guide, we will focus on an implementation using a concrete example. If you are also interested in the general structure of the `MultiFormatReader`, you can find more information [here](../learn/multi-format-reader.md).
+
+## Getting started
+
+**Note: You can find all of the data and the developed python scripts [here](https://download-directory.github.io/?url=https://github.com/FAIRmat-NFDI/pynxtools/tree/master/examples/mock-data-reader).**
```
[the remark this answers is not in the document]
lukaspie: This is expected, since the path does not exist in the master branch yet (will only be there after merging). I will check again after merging that the download actually works.

docs/how-tos/use-multi-format-reader.md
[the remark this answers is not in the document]
lukaspie: changed

docs/how-tos/build-a-plugin.md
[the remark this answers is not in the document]
rettigl: > What do you mean by "provided as a template"? Where exactly?

What I mean is to have this code in the reader.py file in an example reader.

examples/mock-data-reader/config_file.json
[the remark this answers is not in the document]
rettigl: You could use both, if you check for the value of path. But maybe that is not good practice anyhow. I suggest then to keep the examples as they are, but mention there is also the other way for eln entries.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+            ".hdf5": self.handle_eln_file,
+            ".h5": self.handle_eln_file,
+        }
+
+READER = MyDataReader
+```
+Note that here we are adding handlers for three types of data file extensions:
+1) `".hdf5"`, `".h5"`: This will be used to parse in the (meta)data from the instrument's HDF5 file.
+2) `".yml"`, `".yaml"`: This will be used to parse in the (meta)data from the ELN file.
+3) `".json"`: This will be used to read in the **config file**, which is used to map from the (meta)data concepts from the instrument and ELN data to the concepts in the NXDL file.
```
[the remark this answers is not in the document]
lukaspie: changed

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+                result[new_path] = item[()]
+        return result
+
+    # Open the HDF5 file and read its contents
+    with h5py.File(filepath, "r") as hdf:
+        self.hdf5_data = recursively_read_group(hdf)
+
+    return {}
+```
+Note that here we are returning an empty dictionary because we don't want to fill the template just yet, but only read in the HDF5 data for now. We will use the config file later to fill the template with the read-in data.
```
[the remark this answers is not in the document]
lukaspie: added

examples/mock-data-reader/config_file.json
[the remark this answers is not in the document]
lukaspie: Actually, we can just still use `@eln:"/ENTRY[entry]/USER[user]/address"` and change the `get_eln_data` to use either the key or the path. Will do.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+
+READER = MyDataReader
+```
+Note that here we are adding handlers for three types of data file extensions:
+1) `".hdf5"`, `".h5"`: This will be used to parse in the (meta)data from the instrument's HDF5 file.
+2) `".yml"`, `".yaml"`: This will be used to parse in the (meta)data from the ELN file.
+3) `".json"`: This will be used to read in the **config file**, which is used to map from the (meta)data concepts from the instrument and ELN data to the concepts in the NXDL file.
+
+## Reading in the instrument's data and metadata
+First, we will have a look at the HDF5 file.
```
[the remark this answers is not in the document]
lukaspie: added a link to the script.

docs/how-tos/use-multi-format-reader.md
```diff
@@ -0,0 +1,399 @@
...
+    "user": "USER[user]",
+    "instrument": "INSTRUMENT[instrument]",
+    "detector": "DETECTOR[detector]",
+    "sample": "SAMPLE[sample]",
+}
+
+def handle_eln_file(self, file_path: str) -> Dict[str, Any]:
+    self.eln_data = parse_yml(
+        file_path,
+        convert_dict=CONVERT_DICT,
```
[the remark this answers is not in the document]
lukaspie: See above, I implemented both options here.

docs/how-tos/use-multi-format-reader.md
lukaspie: changed

docs/how-tos/use-multi-format-reader.md
lukaspie: Right, I had forgotten about this. Changed.

docs/how-tos/use-multi-format-reader.md
lukaspie: good idea, added

docs/how-tos/use-multi-format-reader.md
lukaspie: I added this in the ELN part where we actually use the `key`.

docs/how-tos/use-multi-format-reader.md
lukaspie: You are right, added using `@attrs`

docs/how-tos/use-multi-format-reader.md
lukaspie: added

docs/learn/multi-format-reader.md
lukaspie: changed

docs/learn/multi-format-reader.md
lukaspie: Currently, it comes afterwards (and after potential hard-coded update of the templates in the reading functions). We can discuss to change the order if you like.

docs/learn/multi-format-reader.md
lukaspie: changed

docs/learn/multi-format-reader.md
lukaspie: good point, added

docs/learn/multi-format-reader.md
lukaspie: i was refering to the example config file given above. Clarified here.

docs/learn/multi-format-reader.md
lukaspie: changed

docs/learn/multi-format-reader.md
lukaspie: 👍

docs/learn/multi-format-reader.md
lukaspie: This can only for the `NXdata` where we have `AXISNAME` and `DATA` and there is some more logic (like `signal`, `axes` and so on) to distinguish the two. The verification handles this special case differently. Other than that, it is indeed not allowed.

docs/learn/multi-format-reader.md
lukaspie: changed

docs/learn/multi-format-reader.md
lukaspie: that's clearer, thanks

docs/learn/multi-format-reader.md
lukaspie: just removed the C from the config to make it shorter

docs/learn/multi-format-reader.md
lukaspie: correct

docs/learn/multi-format-reader.md
lukaspie: removed

docs/how-tos/use-multi-format-reader.md
lukaspie: changed

docs/how-tos/build-a-plugin.md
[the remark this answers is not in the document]
lukaspie: Hm, I see. However, that would basically be an empty example that cannot be used. Not sure if that is the best way to go..

examples/mock-data-reader/config_file.json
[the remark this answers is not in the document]
lukaspie: Sorry, didn't read your previous comment. I agree it's probably best to not mix these. I'll just say that you can also not do any replacement and use the path 👍

docs/how-tos/use-multi-format-reader.md
lukaspie: EDIT: added a longer description what `key` and `path` are above.

docs/how-tos/use-multi-format-reader.md
lukaspie: First part done. Second part not implemented as discussed.

docs/learn/multi-format-reader.md
lukaspie: added

docs/learn/multi-format-reader.md
rettigl: I don't use either option, so in general I don't mind. Let's keep as is at least for now.

docs/how-tos/build-a-plugin.md
```diff
@@ -0,0 +1,159 @@
...
+```json
+{
+  "/entry/instrument/source/type": "None"
+}
+```
+
+For a given NXDL schema, you can generate an empty template with the command
+```console
+user@box:~$ dataconverter generate-template` --nxdl NXmynxdl
+```
```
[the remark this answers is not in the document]
lukaspie: There seems to be some sort of technical solution, but it comes with looks of ugly changes in the css. We are using this in several places (not just in these docs), so I made a new issue to address this separately: #406 .

.github/workflows/build_docs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master  # Triggers deployment on push to the main branch
+      - 398-add-documentation-pages-for-multi-format-reader
```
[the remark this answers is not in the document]
lukaspie: done

scripts/generate_dataconverter_help.py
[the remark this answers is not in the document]
lukaspie: Found the problem, pynxtools wasn't properly installed with its submodule in the CI/CD

docs/how-tos/build-a-plugin.md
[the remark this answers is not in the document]
lukaspie: works now

scripts/generate_dataconverter_help.py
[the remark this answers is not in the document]
lukaspie: Just FYI: in case one wants to run such a script when `mkdocs serve` is called, you need to use the mkdocs hooks to run it: https://github.com/FAIRmat-NFDI/pynxtools/blob/2084b6076dca360dbcd70482745c931019b91e16/mkdocs.yaml#L35

PR #404, opened 2024-08-21
Suggestion for fixing the unintentional situation that since v0.6 the…

mkuehbach:
… dataconverter required a config file if it was called the classical way and in case no parameter yaml file was provided. For programmatically use of the converter the creation of an additional params file with only a few entries is unnecessary overhead especially when all cli arguments are available in the callers scope and thus the CLI call can be issues directly using the classical way dataconverter convert [INPUTFILES] options


mkuehbach:
I also removed the outdated documentation of the command line options as it is a second place where changes have to be applied constantly and the instructions for empowering people to find these are already given in the readme plus follow the typical style how to get help for command line calls

src/pynxtools/dataconverter/convert.py
```diff
@@ -284,28 +284,32 @@ def main_cli():
...
-        "The path to the input data file to read. (Repeat for more than one file.)"
+        "Deprecated: Please use the positional file arguments instead. The path to the "
+        "input data file to read. Repeat for more than one file. default=[] This option "
+        "is required if no '--params-file' is supplied."
     ),
 )
 @click.option(
     "--reader",
-    default="json_map",
+    default=None,
```
mkuehbach, line 294: I removed it because no reader should be preferred here

sherjeelshabih: It's not really preferred. That reader is added as the default behavior for using mapping. We added this to the docs before as well. In the future this can be replaced with the multiformat reader if it fits better. 

If we remove this we cannot directly use the mapping flag. These were quality of life improvements we did before.

```suggestion
    default="json_map",
```

lukaspie: Agreed that it should be removed. The multiformat reader would not fit better because it does not actually implement the json mapping. I guess `json_map` could inherit from the multiformat reader eventually.

src/pynxtools/dataconverter/convert.py
```diff
@@ -333,12 +337,14 @@ def main_cli():
 )
 @click.option(
     "--mapping",
+    default=None,
```
sherjeelshabih, line 340: Related to json_map as QOL.

```suggestion
```

src/pynxtools/dataconverter/convert.py
```diff
@@ -284,28 +284,32 @@ def main_cli():
...
-        "This option is required if no '--params-file' is supplied."
+        "The name of the NeXus application definition NXDL file to use without the "
+        "extension nxdl.xml. This option is required if no '--params-file' is supplied."
     ),
 )
 @click.option(
     "--output",
     default="output.nxs",
-    help="The path to the output NeXus file to be generated.",
+    help="The path to the output NeXus/HDF5 file to be generated.",
```
sherjeelshabih, line 312: Unnecessary tbh.

```suggestion
    help="The path to the output NeXus file to be generated.",
```

sherjeelshabih (dismissed):

src/pynxtools/dataconverter/convert.py
```diff
@@ -392,16 +399,27 @@ def convert_cli(
...
+        else:
+            convert(
+                tuple(file_list) + input_file,
+                reader,
+                nxdl,
+                output,
+                skip_verify,
+                ignore_undocumented=ignore_undocumented,
+                fail=fail,
+            )
```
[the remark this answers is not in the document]
mkuehbach: My observation was that if branching off it works if using the original one it did not but failed with that a config_file was required despite the fact that the default for the click option config was set to zero and nowhere in my call line I pass a config file.

src/pynxtools/dataconverter/convert.py
mkuehbach: No, NeXus is not required to write to HDF5 only

mkuehbach: NeXus file is an umbrella term but that implementation here writes an HDF5 file

sherjeelshabih: It's just semantic tbh. Isn't a HDF5 Nexus file a subset of a Nexus file? I don't mind this going either way. It's just we can explain how we only write HDF5 somewhere else if we need to. This seems to take space for no real added benefit. But again. I don't have a strong opinion on this. I just like to keep it as clean as possible. I'll let @lukaspie give his opinion/last decision on the matter?

src/pynxtools/dataconverter/convert.py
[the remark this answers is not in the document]
sherjeelshabih: With the default=None for config, this should work without branching. I see no real diff. Can you tell me what error you're getting? Did you try again after adding the default=None above?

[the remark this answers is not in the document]
mkuehbach: Sure I did, I tested before https://github.com/FAIRmat-NFDI/pynxtools/commit/e7100a517faf622a903d678e052aa0b379b3cf41, I test with that e7100, and I tested with the latest one, for the latest there were no issues, so naively None is carried over.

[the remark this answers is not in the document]
sherjeelshabih: This has been fixed with my last commit.

mkuehbach:
Tested 30ad68c of this branch solves the problems @sherjeelshabih

src/pynxtools/dataconverter/convert.py
mkuehbach: Reverted

src/pynxtools/dataconverter/convert.py
mkuehbach: Reverted

src/pynxtools/dataconverter/convert.py
lukaspie: It really is just semantics, let's go with what is proposed here.

src/pynxtools/dataconverter/convert.py
```diff
@@ -398,9 +406,9 @@ def convert_cli(
             nxdl,
             output,
             skip_verify,
-            config_file=config,
             ignore_undocumented=ignore_undocumented,
             fail=fail,
+            **kwargs,
```
lukaspie, line 411: This doesn't work because `config_file` is not in kwargs (it is an input to this function). We can simply add it to kwargs if it exists:
```
if config_file:
    kwargs["config_file"] = config_file
```
Actually, this whole discussion is part of a bigger problem, namely that we call 
```
data = data_reader().read(  # type: ignore[operator]
        template=Template(template), file_paths=input_file, **kwargs
    )
```
in the `convert` function. But, most of our reader don't actually accept kwargs, so this will always be a bit unstable.

lukaspie: I added 
```
if config_file:
    kwargs["config_file"] = config_file
```
to the `convert_cli` function.

lukaspie (approved):
I tested with
- `pynxtools-xps`, both with and w/o a config file
- `pyntools-mpes`, always uses config file
- `pynxtools-stm`
- `pynxtools-em`, with the example @mkuehbach send me
- `json_map` -> with examples in `example/json-map`

@mkuehbach feel free to merge if your examples still run through (wasn't sure if you tested something else as well). Thanks for the fix!

src/pynxtools/dataconverter/convert.py
sherjeelshabih: Good find. That fix works. Thank you.

If I remember correctly, Python expands all values of kwargs automatically. So it will pass what it receives. So the signature of the read function in the reader should match the parameters we sent to the cli during invocation. If we run into more errors, we can have this more thoroughly controlled before it's passed to the readers.

lukaspie: Yeah, for now it works. I guess at some point we could declutter these things, but for it's fine.

rettigl: Something like `func_name.__code__.co_varnames` should give you a list of possible keyword arguments, which you can use for raising a suitable error message.

PR #407, opened 2024-08-22
Fix indentation in multiformat reader docs

lukaspie:


mkuehbach (approved):
Lifting each header one level up, LGTM

PR #408, opened 2024-08-22
Check axes of dataset is list or str.

RubelMozumder:
Issue:
It raises error when `axes` attribute of NXdata is empty. though `axes` is optional.

### DIFF
--- a/src/pynxtools/nexus/nexus.py
+++ b/src/pynxtools/nexus/nexus.py
@@ -595,7 +595,7 @@ def get_single_or_multiple_axes(nxdata, ax_datasets, a_item, ax_list, logger):
                 ax_list.append(nxdata[ax_datasets[a_item]])
         else:
             logger.warning(
-                f"The 'axes' attribute is neither a string or a list or an np.ndarray of strings, check {nxdata.name}”
+                f"The 'axes' attribute is neither a string or a list or an np.ndarray of strings, check {nxdata.name}"
             )
 
     except KeyError:
--- a/src/pynxtools/nexus/nexus.py
+++ b/src/pynxtools/nexus/nexus.py
@@ -595,7 +595,7 @@ def get_single_or_multiple_axes(nxdata, ax_datasets, a_item, ax_list, logger):
                 ax_list.append(nxdata[ax_datasets[a_item]])
         else:
             logger.warning(
-                f"The 'axes' attribute is not a string or list or np.ndarray of string, check {nxdata.name}"
+                f"The 'axes' attribute is neither a string or a list or an np.ndarray of strings, check {nxdata.name}”
             )
 
     except KeyError:
--- a/src/pynxtools/nexus/nexus.py
+++ b/src/pynxtools/nexus/nexus.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 import click
 import h5py
 import lxml.etree as ET
+import numpy as np
 
 from pynxtools.definitions.dev_tools.utils.nxdl_utils import (
     add_base_classes,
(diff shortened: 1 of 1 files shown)


src/pynxtools/nexus/nexus.py
```diff
@@ -581,7 +582,7 @@ def get_single_or_multiple_axes(nxdata, ax_datasets, a_item, ax_list):
                     ax_list.append(nxdata[ax_datasets])
             elif a_item == 0:  # positional determination of the dimension number
                 ax_list.append(nxdata[ax_datasets])
-        else:  # multiple axes are listed
+        elif isinstance(ax_datasets, (list, np.ndarray)):  # multiple axes are listed
```
RonHildebrandt, line 585: So, now the case, that there is no axis attribute given. At which line of the code, now there is a change in behvior that causes the error?
Which of these variables of "get_single_or_multiple_axes" describes the attribute, which is not present as it is optional?
Maybe I have to reproduce this to understand it - I may also be limited in the knowledge regarding the pynxtools code still....

RubelMozumder: Previous Code:
If no `ax_datasets` is `none` for loop raise an `NoneType` error.
Present Code:
Now the code will check, is it a `list` or `numpy list` otherwise the code will skip that clause and go to the else clause. 

To understand how the empty list `axis` gets handled, you may need to back of the code where the function has been called and how the empty list of `axis` has been handled.

One Clud: The except block says, the parent function can handle a empty list or partially filled.

src/pynxtools/nexus/nexus.py
```diff
@@ -592,7 +593,15 @@ def get_single_or_multiple_axes(nxdata, ax_datasets, a_item, ax_list):
                 ax_datasets
             ):  # positional determination of the dimension number
                 ax_list.append(nxdata[ax_datasets[a_item]])
+        else:
+            logger.warning(
+                f"The 'axes' attribute is not a string or list or np.ndarray of string, check {nxdata.name}"
```
lukaspie, line 598:
```suggestion
                f"The 'axes' attribute is neither a string or a list or an np.ndarray of strings, check {nxdata.name}”
```

src/pynxtools/nexus/nexus.py
```diff
@@ -602,7 +611,9 @@ def axis_helper(dim, nxdata, signal, axes, logger):
     for a_item in range(dim):
         ax_list = []
         ax_datasets = nxdata.attrs.get("axes")  # primary axes listed in attribute axes
-        ax_list = get_single_or_multiple_axes(nxdata, ax_datasets, a_item, ax_list)
+        ax_list = get_single_or_multiple_axes(
+            nxdata, ax_datasets, a_item, ax_list, logger
```
lukaspie, line 615: Probably the logger should be setup somewhere globally, so it doesn't have to passed down to each function. But since this is the case everywhere in this tools, let's go with what is suggested here.

lukaspie (approved):
lgtm

src/pynxtools/nexus/nexus.py
RubelMozumder: I agree and I also noticed that `logger` has not been used everywhere or at least should be a global logger for each module, so that we can call when it is needed. 
We should work in a new PR and also check the compatibility of the `pynxtools` logger to the `nomad` logger.  Though, I believe we can see the pynxtools log massage in `nomad`. 

But this PR is not intended for logger fix.

lukaspie: Can you make an issue for this please?

PR #409, opened 2024-08-23
Skip overwritting entry type 

RubelMozumder:
Skip overwriting entry type is NeXus parser is called under other parsers for example `XRD`.

### DIFF
--- a/src/pynxtools/nomad/parser.py
+++ b/src/pynxtools/nomad/parser.py
@@ -487,9 +487,9 @@ class NexusParser(MatchingParser):
             if getattr(archive.nexus, var, None) is not None:
                 app_def = var
                 break
-
-        archive.metadata.entry_type = app_def
-        archive.metadata.domain = "nexus"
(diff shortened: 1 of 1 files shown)


RubelMozumder:
@sherjeelshabih, here how i implemented it here: https://github.com/FAIRmat-NFDI/nomad-measurements/pull/108/files#diff-1b617f8d59d5e4f9fbe49cc73ed0caf4b661c56796e7901e776753da7d10e68bR241-R256
First of all, they do not like to overwrite the `entry_name` here in `nomad_measurement`. Even, I overwrite the `entry_name` in `nomad_measurement` still `metadata.domain=NeXus` Which area A does not want.

And, this discussion is going on for a few days.

sherjeelshabih:
I understand what you're dealing with. I'm happy to go along with this change. Could you just check if this works with a .nxs file drag and dropped into a Nomad instance installed with this Pynxtools branch? Then we're fine to go with this.

lukaspie (approved):

PR #415, opened 2024-08-29
Add NX docstring as attribute

### LINKED ISSUE #403: [Improvement] Add NX doc string as attribute
Add the relevant/resolved Nexus doc string as attribute to every written field

lukaspie:
@rettigl this is a way to add the NX docstrings to the HDF5 files as attributes. The idea is that you can pass the `write-docs` flag to the dataconverter and then docstrings are added. By default, this is turned off so as to not change any existing workflows. We can discuss if we want this to happen by default.

Here's an example using the xps reader: 
[output.nxs.zip](https://github.com/user-attachments/files/16815831/output.nxs.zip)

The implementation is relatively trivial. There are however two open questions for me
- **How to handle inherited docs**
  Usually, our appdefs have very sparse documentation since all the docstrings are in the base classes or in appdefs that are further up the chain. My understanding is that the docs are also extended (not overwritten) when you inherit from a class. So technically, one should concatenate the docs of all inherited nodes. However, this gets quite unwieldy and messy as the docs in the files get rather large and sometimes they might even contradict each other. My suggestion (implemented here) was that as we travel up the inherited nodes, as soon as there is a docstring we only use that one (and don't add any docs coming from even earlier nodes). But this is probably not strictly correct. Maybe @sanbrock can comment.

- **Docs for NeXus attributes**
  NeXus attributes are written as HDF5 attributes already. Since HDF5 attributes cannot have attributes themselves, the question is where to place the docs for these attributes? My solution here: write another attribute `<attribute>__docs` (e.g. entry/definition/version__docs) to the HDF5 file.

### DIFF
--- a/.cspell/custom-dictionary.txt
+++ b/.cspell/custom-dictionary.txt
@@ -243,6 +243,7 @@ posteriori
 prescan
 printoptions
 prng
+pseudoxml
 punx
 pynx
 pynxtools
--- a/.cspell/custom-dictionary.txt
+++ b/.cspell/custom-dictionary.txt
@@ -243,7 +243,6 @@ posteriori
 prescan
 printoptions
 prng
-pseudoxml
 punx
 pynx
 pynxtools
--- a/docs/learn/pynxtools/dataconverter-and-readers.md
+++ b/docs/learn/pynxtools/dataconverter-and-readers.md
@@ -84,9 +84,9 @@ Pass `--write-docs` to embed the NeXus concept documentation from the NXDL schem
 pynx convert run --reader <reader-name> --nxdl <NXDL> <files> --write-docs
 ```
 
-Each HDF5 group and field receives a `@docs` attribute, and each HDF5 attribute receives a `@<attr-name>_docs` attribute. When a concept is defined across multiple levels of the NeXus inheritance hierarchy, all contributing docstrings are included, labeled by their concept path (e.g. `NXentry/definition: ...`).
+Each HDF5 group and field receives a `@docs` attribute containing the NXDL docstring, and a `@docs_url` attribute with a direct URL to the concept in the [NeXus manual](https://manual.nexusformat.org/classes) (e.g. `https://manual.nexusformat.org/classes/base_classes/NXdata.html#nxdata-data-field`). Each HDF5 attribute additionally receives `@<attr-name>_docs` and `@<attr-name>_docs_url` companions. When a concept is defined across multiple levels of the NeXus inheritance hierarchy, all contributing docstrings are included in `@docs`, labeled by their concept path (e.g. `NXentry/definition: ...`); `@docs_url` points to the most specific level.
 
-The `--docs-format` option controls how the docstrings are stored:
+The `--docs-format` option controls how the docstrings in `@docs` are stored:
 
 | Format | Description | Best for |
 |--------|-------------|----------|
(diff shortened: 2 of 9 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "src/pynxtools/definitions"]
 	path = src/pynxtools/definitions
 	url = https://github.com/FAIRmat-NFDI/nexus_definitions.git
+	branch = nxmpes-unit-doc-test
```
lukaspie, line 4: reminder to remove

sherjeelshabih:
> - **Docs for NeXus attributes**
>   NeXus attributes are written as HDF5 attributes already. Since HDF5 attributes cannot have attributes themselves, the question is where to place the docs for these attributes? My solution here: write another attribute `<attribute>__docs` (e.g. entry/definition/version__docs) to the HDF5 file.
>

Any practical solution works in my opinion. The only issue will be how to report/add this in the NXDL structure. It's a bit meta over the already meta attributes we have in the NXDL. Will something like adding this renameable field, FIELDNAME__docs, in NXobject work out nicely in the NXDL framework?

lukaspie:
> > * **Docs for NeXus attributes**
> >   NeXus attributes are written as HDF5 attributes already. Since HDF5 attributes cannot have attributes themselves, the question is where to place the docs for these attributes? My solution here: write another attribute `<attribute>__docs` (e.g. entry/definition/version__docs) to the HDF5 file.
> 
> Any practical solution works in my opinion. The only issue will be how to report/add this in the NXDL structure. It's a bit meta over the already meta attributes we have in the NXDL. Will something like adding this renameable field, FIELDNAME__docs, in NXobject work out nicely in the NXDL framework?

I didn't go this far. For groups and fields, I basically added an attribute `@docs`, whereas for the attributes, I used an additional attribute `<attribute>__docs`. I guess the problem would be that all of these are undocumented..

sherjeelshabih:
Ah alright. So it's just for the attributes that you add a suffix. 

You're right. They will remain undocumented. Let's say to see how it goes in use for us we can leave it undocumented. 

It will make it practically easier to understand Nexus files like this. It makes the Nexus files more self sufficient too. And it seems this is the best we can do without overcomplicating it.

lukaspie:
Notes from TF meeting:
- Use single underscore for attributes, like for example `data` and `data_units`
- Can we render the docs a bit nicer? HTML syntax?
- Don't write the documentation to every single instance of the same concept, use internal / softlink
- External links to documentation? `docs` and `docs_url`
- @sanbrock: HDF5 contains _data_, not information. What happens when we want to reuse the data somewhere else? We could prevent this by writing a parallel structure with the docs and linking to that place.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -790,6 +790,64 @@ def get_concept_basepath(path: str) -> str:
     return "/" + "/".join(concept_path)
 
 
+def get_concept_path_from_elem(elem: ET.Element) -> str:
+    """
+    Process individual XML element to generate the NeXus concept path.
+
+    Output is e.g. "NXexperiment:/NXentry/NXinstrument/NXdetector".
```
sanbrock, line 797: this is called classpath, but this can be ambiguous if two concept of the same type exists next two one another. This is why the NeXus vocabulary follows the path we call html name which is unique for a given concept. E.g. NXexperiment:/ENTRY/INSTRUMENT/mydetector. Note that this does not hold the information of the nexus classes referened along the path, so if it is needed, a classpath needs to be returned, too.

lukaspie:
@mkuehbach @rettigl I made some changes here, following the changes to the `NexusNode` object. In principle, I believe this would be ready. The bigger question is _if_ we want to have such a feature. I think since it is optional and only turned on by passing a flag to `pynx convert`, it is relatively harmless and a nice feature to add. But happy to hear your opinions.

## What changed (since rebase onto latest `master`)

- **`writer.py`**: `--write-docs` embeds NeXus concept docstrings as HDF5 string attributes (`@docs`, `@docs_url`, `@<attr>_docs`, `@<attr>_docs_url`) using `NexusNode.get_link()` for the URL and `NexusNode.get_inheritance_concept_paths()` for inheritance-aware doc text.
- **`cli.py` / `convert.py`**: `--write-docs` flag and `--docs-format` option (`default` = raw RST, `plain` = markup-stripped single-line, optimised for h5web).
- **`annotator.py`**: `pynx read` silently skips `@docs`/`@docs_url`/`@*_docs`/`@*_docs_url` attributes to keep output uncluttered.
- **`NXtest.nxdl.xml`**: added `<doc>` to the `version` attribute for test coverage.
- **Tests**: `test_write_docs` parametrized over both formats; `plain` asserts no embedded `\n`.
- **Docs**: new section in `dataconverter-and-readers.md` with usage, format table, and a `!!! warning` on link semantics (write-time and read-time). -> this addresses the tension raised by Sandor in https://github.com/FAIRmat-NFDI/pynxtools/pull/415#issuecomment-2326316890.

## Not yet addressed / open design questions

1. **Repeated concepts** — every instance of the same concept gets its own copy of the docs string. The suggestion was to write docs once and softlink; not implemented (architecturally non-trivial, requires a parallel `/.nexus_docs/` structure or concept-level deduplication).

2. **Link/copy tension** — `--write-docs` intentionally skips HDF5 link nodes (soft link, virtual dataset, external link) to avoid attaching the wrong concept's docs. The alternative (@rettigl: copy data instead of linking when `--write-docs` is active) was discussed but not implemented. Currently only documented with a warning.

3. **HTML rendering** — dropped: NXDL docs use Sphinx RST (`:ref:`, `.. include::`) that docutils cannot resolve standalone. Only `default` and `plain` are offered. `plain` is however good enough that in an h5web viewer one can read the docs sufficiently well:
image

PR #417, opened 2024-09-02
Map to NOMAD basesections

lukaspie:


lukaspie:
Closed in favor of #419

PR #418, opened 2024-09-04
To allow entry_type would be written or not.

RubelMozumder:
The kwarg `wrte_entry_type` is used in xrd in `nomad_measurement`.

Currently, `populate_nexus_subsection` is being called in a step before running the `normaliser` function of the schema (`ELNXrayDiffraction`). So, the `entry_type` is being written in the `populate_nexus_subsection` and when `nomad` sees the `attr` is filled, `nomad` does not update the value of the `entry_type` which would be `ELNXrayDiffraction`. This is problematic, when the external schema like `ELNXrayDiffraction` uses a `nexus` dataconverter in a internal process.

### DIFF
--- a/src/pynxtools/nomad/dataconverter.py
+++ b/src/pynxtools/nomad/dataconverter.py
@@ -66,6 +66,7 @@ def populate_nexus_subsection(
     output_file_path: Optional[str] = None,
     on_temp_file=False,
     nxs_as_entry=True,
+    write_entry_type=True,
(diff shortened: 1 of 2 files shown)


lukaspie (approved):
lgtm

RubelMozumder:
This was needed for XRD, but we are using  entry_type. Now Nexus can also be there as and extra entry where entry type is not a problem.

PR #419, opened 2024-09-06
New attempt: map to NOMAD basesections

lukaspie:
**EDIT: this is the branch that we want to work on going forward**
~~Made a new branch so that we don't loose too much on the other branch.~~

- `NXobject` does not inherit any `BaseSection` (and thus does not have normalization)
- We don't create any new classes for the normalization any more, but only add `normalize functions` (using `super(self.__class__, self)`). This should work since we now don't run into any recursion problems anymore.
- All non NX-object NeXus classes inherit from one of the existing base sections once again


lukaspie:
Closed in favor of #427

PR #428, opened 2024-09-11
Enhanced binary strings decoding

lukaspie:
Following up on discussions in https://github.com/FAIRmat-NFDI/nexus_definitions/pull/283#issuecomment-2343573429


src/pynxtools/nexus/nexus.py
```diff
@@ -31,13 +32,93 @@
...
+
+    Raises:
+        ValueError: If `string_obj` is not one of the supported types or decoding fails.
+    """
+    if not decode:
+        # Return the initial value without decoding
+        if not isinstance(string_obj, (np.ndarray, bytes, str)):
+            raise ValueError(
+                f"Unsupported type {type(string_obj)}. Expected np.ndarray, bytes, or str."
+            )
```
rettigl, line 59: This does not work simply like this, because I fed every value through the decode function before, i.e. also numerical ones.

lukaspie: removed the type checking, now the decoding only happens on
- bytes objects -> returns list
- list of bytes or mixed bytes/str -> returns list of str
- np.arrays of of bytes or mixed bytes/str -> returns np.arrays of str

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "src/pynxtools/definitions"]
 	path = src/pynxtools/definitions
-	url = https://github.com/FAIRmat-NFDI/nexus_definitions.git
\ No newline at end of file
+	url = https://github.com/FAIRmat-NFDI/nexus_definitions.git
+	branch = strict-string-check
```
lukaspie, line 4: reminder to remove

src/pynxtools/nexus/nexus.py
```diff
@@ -27,16 +29,93 @@
...
+        # Handle mixed-type arrays
+        if elem.dtype == object:
+            decoded_array = np.vectorize(
+                lambda x: x.decode(encoding) if isinstance(x, bytes) else x
+            )(elem)
+            return decoded_array.astype(str)  # Ensure all elements are strings
+
+    return decode_or_not(elem, encoding, decode)
+
+    # # Handle lists of bytes or strings
```
sanbrock, line 93: Are arrays of strings covered? or would they be covered by the commented section?

src/pynxtools/nexus/nexus.py
```diff
@@ -27,16 +29,93 @@
...
+    # # For other types, just return the input value unchanged
+    # return elem
+
+
 def get_nxdl_entry(hdf_info):
     """Get the nxdl application definition for an HDF5 node"""
     entry = hdf_info
     while (
         isinstance(entry["hdf_node"], h5py.Dataset)
         or "NX_class" not in entry["hdf_node"].attrs.keys()
```
sanbrock, line 117: should not the attribute names be also checked if they were written by a different encoding?

rettigl: I did not encounter this so far. I have only seen string names from h5 files. Here, it also says so: https://stackoverflow.com/questions/69208885/accessing-hdf5-key-value-store-using-integer-keys

sanbrock: OK

src/pynxtools/nexus/nexus.py
lukaspie: Since I could not just copy all the functionality to pynxtools and just call the function in the nexus_defintions, I'll ended up splitting the funcion. Bytes and list of bytes are convered by the `decode_or_not` function which comes from the dev_tools. The commented part can thus be removed.

src/pynxtools/nexus/nexus.py
lukaspie: Agreed, I think if you put a bytes object as a dataset key, h5py would complain.

.gitmodules
lukaspie: done

lukaspie:
@rettigl @sanbrock this branch should be finished now

mkuehbach (approved):

src/pynxtools/nexus/nexus.py
```diff
@@ -27,16 +29,75 @@
     try_find_units,
     walk_elist,
     write_doc_string,
+    decode_or_not,
 )
 
 
+from typing import Union, Optional, Any
```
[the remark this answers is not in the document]
lukaspie: done

src/pynxtools/nexus/nexus.py
```diff
@@ -27,16 +29,65 @@
...
+        return elem
+
+    # Handle numpy arrays of bytes or strings
+    if isinstance(elem, np.ndarray):
+        if elem.size == 0:
+            return elem  # Return the empty array unchanged
+
+        if elem.dtype.kind == "S":  # Check if it's a bytes array (fixed-length strings)
+            decoded_array = np.vectorize(
+                lambda x: x.decode(encoding).rstrip("\x00")
```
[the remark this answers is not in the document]
lukaspie: added

PR #435, opened 2024-09-19
Extend nx type

RubelMozumder:
1. Extend: NX type by `NX_CHAR_OR_INT` 
2. Include XRD in the list of plugin tests.

### DIFF
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -603,6 +603,7 @@ NEXUS_TO_PYTHON_DATA_TYPES = {
     ),  # > 0 is checked in is_valid_data_field()
     "NX_COMPLEX": (complex, np.ndarray, np.cdouble, np.csingle),
     "NXDL_TYPE_UNAVAILABLE": (str,),  # Defaults to a string if a type is not provided.
+    "NX_CHAR_OR_NUMBER": (str, int, float, np.ndarray, np.char.chararray),
 }
 
 
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -577,23 +577,47 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
 NUMPY_FLOAT_TYPES = (np.half, np.float16, np.single, np.double, np.longdouble)
 NUMPY_INT_TYPES = (np.short, np.intc, np.int_)
 NUMPY_UINT_TYPES = (np.ushort, np.uintc, np.uint)
-
+# np int for np version 1.26.0
+np_int = (
+    np.intc,
+    np.int_,
+    np.intp,
+    np.int8,
+    np.int16,
+    np.int32,
+    np.int64,
+    np.uint8,
+    np.uint16,
+    np.uint32,
+    np.uint64,
+    np.unsignedinteger,
+    np.signedinteger,
+)
+np_float = (np.float16, np.float32, np.float64, np.float64, np.floating)
+np_bytes = (np.bytes_, np.byte, np.ubyte)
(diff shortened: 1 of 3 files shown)


src/pynxtools/dataconverter/helpers.py
```diff
@@ -603,6 +603,7 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
     ),  # > 0 is checked in is_valid_data_field()
     "NX_COMPLEX": (complex, np.ndarray, np.cdouble, np.csingle),
     "NXDL_TYPE_UNAVAILABLE": (str,),  # Defaults to a string if a type is not provided.
+    "NX_CHAR_OR_NUMBER": (str, int, float, np.ndarray, np.char.chararray),
```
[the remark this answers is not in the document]
RubelMozumder: @lukaspie, I update the all type np type which is easy to maintain, if in future is there any new nx datatype. I checked the all the types
```python
def is_dtype_valid(dtype):
    try:
        np.dtype(dtype)
        return True
    except TypeError:
        return False
   ```
To see al the data type are correctly annotated. Would you recheck in your environment?

File: src/pynxtools/dataconverter/helpers.py [hunk 2]
```diff
@@ -577,32 +577,66 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+    np.int32,
+    np.int64,
+    np.uint8,
+    np.uint16,
+    np.uint32,
+    np.uint64,
+    np.unsignedinteger,
+    np.signedinteger,
+)
+np_float = (np.float16, np.float32, np.float64, np.float64, np.floating)
```
lukaspie, line 596: You have np.float64 twice

src/pynxtools/dataconverter/helpers.py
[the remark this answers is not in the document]
lukaspie: looks good, I tested it

src/pynxtools/dataconverter/helpers.py
```diff
@@ -577,32 +577,66 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+    np.int_,
+    np.intp,
+    np.int8,
+    np.int16,
+    np.int32,
+    np.int64,
+    np.uint8,
+    np.uint16,
+    np.uint32,
+    np.uint64,
```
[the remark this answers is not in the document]
RubelMozumder: Currently, we do not have any restriction on the numpy type.

src/pynxtools/dataconverter/helpers.py
RubelMozumder: removed!

RubelMozumder:
> LGTM.
> 
> Will we have to change the np types when going to numpy v2?

Currently no need. As nomad is pinned around 1.26. Let what comes in future.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -577,32 +577,66 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+    np.unsignedinteger,
+    np.signedinteger,
+)
+np_float = (np.float16, np.float32, np.float64, np.floating)
```
rettigl, line 596: @lukaspie @RubelMozumder 
Cant't you just use np.integer and np.floating here directly? I thought all the specific types are subtypes of this.
https://numpy.org/doc/stable/reference/arrays.scalars.html

RubelMozumder: @rettigl, you are right, I was also thinking, there might be a inheritance interface among the type. I did not give that much focus on it. I also see there are some type that are not in the list of types in the figure. I am creating an issue on it, we can make a test in this case.

PR #446, opened 2024-10-01
Package title in citation

lukaspie:
Would make a new release v0.7.4 afterwards

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,10 +1,10 @@
 cff-version: 1.2.0
-title: A Python Library for NeXus-Compliant Experimental Data Conversion and Integration with NOMAD Platform
+title: "pynxtools: A Python Library for NeXus-Compliant Experimental Data Conversion and Integration with NOMAD Platform"
 message:
   If you use this software, please cite it using the
   metadata from this file.
 type: software
-version: 0.7.3
+version: 0.7.4
 authors:
   - given-names: Sherjeel
     family-names: Shabih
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -1,10 +1,10 @@
 cff-version: 1.2.0
-title: A Python Library for NeXus-Compliant Experimental Data Conversion and Integration with NOMAD Platform
+title: 'pynxtools: A Python Library for NeXus-Compliant Experimental Data Conversion and Integration with NOMAD Platform'
```
RubelMozumder, line 2:
```suggestion
title: 'PYNXTOOLS: A Python Library for NeXus-Compliant Experimental Data Conversion and Integration with NOMAD Platform'
```

RubelMozumder (changes requested):

CITATION.cff
lukaspie: Why? The package is called pynxtools (lower case)

RubelMozumder: Both are correct. We should follow the same conversion as FAIRmat community is using, or what do you think? I see that all of the publishes are using capitalization of the name of repo e.g NOMAD Simulation or Nomad Measurent Plugin. 
Actually, I am considering the `Pynxtools` at the starting as a tag to the title.

lukaspie: This is a special cases becaus `NOMAD` is always uppercase (because it is an abbreviation). But `pynxtools` or `nyaml` doesn't have to be.

RubelMozumder: I meant `NOMAD Simulations`, `NOMAD Material Processing Plugin`, where `Simulations` `MAterial` ... are Capitalized. 
Link: https://zenodo.org/communities/fairmat_nfdi/records?q=&f=resource_type%3Asoftware&l=list&p=1&s=10&sort=newest

lukaspie: Hm, maybe `Pynxtools` is indeed okay, e..g `numpy` is also called `NumPy` on their docs. Let's go with `Pynxtools` and `Nyaml´ then.

RubelMozumder (approved):
LGTM!

PR #447, opened 2024-10-08
Adds a logic to match NX class at the same level when no base_section…

sherjeelshabih:
This fixes the issue with voltage_controller/calibration_time in NXiv_temp. Another error shows up with /entry/user/name.

@sanbrock, Lev will have a look into this. When you get time, please check if this makes sense with NeXus inheritance rules generally.

- [ ] Use Florian's NexusTree implementation to generate metainfo


lukaspie:
Since the metainfo generation has changed a lot since this was started, I think we should close this PR (and make a new one if this is still an issue). What do you think @sherjeelshabih?

PR #448, opened 2024-10-09
Move NOMAD examples to plugin

lukaspie:
- [x] build entrypoints 
- [x] include in MANIFEST.in
- [x] include in pyproject.toml
- [x] check if all plugins are really up-to-date -> there may be updates in other NOMAD branches that are not included here.
- [x] decide whether some of the examples or at least some of the data shall be downloaded instead of stored in Git
- [x] can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be tricky
  **EDIT**: we can just have the data and the nomad plugin entrypoint within the plugins themselves
- [x] test for example data and `ExampleUploadEntryPoint`
- [x] test within NOMAD


src/pynxtools/nomad/entrypoints.py
```diff
@@ -43,3 +47,110 @@ def load(self):
...
+    description="""
+        This example presents the capabilities of the NOMAD platform to store and standardize atom probe data.
+        It shows the generation of a NeXus file according to the
+        [NXapm](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXapm.html#nxapm)
+        application definition and a successive analysis of an example data set.
+        The example contains a small atom probe dataset from an experiment with a LEAP instrument to get you started
+        and keep the size of your NOMAD installation small. Once started, we recommend changing the respective
+        input file in the NOMAD Oasis ELN to run the example with your own datasets.
+    """,
+    path="nomad/examples/apm",
```
lukaspie, line 63: Instead of doing this, the best way would probably be to pull the examples from north directly. One can pass the `url` of a zip file to `ExampleUploadEntryPoint` (see [docs](https://nomad-lab.eu/prod/v1/staging/docs/reference/plugins.html#exampleuploadentrypoint)). Just need a way to zip the example folder from NORTH.

rettigl:
> * can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be trickier than

Such a solution would from my point of view be the most logical place to locate these examples, i.e. together with each reader. In any case, we should follow a SSOT approach for these examples.

lukaspie:
> > * can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be trickier than
> 
> Such a solution would from my point of view be the most logical place to locate these examples, i.e. together with each reader. In any case, we should follow a SSOT approach for these examples.

Absolutely, because as of now, we not only need to update them in NORTH and in nomad-FAIR, but also in all distribution branches in `nomad-distro`. Can we not try to download them from NORTH? We need them there anyway because the CI/CD checks that the examples work with the installed image.

rettigl:
> > > * can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be trickier than
> > 
> > 
> > Such a solution would from my point of view be the most logical place to locate these examples, i.e. together with each reader. In any case, we should follow a SSOT approach for these examples.
> 
> Absolutely, because as of now, we not only need to update them in NORTH and in nomad-FAIR, but also in all distribution branches in `nomad-distro`. Can we not try to download them from NORTH? We need them there anyway because the CI/CD checks that the examples work with the installed image.

It feels a bit strange having them in NORTH at all, because there is not a 1:1 mapping of NORTH tools and pynxtools plugins, i.e. there is not a tool for every example/plugin, no? Maybe one could create another example repo with sub-repos for each example, which then gets submoduled where needed? However, then one still needs to update the submodule commits all the time, so also not ideal.

lukaspie:
> > > > * can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be trickier than
> > > 
> > > 
> > > Such a solution would from my point of view be the most logical place to locate these examples, i.e. together with each reader. In any case, we should follow a SSOT approach for these examples.
> > 
> > 
> > Absolutely, because as of now, we not only need to update them in NORTH and in nomad-FAIR, but also in all distribution branches in `nomad-distro`. Can we not try to download them from NORTH? We need them there anyway because the CI/CD checks that the examples work with the installed image.
> 
> It feels a bit strange having them in NORTH at all, because there is not a 1:1 mapping of NORTH tools and pynxtools plugins, i.e. there is not a tool for every example/plugin, no? Maybe one could create another example repo with sub-repos for each example, which then gets submoduled where needed? However, then one still needs to update the submodule commits all the time, so also not ideal.

I think what we could do is have the pynxtools plugins be the SSOT place that have these folders with the example files that shall be used in NOMAD. Then instead of having the files also in NORTH, we can just get them from GitHub in the NORTH CI/CD for testing. And for the NORTH tools that don't have a directly correspondant plugin, we just don't use this.

Then for the NOMAD examples, we use the `url ` option in `ExampleUploadEntryPoint` (see [docs](https://nomad-lab.eu/prod/v1/staging/docs/reference/plugins.html#exampleuploadentrypoint)) and point this to the URL of the examples in the repo.

lukaspie:
> > > > > * can we have a mechanism so that the pynxtools plugins have their own nomad examples and these are brought into the central pynxtools only if the plugin is installed? This will likely be trickier than
> > > > 
> > > > 
> > > > Such a solution would from my point of view be the most logical place to locate these examples, i.e. together with each reader. In any case, we should follow a SSOT approach for these examples.
> > > 
> > > 
> > > Absolutely, because as of now, we not only need to update them in NORTH and in nomad-FAIR, but also in all distribution branches in `nomad-distro`. Can we not try to download them from NORTH? We need them there anyway because the CI/CD checks that the examples work with the installed image.
> > 
> > 
> > It feels a bit strange having them in NORTH at all, because there is not a 1:1 mapping of NORTH tools and pynxtools plugins, i.e. there is not a tool for every example/plugin, no? Maybe one could create another example repo with sub-repos for each example, which then gets submoduled where needed? However, then one still needs to update the submodule commits all the time, so also not ideal.
> 
> I think what we could do is have the pynxtools plugins be the SSOT place that have these folders with the example files that shall be used in NOMAD. Then instead of having the files also in NORTH, we can just get them from GitHub in the NORTH CI/CD for testing. And for the NORTH tools that don't have a directly correspondant plugin, we just don't use this.
> 
> Then for the NOMAD examples, we use the `url ` option in `ExampleUploadEntryPoint` (see [docs](https://nomad-lab.eu/prod/v1/staging/docs/reference/plugins.html#exampleuploadentrypoint)) and point this to the URL of the examples in the repo.

I still think we need to define the `ExampleUploadEntryPoint` instances directly in pynxtools and cannot delegate that to the plugins because you need to declare the nomad plugin entry-points in the pyproject.toml of the nomad plugin itself, like so: 
```
[project.entry-points.'nomad.plugin']
apm_example = "pynxtools.nomad.entrypoints:apm_example"
ellips_example = "pynxtools.nomad.entrypoints:apm_example"
em_example = "pynxtools.nomad.entrypoints:em_example"
```

rettigl:
This sounds like a reasonable approach. With the requirements for the entry points, I am not sure I understand this well enough

docs/how-tos/using-pynxtools-test-framework.md
```diff
@@ -60,3 +60,5 @@ You can also pass additional parameters to `test.convert_to_nexus`:
 - `caplog_level` (str): Can be either "ERROR" (by default) or "warning". This parameter determines the level at which the caplog is set during testing. If it is "WARNING", the test will also fail if any warnings are reported by the reader.
 
 - `ignore_undocumented` (boolean): If true, the test skipts the verification of undocumented keys. Otherwise, a warning massages for undocumented keys is raised
+
+## How to write an integration test for a NOMAD example in a reader plugin
```
lukaspie, line 64: This needs to be updated when it's working

lukaspie:
What I hadn't figured out: we can just have the entry points for the examples in the individual plugins, see https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/32/. That makes it so much cleaner because the data can just live in the plugin repo and we don't need to use the `url` for the entry point.

.github/workflows/plugin_test.yaml
```diff
@@ -22,31 +22,31 @@ jobs:
         include:
 
           - plugin: pynxtools-ellips
-            branch: main
+            branch: bring-in-examples
```
lukaspie, line 25: reminder to revert

src/pynxtools/nomad/entrypoints.py
```diff
@@ -43,3 +64,38 @@ def load(self):
...
+    title="Scanning Tunneling Spectroscopy (STS)",
+    category="FAIRmat examples",
+    description="""
+        This example is for two types of experiments: Scanning Tunneling Microscopy (STM) and Scanning Tunneling Spectroscopy (STS) from Scanning Probe Microscopy.
+        It can transform the data from files generated by a the nanonis software into the NeXus application definition NXsts.
+        The example contains data files from the two specific nanonis software versions generic 5e and generic 4.5.
+    """,
+    url="https://download-directory.github.io/?url=https://github.com/FAIRmat-NFDI/pynxtools-stm/tree/main/examples/nomad/sts",
+    local_path="examples/data/uploads/sts.zip",
+)
```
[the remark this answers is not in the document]
lukaspie: I removed the stm/sts examples here, shall be added in `pynxtools-stm`.

src/pynxtools/testing/nomad_example.py
```diff
@@ -0,0 +1,71 @@
...
+        "scheme.archive.yml",
+        "schema.archive.json",
+        "scheme.archive.json",
+        "intra-entry.archive.json",
+    )
+    path = os.walk(os.path.join(os.getcwd(), example_path))
+    for root, _, files in path:
+        for file in files:
+            if os.path.basename(file).endswith(example_files):
+                yield pytest.param(os.path.join(root, file), id=file)
```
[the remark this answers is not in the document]
lukaspie: Why? You can pass a tuple directly to ´str.endswith`?

src/pynxtools/testing/nomad_example.py
```diff
@@ -0,0 +1,71 @@
...
+    example_files = (
+        "schema.archive.yaml",
+        "schema.archive.yml",
+        "scheme.archive.yaml",
+        "scheme.archive.yml",
+        "schema.archive.json",
+        "scheme.archive.json",
+        "intra-entry.archive.json",
+    )
+    path = os.walk(os.path.join(os.getcwd(), example_path))
```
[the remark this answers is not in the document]
lukaspie: I made it just a str, the actual path handling can be done in the code that calls this, see e.g. [here](https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/44b8ab12723fe97a34571e6651c7f46b50af658b/tests/test_nomad_examples.py#L47) for the mpes reader.

src/pynxtools/nomad/examples/iv_temp/README.md
```diff
@@ -0,0 +1,30 @@
...
+The ELN follows the general structure of NOMAD ELN templates and you may refer to the [documentation](https://nomad-lab.eu/prod/v1/staging/docs/archive.html) or a [YouTube tutorial](https://youtu.be/o5ETHmGmnaI) (~1h)
+for further information.
+When the ELN is saved a NeXus file will be generated from the provided example data.
+You may also view your supplied or generated NeXus files here with the H5Web viewer.
+To do so open the **FILES** tab and just select a `.nxs` file.
+
+# Using a Jupyter Notebook
+
+This example comes with a very simple Jupyter Notebook that shows how one could easily get access to a Python environment with access to all your data in one place.
+To give this a go, click the **FILES** tab and select `iv_temp.ipynb`. Feel free to modify this or just create a new one to try!
```
RubelMozumder, line 24:
```suggestion
To give this a go, click the **FILES** tab and select `IV_temp.ipynb`. Feel free to modify this or just create a new one to try!
```

src/pynxtools/testing/nomad_example.py
```diff
@@ -0,0 +1,129 @@
...
+    # Check if the provided path exists
+    if not os.path.exists(example_path):
+        raise FileNotFoundError(f"The directory '{example_path}' does not exist.")
+
+    # Walk through the specified directory
+    for root, _, files in os.walk(example_path):
+        for file in files:
+            normalized_file = file.lower()  # Normalize to lower case
+            if os.path.basename(normalized_file).endswith(example_files):
+                yield pytest.param(os.path.join(root, file), id=file)
```
RubelMozumder, line 69: As far as I can remember, I had issue with the keyoward `yield` while running test in `pynxtools-stm`. Would you check it if there was something that I missed? Other than that it is fine.

RubelMozumder (approved):
LGTM!

src/pynxtools/testing/nomad_example.py
lukaspie: I had to change the structure of the pytest params in `pynxtools-stm` a bit (basically converting to a list, see https://github.com/FAIRmat-NFDI/pynxtools-stm/blob/7ee43248ee4cf74b5032138c21016cdf6b454f77/tests/test_nomad_examples.py#L52), but then it works.

src/pynxtools/nomad/examples/iv_temp/README.md
```diff
@@ -0,0 +1,30 @@
+
+
+# Introduction
+
+This is an example of a PID controlled sensor sweep scan. The temperature is set using a PID controller. Then for the set temperature a voltage sweep is performed. For each of the voltages, the current is measured. This is repeated for a given list of temperatures.
```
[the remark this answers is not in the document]
lukaspie: done

src/pynxtools/nomad/examples/iv_temp/README.md
lukaspie: changed

docs/how-tos/using-pynxtools-test-framework.md
lukaspie: done

src/pynxtools/testing/nomad_example.py
lukaspie: The only reason why you need to use list in pynxxtools-stm is that you want to concatenate the two generator outputs of this function because you run the same tests using two different folders. Actualy, it may even be better if you split the test [here](https://github.com/FAIRmat-NFDI/pynxtools-stm/blob/57d96d25efdef84b37fb945364ef17470bb6c36a/tests/test_nomad_examples.py#L44) into two different tests. In none of the other repos we have to use the list conversion, so I think it's fine with the `yield` keyword.

By the way, the reason why I am using this is that NOMAD uses the same functionality: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/tests/parsing/test_archive_parser.py#L57

PR #449, opened 2024-10-09
Simple modification.

RubelMozumder:


lukaspie:
As the ELN generator has been substantially changed, I think this PR is probably not needed anymore. What do you think @RubelMozumder?

RubelMozumder:
Resolved in PR #607

PR #451, opened 2024-10-09
replace BasicEln

sanbrock:


src/pynxtools/nomad/utils.py
```diff
@@ -72,6 +72,7 @@ def __rename_nx_for_nomad(
 
     if name and name.startswith("NX"):
         name = __REPLACEMENT_FOR_NX + name[2:]
+        name = name[0].upper() + (name[1:] if len(name) > 1 else "")
```
lukaspie, line 75: Why not ```name = name[:1].upper() + name[1:]```? I don't think we need the len check.

PR #452, opened 2024-10-17
Add notes to How-To to use dev install for verify_nexus

RonHildebrandt:
Reason is, the default installation of the branch did not work. The command:

```
pip install git+https://github.com/FAIRmat-NFDI/pynxtools@hdf-based-validation
```
resulted in:
```
WARNING: Built wheel for pynxtools is invalid: Wheel has unexpected file name: expected '0.6.0.post1.dev29+gde56834', got '0.6.0.post1.dev29+gde56834.d20241017'
Failed to build pynxtools
```
It seems somehow, that a different commit version is required.
With the addition of the development install, the user at least can now use this directly (as this will be linked in the Tutorial).

### DIFF
--- a/docs/how-tos/validate-nexus-file.md
+++ b/docs/how-tos/validate-nexus-file.md
@@ -120,6 +120,23 @@ Options:
   --help  Show this message and exit.
 ```
 
+*Development verion installation*
+
+If this installation procedure above does not work, you can use the devlopment installation by using git:
+```
+python -m venv .py39
+source .py39/bin/activate
+git clone https://github.com/FAIRmat-NFDI/pynxtools.git
+cd pynxtools/
+git checkout hdf-based-validation
+git submodule sync –recursive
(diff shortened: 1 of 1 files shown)


docs/how-tos/validate-nexus-file.md
```diff
@@ -120,6 +120,23 @@ Options:
   --help  Show this message and exit.
 ```
 
+*Development verion installation*
```
rettigl, line 123:
```suggestion
*Development version installation*
```

docs/how-tos/validate-nexus-file.md
```diff
@@ -120,6 +120,23 @@ Options:
...
+
+If this installation procedure above does not work, you can use the devlopment installation by using git:
```
rettigl, line 125:
```suggestion
If this installation procedure above does not work, you can use the development installation by using git:
```

PR #454, opened 2024-10-21
bump ruff to >v0.6

lukaspie:
Same version as in NOMAD


RubelMozumder (approved):
LGTM!

rettigl:
Also here: Linting CI/CD is still being run with ruff=0.5.5

lukaspie:
Yeah, you are right, I didn't remember that we are actually installing the `dev-requirements` instead of doing `uv pip install -e ."[dev]"`.

I updated the `dev-requirements`, now it should be okay

PR #455, opened 2024-10-28
carefull name generation for new BasicELN archive and population NXroot

sanbrock:


sanbrock:
> I am not sure what exactly is now happening in lines 911-914 of src/pynxtools/nomad/schema.py. NXroot related parts look good to me.

Previously, the filename was constructed directly from lab_id. Since it can contain non-allowed characters (e.g. ':'), if it happens, a unique, but conform filename is generated here.

GinzburgLev:
> > I am not sure what exactly is now happening in lines 911-914 of src/pynxtools/nomad/schema.py. NXroot related parts look good to me.
> 
> Previously, the filename was constructed directly from lab_id. Since it can contain non-allowed characters (e.g. ':'), if it happens, a unique, but conform filename is generated here.

I see; that makes sense now, thank you for clarification.

GinzburgLev (approved):

PR #457, opened 2024-10-31
Fix type hints

lukaspie:
The previous type hints were only working for `python>3.9` and were breaking the tests on`python>3.8` in the plugins.


RubelMozumder (approved):
LGTM!

PR #461, opened 2024-11-12
fix case if the handle function inside get_data is ill defined

RonHildebrandt:
I had the problem that, I renamed a function and forgot it at another position. This was the case, as the name in:

self.extensions = {".yml": self.handle_eln_file}
was different from the defined in the new created MultiformatReader.

Therefore: "self.extensions.get(extension, lambda _: {})(file_path)" was None and then failed in updateing the template. Not sure if this is needed to catch this case (should be avoidable if you know what you do).

### DIFF
--- a/src/pynxtools/dataconverter/readers/multi/reader.py
+++ b/src/pynxtools/dataconverter/readers/multi/reader.py
@@ -421,8 +421,10 @@ class MultiFormatReader(BaseReader):
             if not os.path.exists(file_path):
                 logger.warning(f"File {file_path} does not exist, ignoring entry.")
                 continue
-
-            template.update(self.extensions.get(extension, lambda _: {})(file_path))
+            if self.extensions.get(extension, lambda _: {})(file_path) is None:
+                logger.warning(f"Function for parsing file type with extension '{extension}' not found.")
+            if self.extensions.get(extension, lambda _: {})(file_path) is not None:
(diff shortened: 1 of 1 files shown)


RonHildebrandt:
This issue was related to "I have no Idea what I'm doing exactly" - Yea, I forgot to rename a function in my Multiformat reader, and was wondering about this.
(i.e. I set for a specific file type a specific definiton to handle this file, but I forgot to define it (or in my case, i did not rename both definitions)). NOt sure if you want to catch this case.

Think you can judge this better if it is necessary. Usually, this error should not occur

lukaspie:
@RonHildebrandt so you are saying this would occur if you were using e.g.
```python
self.extensions = {
    ".txt": self.handle_txt
}
```
and then `handle_txt` is not defined in the reader?

In that case, wouldn't there be an attribute/name error when you invoke the reader? So it wouldn't even run?

I still don't quite understand how `self.extensions.get(extension, lambda _: {})(file_path)` can ever be None. If the extension is not in `self.extensions`, this call would return the lambda function `lambda _: {}`.

RonHildebrandt:
@lukaspie : Yea, something like this was the case. But, as I pointed out: Most probably a unnecssary beginner error, as Ive now understood much better the reader.

Therefore, this is not necessary.

The PR can be closed.

PR #466, opened 2024-11-15
Entry type2

sanbrock:


src/pynxtools/nomad/parser.py
```diff
@@ -491,18 +491,25 @@ def parse(
                 app_sec = getattr(self.nx_root, app)
                 try:
                     app_entry = getattr(app_sec, "ENTRY")
+                    if len(app_entry) < 1:
+                        raise AttributeError()
                 except AttributeError:
                     app_entry = getattr(app_sec, "entry")
+                    if len(app_entry) < 1:
+                        raise AttributeError()
```
RubelMozumder, line 499: A quick question, app_entry is not being used anywhere. Or is it to use in the future?

sanbrock: it is only to check if exists, so if no real entry is found then we do not add a corresponding AppDef to the entry_type

RubelMozumder: Right! I missed line:499 Attribute error.

lukaspie, line 499: Could we make this a bit cleaner? Also would be nice to have some information in the error message.
```suggestion
                for attr_name in ("ENTRY", "entry"):
                    app_entry = getattr(app_sec, attr_name, None)
                    if app_entry and len(app_entry) >= 1:
                        break
                    else:
                        raise AttributeError("No valid entry found in the NeXus file.")
```

lukaspie (changes requested):
One minor comment

src/pynxtools/nomad/parser.py
sanbrock: The suggested code would raise exception even if ENTRY is not found, but we need to quit only if neither 'ENTRY' nor 'entry' is present and even in this case we want to catch this exception as it is not a real error at this point, but only an indication for what type of Experiment label to put into 'entry_type'.

lukaspie: Ok, let's keep your version then.

lukaspie (approved):
lgtm

PR #468, opened 2024-11-19
Allow kwargs in ReaderTest

lukaspie:
This would allows us to pass `config_file=<some-file>` to the `ReaderTest`, which makes it in line with how the readers work in general.

### DIFF
--- a/tests/dataconverter/test_readers.py
+++ b/tests/dataconverter/test_readers.py
@@ -29,6 +29,7 @@ from _pytest.mark.structures import ParameterSet
 from pynxtools.dataconverter.convert import get_names_of_all_readers, get_reader
 from pynxtools.dataconverter.helpers import generate_template_from_nxdl
 from pynxtools.dataconverter.readers.base.reader import BaseReader
+from pynxtools.dataconverter.readers.multi.reader import MultiFormatReader
 from pynxtools.dataconverter.template import Template
 from pynxtools.dataconverter.validation import validate_dict_against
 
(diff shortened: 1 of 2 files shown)


tests/dataconverter/test_readers.py
```diff
@@ -51,32 +51,12 @@ def get_all_readers() -> List[ParameterSet]:
...
-            "EmNionReader",
-        ):
-            readers.append(
-                pytest.param(
-                    reader, marks=pytest.mark.skip(reason="Missing test data.")
-                )
-            )
-        else:
-            readers.append(pytest.param(reader))
+        readers.append(pytest.param(reader))
```
RubelMozumder, line 55: Will the reader test take place for all the readers like STM, XPS and so on?
We have ci/cd to test that readers, right? No?

tests/dataconverter/test_readers.py
```diff
@@ -51,32 +51,12 @@ def get_all_readers() -> List[ParameterSet]:
...
     return readers
 
 
-@pytest.mark.parametrize("reader", get_all_readers())
-def test_if_readers_are_children_of_base_reader(reader):
-    """Test to verify that all readers are children of BaseReader"""
-    if reader.__name__ != "BaseReader":
-        assert isinstance(reader(), BaseReader)
-
-
```
RubelMozumder, line 79: Can't we keep it? Probably it is hampering multi-reader from Ron. We can also include the multi-reader in the assert statement.

tests/dataconverter/test_readers.py
lukaspie: Here, we don't have the reader plugins installed, so the deleted part was not doing anything anymore.

tests/dataconverter/test_readers.py
lukaspie: Actually, that was my plan, must have messed up the commits. I re-added and included the multi-reader in the assert statement.

RubelMozumder (approved):
LGTM!

PR #470, opened 2024-11-21
Small updates to multi reader

lukaspie:
This solves two problems: 
1) If, in your config file, you have a list with a wildcard like this `"peakPEAK[peak*]/data/@xes":["@attrs:component*/kx, @attrs:component*/ky"]`, all of these values in the list gets replace as for non-list cases.
2) If you have multiple options to resolve a wildcard value like 
`"peakPEAK[peak*]/intensity":"['@attrs:component*/data_cps', '@attrs:component*/data']"`, where the reader first tries to fill from the path `component*/data_cps` and then from `@attrs:component*/data'`, there was an issue with the dict iterator before if the first attempt at filling does not work. This is fixed here.


lukaspie:
> The second, with the .copy is most probably necessary, as here during iteration some changes were made? Anyway, I think its always saver todo it like this.

Exactly, that was the problem. Thanks for the review.

PR #474, opened 2024-12-01
initial skeleton for NeXus App

sanbrock:


src/pynxtools/nomad/schema.py
```diff
@@ -835,6 +836,7 @@ def init_nexus_metainfo():
     nexus_section = Section(
         validate=VALIDATE, name=__GROUPING_NAME, label=__GROUPING_NAME
     )
+    nexus_section.base_sections = [Schema.m_def]
```
sherjeelshabih, line 839: Just for future reference. Could you describe how this affects the metainfo generation?

src/pynxtools/nomad/schema.py
```diff
@@ -799,6 +799,7 @@ def __create_package_from_nxdl_directories(nexus_section: Section) -> Package:
                 SubSection(section_def=section, name=section.name)
             )
 
+    package.section_definitions.append(nexus_section)
```
sherjeelshabih, line 802: Just for future reference. Could you describe how this affects the metainfo generation?

src/pynxtools/nomad/schema.py
sanbrock: added

src/pynxtools/nomad/schema.py
sanbrock: added

PR #476, opened 2024-12-03
allow the use of wildcard definitions for the reader tests

rettigl:


rettigl:
This is needed e.g. for the igor reader that I am currently cleaning up and preparing for here: https://github.com/rettigl/pynxtools-igor

lukaspie:
Quick question: is the igor-reader then supposes to write to _any_ or to none of the application definitions? We should maybe allow only one of these options with this wildcard notation and indicate that in the docs as well.

rettigl:
> Quick question: is the igor-reader then supposes to write to _any_ or to none of the application definitions? We should maybe allow only one of these options with this wildcard notation and indicate that in the docs as well.

This wildcard notation is anyways supported by the data converter, see https://github.com/FAIRmat-NFDI/pynxtools/blob/cf4b792f6a743c96f11e639bdbe130aa4bd389ef/src/pynxtools/dataconverter/convert.py#L154
The Igor reader is intended to read from Igor ibw or pxp files, and allow to write to any appdef, or just NXroot

lukaspie (approved):

PR #481, opened 2024-12-09
Integrate pynxtools igor

lukaspie:


rettigl (approved):
Generally looks good to me, however I am wondering why no tests are failing, as there is no release of pynxtools-igor yet. Maybe we should include such a test (i.e. which installs pynxtools[convert])
The strategy for merging should be:
- Merge this
- Make a new pynxtools release (with #476)
- Fix requirements in pynxtools-igor to the new version
- Make first pynxtools-igor release

lukaspie:
Yes, we don't actually test which plugins are installable or installed. We can make a test for that (see #482). The plugin tests are anyway run from the github repos though.

PR #484, opened 2024-12-10
Connecting to base sections

sanbrock:


src/pynxtools/nomad/schema.py
```diff
@@ -913,6 +947,29 @@ def normalize_sample(self, archive, logger):
     super(current_cls, self).normalize(archive, logger)
 
 
+def normalize_entry(self, archive, logger):
+    """Normalizer for sample section."""
```
lukaspie, line 951: These docstrings (also the one below) are not correct

lukaspie (approved):
The content looks good in what it's trying to achieve. I am still unsure if connectin `NXentry` to `ActivityStep` is correct.

I left some comments on docstrings. Would also be good to have the changes in the PR summary to keep track later on.

src/pynxtools/nomad/schema.py
sanbrock: fixed

PR #486, opened 2024-12-16
using readonly instead of user_editability; enabling injesting multip…

sanbrock:
…le methods to the base classes and not only the normalization; Entry to inherit also from Task, so Workflow click drives there directly

### DIFF
--- a/src/pynxtools/nomad/schema.py
+++ b/src/pynxtools/nomad/schema.py
@@ -106,8 +106,8 @@ __BASESECTIONS_MAP: Dict[str, Any] = {
     "NXsample": [CompositeSystem],
     "NXsample_component": [Component],
     "NXidentifier": [EntityReference],
-    "NXentry": [ActivityStep, Task],
-    "NXprocess": [ActivityStep, Task],
+    "NXentry": [ActivityStep],  # , Task],
+    "NXprocess": [ActivityStep],  # , Task],
     "NXdata": [ActivityResult],
     # "object": BaseSection,
 }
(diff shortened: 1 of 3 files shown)


src/pynxtools/nomad/schema.py
```diff
@@ -105,8 +106,8 @@
     "NXsample": [CompositeSystem],
     "NXsample_component": [Component],
     "NXidentifier": [EntityReference],
-    "NXentry": [ActivityStep],
-    "NXprocess": [ActivityStep],
+    "NXentry": [ActivityStep],  # , Task],
```
lukaspie, line 109: This should eventually becomes its own workflow (nested in the workflow of the whole nxs file) containing its own Tasks. But fine for now

lukaspie (changes requested):
We should definitely also think how we can keep this clean and maintainable in the future. Maybe we can separate the actual schema creation a bit from the definition of new base section classes and normalization functions (i.e, move these to a new python file in the future)

src/pynxtools/nomad/schema.py
```diff
@@ -156,7 +156,11 @@ def normalize(self, archive, logger):
             archive.workflow2 = Workflow(name=self.name)
         # steps to tasks
         act_array = archive.workflow2.tasks
-        existing_items = {(task.name, task.section) for task in act_array}
+        existing_items = {
+            (task.name, task.section)
+            for task in act_array
+            if hasattr(task, "name") and hasattr(task.section, "section")
```
sanbrock, line 162: Task-s registered in archive.workflow2.tasks have both 'name' and 'section' defined. It is not a problem if they are None. In fact, we need also None-s be listed in such case.

src/pynxtools/nomad/schema.py
```diff
@@ -165,7 +169,11 @@ def normalize(self, archive, logger):
         act_array.extend(new_items)
         # samples to inputs
         act_array = archive.workflow2.inputs
-        existing_items = {(link.name, link.section) for link in act_array}
+        existing_items = {
+            (link.name, link.section)
+            for link in act_array
+            if hasattr(link, "name") and hasattr(link.section, "section")
```
sanbrock, line 175: same applies here

src/pynxtools/nomad/schema.py
```diff
@@ -175,7 +183,11 @@ def normalize(self, archive, logger):
 
         # results to outputs
         act_array = archive.workflow2.outputs
-        existing_items = {(link.name, link.section) for link in act_array}
+        existing_items = {
+            (link.name, link.section)
+            for link in act_array
+            if hasattr(link, "name") and hasattr(link.section, "section")
```
sanbrock, line 189: and here

src/pynxtools/nomad/schema.py
RubelMozumder: Initially, I got attribute errors. Later investigating the issue I see it was a bit older nomad develop branch.

PR #487, opened 2024-12-19
Bump ruff version.

RubelMozumder:


lukaspie:
Is there a reason to bump this other than keeping up to date with the latest version? I think so far we have tried to stay in line with what is used in NOMAD, but bumping to a newer version is totally fine for me too.

In any case, can you please also update the ruff version in the `.pre-commit.yaml` file?

lukaspie (approved):
LGTM

PR #512, opened 2025-01-10
replace nomad plugin logo

lukaspie:
@mkuehbach the new docs would look like this, is this what you wanted? Clicking on that icon also just brings you back to the home page of the docs now.
image

I also added a navigation in the footer and changed the favicon to the nomad plugins logo.


mkuehbach (commented):
Technically the changes look good to me.
I was not aware though that such new logo at all exists, we should maybe ask the management if we ought
are allowed to use that logo for pynxtools (and all of its FAIRmat-maintained pynxtools-plugins), 

i.e. **what is the usage policy when software tools are official NOMAD plugins to qualify for such logo?**

If we get the green light from the management, sure merge this in. Thank you.

lukaspie:
> i.e. **what is the usage policy when software tools are official NOMAD plugins to qualify for such logo?**

These are the logos that are used by all of the NOMAD plugins that are generated from the template (see [here](https://github.com/FAIRmat-NFDI/cookiecutter-nomad-plugin/tree/main/%7B%7Bcookiecutter.plugin_name%7D%7D/docs/assets)), so I think it's safe to say we can use them.

sanbrock (approved):
LGTM

PR #513, opened 2025-01-10
use pynxtools.nomad.schema.Root

sanbrock:


sanbrock:
https://github.com/FAIRmat-NFDI/pynxtools/pull/525 extends this one

Issue #393, opened 2024-07-28
[Bug]: Validation error when passing a "0" integer


rettigl:
### Contact Details

rettig@fhi-berlin.mpg.de

### What happened?

When a field with type "NX_FLOAT" gets filled by a "0" integer (rather than "0.0" float), the checking routine reports this key as undocumented. Specifically, the value is converted to float here:
https://github.com/FAIRmat-NFDI/pynxtools/blob/ef013699e2a5bda339a12f6f3b919b434058bb79/src/pynxtools/dataconverter/helpers.py#L668
This returns 0, which in turn makes this check fail:
https://github.com/FAIRmat-NFDI/pynxtools/blob/ef013699e2a5bda339a12f6f3b919b434058bb79/src/pynxtools/dataconverter/validation.py#L588
and hence the value ends up in undocumented. 
This is tested on the multi-format-reader, but should apply to main as well.

### Relevant log output

```shell
This is output from the Tutorial 3 of the sed package, which uses the mpes reader and pynxtools converter internally:

Input:
sp.binned.attrs["metadata"]["energy_calibration"]["tof"] = 0
sp.save(data_path + "/binned.nxs")

Output:
Using mpes reader to convert the given files:  
• ../sed/config/NXmpes_config.json
Field /ENTRY[entry]/PROCESS_MPES[process]/energy_calibration/original_axis written without documentation.
The output file generated: /mnt/pcshare/users/Laurenz/AreaB/sed/sed/tutorial/datasets/WSe2/binned.nxs.

Input: 
sp.binned.attrs["metadata"]["energy_calibration"]["tof"] = 0.0
sp.save(data_path + "/binned.nxs")

Output:
Using mpes reader to convert the given files:  
• ../sed/config/NXmpes_config.json
The output file generated: /mnt/pcshare/users/Laurenz/AreaB/sed/sed/tutorial/datasets/WSe2/binned.nxs.
```

GinzburgLev:
It seems that there are potentially two issue here, which are somewhat separated:

[...]

PR #522, opened 2025-01-16
Bug fix for "validation error when passing zero"

GinzburgLev:
Fix for the first part of this issue:

https://github.com/FAIRmat-NFDI/pynxtools/issues/393

I am unsure at the moment if the second is okay as it is now or if it is something that needs to be fixed too.


sanbrock (approved):
LGTM

PR #524, opened 2025-01-17
Draft: formula parsing in multiformatreader

### LINKED ISSUE #521: Add support for formulas/conversions in multiformat reader
Sometimes it would be helpful to have an "@formula" syntax to the multiformat reader, that e.g. allows unit conversion, or to calculate a number based on some other quantities in a nexus tree during conversion.

@lukaspie do you think this would be feasible to implement?

An example: We measure the average laser power and the repetition rate, from which one can automatically get the pulse energy.

lukaspie:
@rettigl this is an initial draft for what we were discussing in the PR.

The initial structure seems to work, but the actual formula evaluation doesn't yet, mostly because the regex for replacing with the key resolving is broken.

I'll continue working on this after my vacation (so from the 27th).


lukaspie:
> Is this still hot?

I want to come back to this at some point soon-ish, so please leave the PR open.

PR #527, opened 2025-01-23
Except unit uni pint quantity.

RubelMozumder:


RubelMozumder:
Involved with the PRs
- [ ] Fix XRD: https://github.com/FAIRmat-NFDI/pynxtools-xrd/pull/17
- [ ] Fix XPS: 
- [ ] Fix Igor:

src/pynxtools/dataconverter/writer.py
```diff
@@ -259,8 +259,10 @@ def _put_data_into_hdf5(self):
 
         def add_units_key(dataset, path):
             units_key = f"{path}/@units"
-            if units_key in self.data.keys() and self.data[units_key] is not None:
-                dataset.attrs["units"] = self.data[units_key]
+            units = self.data.get(units_key, None)
+            units = str(units) if isinstance(units, pint.Unit) else units
```
lukaspie, line 263: Why not use the `pint.UnitRegistry` (like in NOMAD)? See how I am using it in the pynxtools-xps [here](https://github.com/FAIRmat-NFDI/pynxtools-xps/blob/main/src/pynxtools_xps/reader_utils.py#L553).

lukaspie: Actually, using just `if isinstance(units, pint.Unit)` does not seem to work.

If I do the following
```
import pint

units = "m"
print(isinstance(units, pint.Unit))
```
that prints `False`, so obviously this unit check does not work as intended.

`pint.Unit` is just a class to implement a unit, not something to be used directly to check a string is a pint unit. We need to use `pint.UnitRegistry`, like in NOMAD.

src/pynxtools/dataconverter/writer.py
```diff
@@ -259,8 +259,22 @@ def _put_data_into_hdf5(self):
...
+                units = str(units)
+            else:
+                try:
+                    ureg.Unit(units)
+                except pint.errors.UndefinedUnitError as exc:
+                    massage = (
+                        f"Units provided for path: '{path}@units' are not valid."
+                        f" Please provide a valid unit."
+                    )
+                    raise InvalidDictProvided(massage) from exc
```
lukaspie, line 275: Shall this raise an error or simply be logged as a warning? The file can be written without this wrong unit, right?

src/pynxtools/dataconverter/writer.py
```diff
@@ -259,8 +259,22 @@ def _put_data_into_hdf5(self):
...
+            if units is None:
+                return
+            if isinstance(units, pint.Unit):
+                units = str(units)
+            else:
+                try:
+                    ureg.Unit(units)
+                except pint.errors.UndefinedUnitError as exc:
+                    massage = (
+                        f"Units provided for path: '{path}@units' are not valid."
```
lukaspie, line 272:
```suggestion
                        f"Units '{units}' provided for path: '{path}@units' are not valid."
```

src/pynxtools/dataconverter/writer.py
```diff
@@ -259,8 +259,22 @@ def _put_data_into_hdf5(self):
...
+            units = self.data.get(units_key, None)
+            if units is None:
+                return
+            if isinstance(units, pint.Unit):
+                units = str(units)
+            else:
+                try:
+                    ureg.Unit(units)
+                except pint.errors.UndefinedUnitError as exc:
+                    massage = (
```
lukaspie, line 271:
```suggestion
                    message = (
```

src/pynxtools/dataconverter/writer.py
lukaspie, line 275: Shall this raise an error or simply be logged as a warning? The file can be written without this wrong unit, right?
```suggestion
                    logger.warning(message) 
```

PR #528, opened 2025-01-23
update ruff version and format files.

RubelMozumder:
I had the same changes here #527 for clean history.


lukaspie:
Can you apply the same change to the pre-commit yaml?

Also, there is a version 0.9.3 of ruff now? Shall we upgrade this once more here?

RubelMozumder:
> Can you apply the same change to the pre-commit yaml?
> 
> Also, there is a version 0.9.3 of ruff now? Shall we upgrade this once more here?

Updated to 9.3.

lukaspie (approved):
LGTM

PR #537, opened 2025-02-05
add array size and ndim as attributes

rettigl:
@sanbrock 
I added here size and ndim as attributes for arrays, to be eventually able to search for them. If you agree, please integrate into no_inner_sections


sanbrock (approved):
LGTM

PR #538, opened 2025-02-06
remove mean from stats fields, and fix naming

rettigl:
Previously, the values and names where not matched correctly, I believe this fixes it. I removed mean, as this is set to the __field value anyways.


lukaspie:
This is now already implemented in `no_inner_sections` branch, so this PR is redundant and can be closed.

PR #553, opened 2025-02-18
Apm norm

sanbrock:


GinzburgLev (approved):
Discussed with Sandor; looks good to me.

Issue #440, opened 2024-09-27
[Bug]: Revisit np type


RubelMozumder:
### Contact Details

rubel.mozumder@outlook.com

### What happened?

Please follow PR #435. FYI @rettigl

RubelMozumder:
@rettigl I just tried to use the `np.integer` instead of the long list of numpy data type (discussed in PR #435, [here](https://github.com/FAIRmat-NFDI/pynxtools/pull/435#discussion_r1778164369])
```
In [28]: np_int = (
    ...:     np.intc,
    ...:     np.int_,
    ...:     np.intp,
    ...:     np.int8,
    ...:     np.int16,
    ...:     np.int32,
    ...:     np.int64,
    ...:     np.uint8,
    ...:     np.uint16,
    ...:     np.uint32,
    ...:     np.uint64,
    ...:     np.uint,
    ...:     np.unsignedinteger,
    ...:     np.signedinteger,
    ...: )
```
Though the type mentioned above is the subclass of `np.integer`, but instances of them should not be the install of `np.integer`. I provided the test result here to test it if you want.
```

In [28]: np_int = (
    ...:     np.intc,
    ...:     np.int_,
    ...:     np.intp,
    ...:     np.int8,
    ...:     np.int16,
    ...:     np.int32,
    ...:     np.int64,
    ...:     np.uint8,
    ...:     np.uint16,
    ...:     np.uint32,
    ...:     np.uint64,
    ...:     np.uint,
    ...:     np.unsignedinteger,
    ...:     np.signedinteger,
    ...: )

In [29]: 

In [29]: xx = "\n"
    ...: yy = "\n"
    ...: for x in np_int:
    ...:     x_data = np.array([4], dtype=x)
    ...: 
    ...: 
    ...:     yy += f'{x} : {isinstance(x_data, np.integer)}\n'
    ...:     xx += f'{x} : {issubclass(x, np.integer)} \n'
    ...: print('Check if instance np.integer::\n ', yy,'\n')
    ...: print("Check if subclass of np.integer :: ", xx)
<ipython-input-29-f3ce3646c671>:4: DeprecationWarning: Converting `np.unsignedinteger` to a dtype is deprecated. The current result is `np.dtype(np.uint)` which is not strictly correct. Note that the result depends on the system. To ensure stable results you may want to use `np.uint64` or `np.uint32`.
  x_data = np.array([4], dtype=x)
[...]

PR #554, opened 2025-02-18
Fix the nx_char type for numpy to  and .

RubelMozumder:
@lukaspie, you can also check this [issue](https://github.com/FAIRmat-NFDI/pynxtools/issues/440) with here.

Fix NX data type validation errors:
1. Issue1: The Data type of an element of a list or array is not being checked against the intended data type, e.g., NX_INT. (Status: Fixed)
2. Issue2: If possible, provided data is converted to the intended data type (e.g. "3" -> 3) and no warning is raised. (Status: Fixed, note: boolean conversion is allowed, e.g. `true`->True).


lukaspie:
What is this solving? The linked issue is about integers, this is about strings/bytes.

Why don't we allow chararrays anymore?

rettigl:
Please see #555 and the discussed way for solving this. We should remove arrays generally, and check the array dtype instead.

RubelMozumder:
> What is this solving? The linked issue is about integers, this is about strings/bytes.
> 
> Why don't we allow chararrays anymore?

chararry is a data structure like `ndarray` which stores data according to the premitive datatype. With the current implementation, we are not checking any data structure like list or ndarray but rather the primitive types of that ndarray such as int, bytes_, float, str_, bool.

RubelMozumder:
@GinzburgLev 

Can you remember which issue was handled in the modification  https://github.com/FAIRmat-NFDI/pynxtools/blame/nx_char_type/src/pynxtools/dataconverter/helpers.py#L729 ?
Currently, this code converts an integer to a sting and skips the validation warning.

Then I can also test my code for that issue as well.

GinzburgLev:
> @GinzburgLev
> 
> Can you remember which issue was handled in the modification https://github.com/FAIRmat-NFDI/pynxtools/blame/nx_char_type/src/pynxtools/dataconverter/helpers.py#L729 ? Currently, this code converts an integer to a sting and skips the validation warning.
> 
> Then I can also test my code for that issue as well.

This change was addressing the following issue: https://github.com/FAIRmat-NFDI/pynxtools/issues/393

The example from this issue should do: nexus file with one of the datasets with type NX_FLOAT, filled with integer = 0 (such as mpes example upload file, with sp.binned.attrs["metadata"]["energy_calibration"]["tof"] = 0 instead of 0.0). The is_valid_data_field() function in helpers.py successfully converted it to float(0.0), but this result was later interpreted as False (as in, failure to convert) by whatever was calling is_valid_data_field(). I can send you the files I used for testing (to big to be attached here).

RubelMozumder:
> > @GinzburgLev
> > Can you remember which issue was handled in the modification https://github.com/FAIRmat-NFDI/pynxtools/blame/nx_char_type/src/pynxtools/dataconverter/helpers.py#L729 ? Currently, this code converts an integer to a sting and skips the validation warning.
> > Then I can also test my code for that issue as well.
> 
> This change was addressing the following issue: #393
> 
> The example from this issue should do: nexus file with one of the datasets with type NX_FLOAT, filled with integer = 0 (such as mpes example upload file, with sp.binned.attrs["metadata"]["energy_calibration"]["tof"] = 0 instead of 0.0). The is_valid_data_field() function in helpers.py successfully converted it to float(0.0), but this result was later interpreted as False (as in, failure to convert) by whatever was calling is_valid_data_field(). I can send you the files I used for testing (to big to be attached here).

Thanks, I see it is not from your code. Can you help me to reproduce this error?

GinzburgLev:
> > > Then I can also test my code for that issue as well.
> > 
> > 
> > This change was addressing the following issue: #393
> > The example from this issue should do: nexus file with one of the datasets with type NX_FLOAT, filled with integer = 0 (such as mpes example upload file, with sp.binned.attrs["metadata"]["energy_calibration"]["tof"] = 0 instead of 0.0). The is_valid_data_field() function in helpers.py successfully converted it to float(0.0), but this result was later interpreted as False (as in, failure to convert) by whatever was calling is_valid_data_field(). I can send you the files I used for testing (to big to be attached here).
> 
> Thanks, I see it is not from your code. Can you help me to reproduce this error?

In the version of pynxtools that we had back then, I used dataconverter in the following way:

```
dataconverter --reader mpes --nxdl NXmpes raw_files/mZlhpEzOQY6Ms4hJbIIu3g/config_file.json raw_files/mZlhpEzOQY6Ms4hJbIIu3g/MoTe2_float.h5

dataconverter --reader mpes --nxdl NXmpes raw_files/mZlhpEzOQY6Ms4hJbIIu3g/config_file.json raw_files/mZlhpEzOQY6Ms4hJbIIu3g/MoTe2_int.h5
```

The first one worked fine, no errors; the second gave something like "Field /ENTRY[entry]/PROCESS_MPES[process]/energy_calibration/original_axis written without documentation." (aside from usual messages). The resulting file was missing corresponding field (or its value, I am not sure anymore) (output.nxs/entry/process/energy_calibration/original_axis)

RubelMozumder:
Currently, verification of NX data type does not convert the data to other data type anymore rather pop up an warning for any data type inconsistency. For example data `2.0` for `NX_INT` would not convert to `2` and vice-versa. In NeXus We have specific, data type `NX_INT` and `NX_FLOAT` in this case. If `2.0` and `2` is equivalent one can use `NX_NUMBER` in this case.

Such conversion creates issues in some cases such as for numbers `>1` to `true` and any string `"2"` to int/float 2/2.0 and silently validator pass them without any warning massages. Such issues is also observed in here #555

RubelMozumder:
- [ ] Nyaml needs to be fixed and merged before merging this PR, so that we can skip the circumvent the errors.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -591,71 +591,93 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
     np.uint64,
+    np.uint,
     np.unsignedinteger,
     np.signedinteger,
 )
 np_float = (np.float16, np.float32, np.float64, np.floating)
-np_bytes = (np.bytes_, np.byte, np.ubyte)
-np_char = (np.str_, np.char.chararray, *np_bytes)
+# Not to be confused with `np.byte` and `np.ubyte`, these store
+# and integer of `8bit` and `unsigned 8bit` respectively.
```
lukaspie, line 600:
```suggestion
# Not to be confused with `np.byte` and `np.ubyte`, these store
# integers of `8bit` and `unsigned 8bit` respectively.
```
What exactly is not to be confused? Maybe write "np.xxx is not to be confused..."

lukaspie (approved):
LGTM, thanks for the fix. Left some small comments.

Let's wait with the merge until we have the correct definitions here and the plugin tests pass.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -685,28 +707,31 @@ def convert_str_to_bool_safe(value):
...
                 if value is None:
                     raise ValueError
-            output_value = accepted_types[0](value)
-        except ValueError:
+                return True, value
+
+            collector.collect_and_log(
+                path, ValidationProblem.InvalidType, accepted_types, nxdl_type
+            )
+            return False, value
```
rettigl, line 733: This return value False here is interpreted as "undocumented field" in the calling function. So we get an additional wrong warning if the dtype does not match.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -685,28 +707,31 @@ def convert_str_to_bool_safe(value):
...
+    convert, it raises an Exception.
 
-    If it fails to convert, it raises an Exception.
-
-    Returns two values: first, boolean (True if the the value corresponds to nxdl_type,
-    False otherwise) and second, result of attempted conversion or the original value
-    (if conversion is not needed or impossible)
+    Returns two values:
+        boolean (True if the the value corresponds to nxdl_type, False otherwise)
+        converted_value bool value.
```
rettigl, line 717: Add typing annotation. The calling function expects a single bool.

src/pynxtools/dataconverter/validation.py
rettigl, line 1: Adjust all occurences of is_valid_data_field (e.g. line 552)

src/pynxtools/dataconverter/helpers.py
```diff
@@ -591,71 +591,93 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+            args = (obj, accepted_types) if accepted_types is not None else (obj,)
+            if not checker(*args):
+                return False
+        return True
+
+    # default checker
+    tmp_arr = None
+    if isinstance(objects, list):
+        # Handles list and list of list
+        tmp_arr = np.array(objects)
```
rettigl, line 666: Will this work of the dtypes within a list are not the same?

src/pynxtools/dataconverter/validation.py
rettigl: @RubelMozumder @sanbrock Returning False in L552 if the data type of a field does not fit effectively makes this field "undocumented". Is this intentional? The produced warnings at least are rather contradictive:

```
WARNING: The value at /ENTRY[entry]/data/@axes should be one of: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>), as defined in the NXDL as NX_CHAR.
WARNING: Field /ENTRY[entry]/data/@axes written without documentation.
```

rettigl: @sanbrock @sherjeelshabih Also, apparently during this second round of checking for undocumented fields, that are not being found before, no checks for enumerations are being done. I don't really understand the logic of this checking completely, but I suggest a checking for enums should happen here as well. 
Why can't we re-use the mechanism in recurse_tree here, i.e. the handling_map and corresponding functions?

src/pynxtools/dataconverter/helpers.py
RubelMozumder: I see that it is just for internal logic (please see line 750-766 in validation.py`. In `is_docmented` it checks the datatype, node type (is field or group). But the real type of the errors are collect and print according to the validation error type in and from `collect` function.

rettigl: I don't understand what you are trying to say here. My point is that is_documented returns False if the datatype does not match, which I would say is wrong, because it _is_ documented, but just has the wrong data type.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -685,28 +707,31 @@ def convert_str_to_bool_safe(value):
 def is_valid_data_field(value, nxdl_type, path):
     # todo: Check this funciton and wtire test for it. It seems the funciton is not
     # working as expected.
-    """Checks whether a given value is valid according to what is defined in the NXDL.
+    """Checks whether a given value is valid according to the type defined in the NXDL.
 
-    This function will also try to convert typical types, for example int to float,
-    and return the successful conversion.
+    This function also converts bool value comes in str format. In case, it fails to
+    convert, it raises an Exception.
```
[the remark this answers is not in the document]
rettigl: This behavior was btw. introduced in #522, I believe. Maybe @GinzburgLev or @lukaspie can comment?

src/pynxtools/dataconverter/helpers.py
rettigl: This behavior was btw. introduced in https://github.com/FAIRmat-NFDI/pynxtools/pull/522, I believe. Maybe @GinzburgLev or @lukaspie can comment?

src/pynxtools/dataconverter/helpers.py
rettigl: The return handling is correct, nevertheless I suggest to add typing everywhere.

sherjeelshabih:
Agreed. I already plan to merge what Rubel has here with some changes I want to keep in the other branch.

src/pynxtools/dataconverter/helpers.py
RubelMozumder: Doc is updated. You might be talking about the `return is_valid_data_field(mapping[key], node.dtype, key)[0]` in `validation.py. It returns a single first value. Extra variable is added.

src/pynxtools/dataconverter/helpers.py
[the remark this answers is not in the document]
RubelMozumder: I do not see the InvalidDateTime warning stored anywhere.

src/pynxtools/dataconverter/helpers.py
RubelMozumder: I think it will work. Because, numpy will check if all the elments have the same size of bit atleast converable to a common datatype, Otherwise it fails. If there is heterogeneous datatype (like string, nemeric) numpy  will choose unicode `U`/ python `O` or something like that.

RubelMozumder: Ahh, it will not work in specificatino of `NX_FLOAT` or `NX_INT`

src/pynxtools/dataconverter/helpers.py
```diff
@@ -591,71 +591,93 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
 }
 
 
-def check_all_children_for_callable(objects: list, check: Callable, *args) -> bool:
-    """Checks whether all objects in list are validated by given callable."""
-    for obj in objects:
-        if not check(obj, *args):
-            return False
+def check_all_children_for_callable(
+    objects: Union[list, np.ndarray],
```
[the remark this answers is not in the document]
RubelMozumder: All the datastructure are not officially mentioned in `NeXus` datatype. Only allowd datastructures are array or list which is equivalent to array. But here I added check for `Tuple`, I will not update the parameter annotation as we are not expecting that.

src/pynxtools/dataconverter/helpers.py
RubelMozumder: I wanted to explain the code a bit, to clarify that in that code, no different type validation errors are collected, as you are assuming `undocumented field` warning will pop up.
By the way, I removed the second value from return statement as it is not being used anywhere.

src/pynxtools/dataconverter/helpers.py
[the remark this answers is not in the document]
rettigl: I see you reject tuples now, this is also fine with me.

src/pynxtools/dataconverter/validation.py
```diff
@@ -534,8 +534,8 @@ def is_documented(key: str, node: NexusNode) -> bool:
             collector.collect_and_log(
                 f"{key}", ValidationProblem.MissingUnit, node.unit
             )
-
-        return is_valid_data_field(mapping[key], node.dtype, key)[0]
+        is_documented_flag = is_valid_data_field(mapping[key], node.dtype, key)
+        return is_documented_flag
```
rettigl, line 538: This would still tread a field with wrong datatype as undocumented. Is that really what we want?

RubelMozumder: Sorry, I am confused what is exact suggestion here?

rettigl: I will make a suggestions in another PR

RubelMozumder: I think you are getting two warnings (as you mentioned above), one is for `... without documentation` (which is wrong) and another one is related NX data type (expected). 
But I see different behavior in our test warning,
```
tests/dataconverter/test_helpers.py:714: AssertionError
----------------------------- Captured stderr call -----------------------------
WARNING: The value at /ENTRY[my_entry]/NXODD_name[nxodd_name]/int_value should be one of: (<class 'int'>, <class 'numpy.integer'>), as defined in the NXDL as NX_INT.
------------------------------ Captured log call -------------------------------
WARNING  pynxtools:helpers.py:98 WARNING: The value at /ENTRY[my_entry]/NXODD_name[nxodd_name]/int_value should be one of: (<class 'int'>, <class 'numpy.integer'>), as defined in the NXDL as NX_INT.
```
There is no `... without documentation warning`.

RubelMozumder: Though `@axes` is defined in `NXmpes_arpes.nxdl.xml` (https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/contributed_definitions/NXmpes_arpes.nxdl.xml#L344), but the validation check can not pick that attribute from app def. So, it raise the `without documentation warning`.
This is not an issue with `is_documented` function but might be an issue with algorithm how it resolves the nexus tree and node.

You can continue to comment in this PR or another PR. I just did a small research how the `undocumention` is coming only for a few quantities e.g. `axes` for others where we have false datatype.

src/pynxtools/dataconverter/helpers.py
rettigl: See #565

src/pynxtools/dataconverter/validation.py
rettigl: Added in #565

src/pynxtools/dataconverter/validation.py
rettigl: I will merge these to #565 to have a working version there

src/pynxtools/dataconverter/helpers.py
```diff
@@ -591,71 +591,93 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+    checker: Optional[Callable] = None,
```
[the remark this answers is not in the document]
rettigl: or test_function

lukaspie:
From the plugin tests, it does seem that the functionality is working as intended now. We get some easy-to-solve tests for some of the plugins, but we also get warnings like this (this is in `pynxtools-mpes`):
```
WARNING  pynxtools:helpers.py:98 WARNING: The value at /ENTRY[entry]/data/AXISNAME_indices[@energy_indices] should be one of the following Python types: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>, <class 'numpy.chararray'>), as defined in the NXDL as NX_CHAR.
```` 
This is an underlying problem with the definitions that we can only fix by using the latest definitions from the `fairmat` branch. However, this depends on #578 working, so that may need to be merged before this PR.

rettigl:
> This is an underlying problem with the definitions that we can only fix by using the latest definitions from the `fairmat` branch. However, this depends on #578 working, so that may need to be merged before this PR.

That's a fair point. An alternative to disentangle all these tasks a bit would be to add certain warnings to be expected temporarily, or live with some failing tests for a while, if we know that they should fail.

rettigl:
Tests fail because it checks out the wrong definitions:
grafik
grafik
@lukaspie any idea why?

rettigl:
Closed in favor of #585

Issue #556, opened 2025-02-18
Enum checking of attributes


rettigl:
Currently, attributes are not being checked against possible enumerations. This functionality should be added.

Steps to reproduce: 
Follow the conversion notebook here: https://nomad-lab.eu/prod/v1/gui/user/uploads/upload/id/iM3-bEttS060E2-GIqaSBA
The `@vector` attributes of the transformation classes should be specific strings according to NXmpes_arpes, and the numeric values passed don't match these strings (see also #555). Instead, these should be checked agains the enumeration, and a warning be raised potentially.

rettigl:
Possible fix in #557

Issue #555, opened 2025-02-18
Datatype checking for array dtype


rettigl:
Currently, the data type of numpy arrays is not being checked against the NX datatype, i.e. e.g. numeric arrays are treated as valid entries for NX_CHAR fields, or string arrays as valid entries for NX_NUMBER. Instead, the dtype of an array should be checked against the possible primitive datatypes.
Relevant code:
https://github.com/FAIRmat-NFDI/pynxtools/blob/d3b15e25df9eff3a7361e7fb155041ed8250523f/src/pynxtools/dataconverter/helpers.py#L611

RubelMozumder:
@rettigl I tried the *_specsscan_conversion_to_NeXus.ipynb with the current modification available here https://github.com/FAIRmat-NFDI/pynxtools/pull/554/files. With that modification I see the warning message from the mentioned notebook as follows:
```output
Using mpes reader to convert the given files:  
• NXmpes_arpes_config.json
• Scan1496.yaml
WARNING: The value at [/ENTRY](https://file+.vscode-resource.vscode-cdn.net/ENTRY)[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/TRANSFORMATIONS[transformations]/analyzer_rotation/@vector should be one of: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>), as defined in the NXDL as NX_CHAR.
WARNING: The value at [/ENTRY](https://file+.vscode-resource.vscode-cdn.net/ENTRY)[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/TRANSFORMATIONS[transformations]/analyzer_elevation/@vector should be one of: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>), as defined in the NXDL as NX_CHAR.
WARNING: The value at [/ENTRY](https://file+.vscode-resource.vscode-cdn.net/ENTRY)[entry]/INSTRUMENT[instrument]/ELECTRONANALYSER[electronanalyser]/TRANSFORMATIONS[transformations]/analyzer_dispersion/@vector should be one of: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>), as defined in the NXDL as NX_CHAR.
[...]

PR #557, opened 2025-02-18
add enum checking for attributes

rettigl:
Possible fix for #556


rettigl:
@sherjeelshabih please use this as a starting point.

rettigl:
@lukaspie Please check the warnings for NXxps, they are probably relevant, similar as for NXmpes_arpes.

lukaspie:
@rettigl thanks for the pointer. Indeed, the `@vector` attributes in `NXxps` are also wrongly assigned as `NX_CHAR`. I get warnings like `WARNING: The value at /ENTRY[Survey]/SAMPLE[sample]/transformations/sample_normal_tilt_azimuth_angle/@vector should be on of the following strings: ['[0, 0, -1]']`, whichmakes sense. 

I started making them `NX_NUMBER` in my refactoring branch: https://github.com/FAIRmat-NFDI/nexus_definitions/pull/329. I will test again if after updating the definitions in pynxtools, I also get errors for these attributes. If I understand #555 correctly, we should continue getting errors.

sherjeelshabih:
I added a small check for discussion. I now add list types (based on our convention) as enumeration objects in the validation tree. I tested with passing a list [0, 0, 1] and that works as expected now. If there are more examples, I can have a look.

rettigl:
> However, I was expecting that I still get a similar warning as before (see above) since the thee `@vector` attributes in `NXxps` are still wrongly defined as `NX_CHAR` and I write a list of ints like '[0, 0, -1]' into the file. The idea for such lists would be that each individual item in the list is checked against the datatypes (which should lead to an error in my case) AND that the whole list is checked against the complete list in the enum. Is that done here already or is this planned in a separate PR? Does this come with #554?

I checked the combination of this PR and #554. This seems to work more or less. The vectors are correctly checked for the enums, and a warning is given if the dtype is wrong (ints passed for NX_CHAR). 

However, the checking of list of strings seems to not work anymore now, as example the axes attribute in data. I pass a tuple with ("angular0", "angular1", "energy"), and it produces these errors:

```
WARNING: The value at /ENTRY[entry]/data/@axes should be one of: (<class 'str'>, <class 'numpy.str_'>, <class 'numpy.bytes_'>), as defined in the NXDL as NX_CHAR.
WARNING: Field /ENTRY[entry]/data/@axes written without documentation.
```

@RubelMozumder This might be due to that tuples are not considered in #554, and also maybe not here?

src/pynxtools/dataconverter/validation.py
```diff
@@ -476,6 +476,20 @@ def handle_attribute(node: NexusNode, keys: Mapping[str, Any], prev_path: str):
...
+                    f"{prev_path}/{variant if variant.startswith('@') else f'@{variant}'}"
+                ]
+                not in node.items
+            ):
+                collector.collect_and_log(
+                    f"{prev_path}/{variant if variant.startswith('@') else f'@{variant}'}",
+                    ValidationProblem.InvalidEnum,
+                    node.items,
+                )
+
```
rettigl, line 492: This does not quite work yet. If there are no variants, it does not check the enum rn.

rettigl: The same is actually apparently true for fields... (I copied the code from there).

rettigl: I added a fix for this, should work correctly now. Tests would still be very good to add.

sherjeelshabih: Thanks for the fix. I'll check it out. I'll add a test as well tomorrow.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -699,18 +656,14 @@ def is_valid_data_field(value, nxdl_type, path):
     accepted_types = NEXUS_TO_PYTHON_DATA_TYPES[nxdl_type]
     output_value = value
 
+    if isinstance(value, list):
+        return all(is_valid_data_field(v, nxdl_type, path) for v in value), output_value
+
```
rettigl, line 661: This produces one warning per element in a list, if the datatype is not correct. @RubelMozumder had a different solution for this in #554.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -578,66 +578,23 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
-    ),
-    "NX_BOOLEAN": (bool, np.ndarray, *np_bool),
-    "NX_CHAR": (str, np.ndarray, *np_char),
-    "NX_DATE_TIME": (str,),
-    "NX_FLOAT": (float, np.ndarray, *np_float),
-    "NX_INT": (int, np.ndarray, *np_int),
+    "ISO8601": (str),
+    "NX_BINARY": (bytes, bytearray, np.byte, np.ubyte, np.ndarray),
+    "NX_BOOLEAN": (bool, np.ndarray, np.bool_),
+    "NX_CHAR": (str, np.ndarray, np.chararray),
```
rettigl, line 586: This still allows to write numeric arrays to an NX_CHAR, which is not correct. This issue is also solved by #554

src/pynxtools/dataconverter/helpers.py
```diff
@@ -699,18 +656,14 @@ def is_valid_data_field(value, nxdl_type, path):
...
-            if accepted_types[0] is bool and isinstance(value, str):
-                value = convert_str_to_bool_safe(value)
-                if value is None:
-                    raise ValueError
-            output_value = accepted_types[0](value)
-        except ValueError:
-            collector.collect_and_log(
-                path, ValidationProblem.InvalidType, accepted_types, nxdl_type
-            )
-            return False, value
```
rettigl, line 713: I would keep the boolean string to bool conversion, this certainly makes sense and We could even think about doing this for all datatypes with safe literal conversion, as you did for the enums.

rettigl (commented):
There are a few issues with this after the last commit, see comments. For me, the combination of this branch before the last commit, and #554 seems to work fine.

rettigl:
@sanbrock This here adds lists as possible elements of enumerations for the dataconverter. However, read_nexus apparently still does not support this. In the debug log, I don't get any warning or so about a wrong enum:
```
DEBUG: ===== ATTRS (//entry/sample/transformations/sample_azimuth@vector)
DEBUG: value: [0 0 2] 
DEBUG: classpath: ['NXentry', 'NXsample', 'NXtransformations', 'NX_NUMBER']
DEBUG: classes:
NXmpes_arpes.nxdl.xml:/ENTRY/SAMPLE/transformations/sample_azimuth
NXtransformations.nxdl.xml:/AXISNAME
DEBUG: NXmpes_arpes.nxdl.xml:/ENTRY/SAMPLE/transformations/sample_azimuth@vector - [NX_NUMBER]
DEBUG: <<REQUIRED>>
DEBUG: enumeration (NXmpes_arpes.nxdl.xml:/ENTRY/SAMPLE/transformations/sample_azimuth/vector):
DEBUG: -> [0, 0, 1]
```

Nomad, on the other hand, reports a warning when parsing the file:
```

WARNING  nomad.processing     2025-02-26T16:09:43 error while setting attribute sample_tilt___vector in pynxtools.nomad.schema.Mpes_arpes__ENTRY__SAMPLE__transformations:Section as pynxtools.nomad.schema.Mpes_arpes__ENTRY__SAMPLE__transformations.sample_tilt___vector:Quantity
  - exception: Traceback (most recent call last):
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/pynxtools/src/pynxtools/nomad/parser.py", line 245, in _populate_data
        current.m_set(metainfo_def, attribute)
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 1251, in m_set
        return definition.__set__(self, value, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 720, in wrapper
        return method(self, obj, value, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 3319, in __set__
        m_quantity.value = self.type.normalize(
                           ^^^^^^^^^^^^^^^^^^^^
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/data_type.py", line 787, in normalize
        return self._normalize_impl(value, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/data_type.py", line 1072, in _normalize_impl
        return self._validate_value(value)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/femtolab_admin/nomad/nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/data_type.py", line 1057, in _validate_value
        raise ValueError(f'{value} is not a value of this enumeration.')
    ValueError: [1 0 0] is not a value of this enumeration.
  - exception_hash: NT_2ZeFxcIb-dn2KEP5pzWvm0WKG
  - nomad.commit:
  - nomad.deployment: oasis
  - nomad.entry_id: THOgExlK-anJLuKyUpwvjeM4JdlY
  - nomad.mainfile: Scan1496.nxs
  - nomad.processing.logger: nomad.processing
  - nomad.processing.parser: pynxtools.nomad.entrypoints:nexus_parser
  - nomad.processing.proc: Entry
  - nomad.processing.process: process_entry
  - nomad.processing.process_status: RUNNING
  - nomad.processing.process_worker_id: iBGbnDdVRq2UUr2BfJj26w
  - nomad.processing.step: pynxtools.nomad.entrypoints:nexus_parser
  - nomad.processing.target_name: vector
  - nomad.service: worker
  - nomad.upload_id: cpbp-fGeRVq7NfIbM2rMVA
  - nomad.version: 1.3.14rc2
  - taskName: None
```

Do you know where to fix this?

rettigl:
Closed in favor of #554

PR #558, opened 2025-02-21
accept open enumerations as no real enumerations

sanbrock:


src/pynxtools/nexus-version.txt
mkuehbach, line 1: f7ba is an even older commit than 6e31 might affect spm specifically

src/pynxtools/nomad/schema.py
```diff
@@ -469,17 +469,18 @@ class nexus definition.
...
 
-def __get_enumeration(xml_node: ET.Element) -> Optional[MEnum]:
+def __get_enumeration(xml_node: ET.Element) -> Tuple[Optional[MEnum], bool]:
     """
     Get the enumeration field from xml node
     """
     enumeration = xml_node.find("nx:enumeration", __XML_NAMESPACES)
     if enumeration is None:
-        return None
+        return None, None
```
mkuehbach, line 478: None is not bool why not false?

sanbrock: None is neither True, nor False.

mkuehbach: clear your return type annotation for the second one is bool though i.e. constrained would using that be cleaner,, agree does not change your logic in L550

sanbrock: done

mkuehbach (approved):
lgtm

PR #559, opened 2025-02-21
Example update

sanbrock:


lukaspie:
Sorry, didn't mean to improve, just wanted to ask the question above.

src/pynxtools/nomad/examples/iv_temp/IV_temp.archive.json
```diff
@@ -7,7 +7,8 @@
             "IV_temp.mapping.json",
             "IV_temp.pickle"
         ],
-        "output": "IV_temp.nxs"
+        "output": "IV_temp.nxs",
+        "nexus_view": "../upload/archive/mainfile/IV_temp.nxs#/data"
```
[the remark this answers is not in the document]
sanbrock: NOMAD reference to the entry of the given main file.

src/pynxtools/nomad/examples/iv_temp/eln_data.yaml
```diff
@@ -0,0 +1,2 @@
+author: "test"
+note: "Test note"
```
lukaspie, line 2: My suggestion would be not to ship the `eln_data.yaml` file with the examples. 

Rather, 
1) use `"pynxtools.nomad.dataconverter.ElnYamlConverter"` in `src/pynxtools/nomad/examples/iv_temp/IV_temp.schema.archive.yaml` 
2) ship an `.archive.json` that uses this schema to pre-fil the ELN and convert to `eln_data.yaml`
3) ship a _different_ `.archive.json` that use the `pynxtools.nomad.dataconverter.NexusDataConverter` to convert the example file + the `eln_data.yaml` file

The idea being that the ELN filling/conversion and NeXus file creation are subsequent actions. But, we can discuss how our examples should work by default.

Compare the structure of the [XPS](https://github.com/FAIRmat-NFDI/pynxtools-xps/tree/main/src/pynxtools_xps/nomad/examples) and [MPES](https://github.com/FAIRmat-NFDI/pynxtools-mpes/tree/main/src/pynxtools_mpes/nomad/examples) example for what I mean.

lukaspie: My point was exactly to use what you said in the bullet points, but to separate the steps of `eln_data.yaml` creation and NeXus file generation. In that case, we should not need the last point (because the separate steps clearly show how to use the `ElnYamlConverter`).

I am not sure we need a special example that includes the `eln_data.yaml`. The main advantage of NOMAD is that you have the GUI, so you don't need to work in the yaml file itself.

What I am thinking about is how to make it easier to create the ELN YAML schema in the first place. Currently, the user needs to create their own `scheme.archive.yaml` file depending on which elements they don't have in the native data file and thus need to supply through the ELN. We could think about a solution with the config file of the MultiFormatReader, which would create a NOMAD ELN schema based on all the concepts marked with `@eln`. So you auto-create the schema file based on the appdef and the config file.

lukaspie: Will be addressed in seperate PR.

src/pynxtools/nomad/utils.py
```diff
@@ -69,8 +69,8 @@ def __rename_nx_for_nomad(
         Optional[str]: The renamed NXDL name, with group names capitalized,
         or None if input is invalid.
     """
-    if name == "NXobject":
-        return name
+    # if name == "NXobject":
```
lukaspie, line 72: Shall we just remove this?

rettigl:
@sanbrock @lukaspie  Why has this been merged if it breaks test?

lukaspie:
> @sanbrock @lukaspie Why has this been merged if it breaks test?

This was intentional because these tests will be fixed by #578 (all but one are actually fixed there).

rettigl:
> > @sanbrock @lukaspie Why has this been merged if it breaks test?
> 
> This was intentional because these tests will be fixed by #578 (all but one are actually fixed there).

But why was this one then not first merged into #578? Now we have a broken master...

lukaspie:
That was my suggestion, but Sandor wanted to merge right away...

PR #565, opened 2025-02-28
Fix for checking of undocumented keys

rettigl:
-check enums for documented keys in second iteration
-Don't return False if issue with data type of documented keys


rettigl:
As far as I can see, the return value if is_valid_data_field is nowhere used. So this function just logs errors, which is fine. One could consider including the enum check is this function, actually, then one does not need to repeat this code.

rettigl:
This somewhat cleans up the type checking code and includes the enums there. Not sure how far all possible cases are covered in the tests, and these still need to be run (manually they pass).

sherjeelshabih (approved):

PR #573, opened 2025-03-04
Add more tests

rettigl:


src/pynxtools/data/NXtest.nxdl.xml
```diff
@@ -28,9 +28,13 @@
             </field>
         </group>
         <group type="NXdata" name="NXODD_name">
+            <field name="anamethatRENAMES" nametype="any" type="NX_INT" units="NX_UNITLESS"/>
```
lukaspie, line 31: That needs to be `nameType`

rettigl: Done, thanks.

PR #577, opened 2025-03-05
Fixes the types and removes bytes from NX_char as that creates failures

sherjeelshabih:


src/pynxtools/dataconverter/helpers.py
```diff
@@ -575,110 +575,52 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
-# an integer of `8bit` and `unsigned 8bit` respectively.
-np_bytes = (np.bytes_,)
-np_char = (
-    np.str_,
-    np.bytes_,
-    np.chararray,
-)  # Only numpy Unicode string and Byte string
-np_bool = (np.bool_,)
-np_complex = (np.complex64, np.complex128, np.cdouble, np.csingle, np.complex_)
+nx_char = (str, np.chararray)
```
rettigl, line 578: I don't think the np.chararray makes sense here. It is actually deprecated. Checking an array of strings with the new function agains str actually works.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -575,110 +575,52 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
+
 NEXUS_TO_PYTHON_DATA_TYPES = {
     "ISO8601": (str,),
-    "NX_BINARY": (
-        bytes,
-        *np_bytes,
-    ),
-    "NX_BOOLEAN": (bool, *np_bool),
-    "NX_CHAR": (str, *np_char),
+    "NX_BINARY": (bytes, bytearray, np.byte, np.ubyte),
```
rettigl, line 585: I am not sure if not actually np.bytes_ is the correct one here. np.byte or np.ubyte hold single byte ints, see Rubel's comment
```suggestion
    "NX_BINARY": (bytes, bytearray, np.bytes_),
```

src/pynxtools/dataconverter/helpers.py
rettigl: but there is np.character, see here: https://numpy.org/doc/2.1/reference/arrays.scalars.html
```suggestion
nx_char = (str, np.character)
```

src/pynxtools/dataconverter/helpers.py
```diff
@@ -575,110 +575,52 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
-        *np_float,
-    ),
+    "NX_FLOAT": nx_float,
+    "NX_INT": nx_int,
+    "NX_UINT": (np.unsignedinteger,),
+    "NX_NUMBER": nx_number,
+    "NX_POSINT": nx_int,  # > 0 is checked in is_valid_data_field()
+    "NX_COMPLEX": (complex, np.complexfloating,),
+    "NX_CHAR_OR_NUMBER": nx_char + nx_number,
+    "NXDL_TYPE_UNAVAILABLE": (str,),  # Defaults to a string if a type is not provided.
```
rettigl, line 596:
```suggestion
    "NXDL_TYPE_UNAVAILABLE": (nx_char,),  # Defaults to a string if a type is not provided.
```

src/pynxtools/dataconverter/helpers.py
```diff
@@ -575,110 +575,52 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
 }
 
 
 def check_all_children_for_callable(
-    objects: Union[list, np.ndarray],
-    checker: Optional[Callable] = None,
-    accepted_types: Optional[tuple] = None,
+    objects: Union[list, np.ndarray], checker: Optional[Callable] = None, *args
```
rettigl, line 601: let's rename checker to test_function or check_function

tests/dataconverter/test_helpers.py
```diff
@@ -512,22 +512,11 @@ def fixture_filled_test_data(template, tmp_path):
             alter_dict(
                 TEMPLATE,
                 "/ENTRY[my_entry]/NXODD_name[nxodd_name]/char_value",
-                np.array(["1", "2", "3"], dtype=np.bytes_),
+                np.char.array(["1", "2", "3"]),
```
rettigl, line 515: Elements here are actually str, this is not a chararray before they were in np.character. In any case, this will be parsed element-wise so we don't need chararray. I would keep the old version here.

tests/dataconverter/test_helpers.py
```diff
@@ -512,22 +512,11 @@ def fixture_filled_test_data(template, tmp_path):
...
-        #     alter_dict(
-        #         TEMPLATE,
-        #         "/ENTRY[my_entry]/NXODD_name[nxodd_name]/char_value",
-        #         np.char.chararray(["1", "2", "3"]),
-        #     ),
-        #     (""),
-        #     id="numpy-chararray",
-        # ),
-        # TODO add test array of char
-        # TODO add test for numpy array of char and chararray
```
rettigl, line 530: Or do both

rettigl (changes requested):
I would change a few things, see below

src/pynxtools/dataconverter/helpers.py
```diff
@@ -575,110 +575,57 @@ def is_value_valid_element_of_enum(value, elist) -> Tuple[bool, list]:
...
 
     def is_greater_than(num):
-        return num.flat[0] > 0 if isinstance(num, np.ndarray) else num > 0
-
-    if isinstance(value, list):
-        return check_all_children_for_callable(objects=value, checker=is_greater_than)
+        return num > 0
 
-    return value.flat[0] > 0 if isinstance(value, np.ndarray) else value > 0
+    return check_all_children_for_callable(objects=value, checker=is_greater_than)
```
rettigl, line 628:
```suggestion
    return check_all_children_for_callable(objects=value, check_function=is_greater_than)
```

PR #579, opened 2025-03-07
Do not log `read_nexus` to the console in tests

lukaspie:
Printing the whole log only unnecessarily fills up the test log without helping too much. To still see the line-by-line difference, I am now using `difflib`.


sanbrock (approved):
LGTM

PR #580, opened 2025-03-07
Update NeXus search app in pynxtools

GinzburgLev:
A generic app for search/statistics visualization for NXsensor_scan and inheriting classes (currently NXiv_temp, NXspm, NXstm, NXafm, NXsts); also will work for NOMAD CAMELS entries (most likely, version>=1.8.0).
Please let me know if:
- NXspm/sts/stm/afm etc should be removed from the filter, as they have a dedicated app with more suitable options
- NOMAD CAMELS should be mentioned more/less/not at all (currently only menu/protocols(NOMAD CAMELS and protocol column, which is not shown by default) have filters that work only for them)
- Any functionality / descriptions shall be added or improved
- Any way to replace explicit list of classes inheriting from NXsensor_scan with something more automated exists

For tests, files from corresponding example uploads will work. Test CAMELS files can be found here: https://box.hu-berlin.de/d/2812ff0b8104478ba160/   (just put any of them in an upload without any extra files/schemas)

### DIFF
--- a/src/pynxtools/nomad/entrypoints.py
+++ b/src/pynxtools/nomad/entrypoints.py
@@ -71,37 +71,29 @@ from nomad.config.models.ui import (
     Column,
     Menu,
     MenuItemHistogram,
-    MenuItemOption,
     MenuItemPeriodicTable,
     MenuItemTerms,
+    MenuSizeEnum,
     SearchQuantities,
 )
 
 schema = "pynxtools.nomad.schema.Root"
 
-nxdefs_inheriting_from_sensorscan = [
-    "NXsensor_scan",
-    "NXiv_temp",
-    "NXspm",
-    "NXsts",
-    "NXafm",
-    "NXstm",
-]
 
-nxsensor_scan_app = AppEntryPoint(
-    name="NXsensor_scan App",
-    description="App for NXsensor_scan and inheriting from it classes.",
+nexus_app = AppEntryPoint(
+    name="NeXus App",
+    description="Simple Generic NeXus app.",
     app=App(
         # Label of the App
-        label="Simple Scan",
+        label="NeXus",
         # Path used in the URL, must be unique
-        path="nxsensor_scan",
+        path="nexusapp",
         # Used to categorize apps in the explore menu
         category="Experiment",
         # Brief description used in the app menu
-        description="Search simple scan experiments (NXsensor_scan and inheriting classes), including output of NOMAD CAMELS",
+        description="A simple search app customized for generic NeXus data.",
         # Longer description that can also use markdown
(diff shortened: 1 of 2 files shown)


src/pynxtools/nomad/entrypoints.py
```diff
@@ -71,13 +71,255 @@ def load(self):
...
+        # package.
+        search_quantities=SearchQuantities(
+            include=[f"*#{schema}"],
+        ),
+        # Controls which columns are shown in the results table
+        columns=[
+            Column(title="Entry ID", search_quantity="entry_id", selected=True),
+            Column(
+                title="File Name",
+                search_quantity=f"data.name#{schema}",
```
sanbrock, line 117: which one of these are specific to NXsensor_scan and not inherited from NXentry

GinzburgLev: 1. Both things under menu->Protocols (this is actually CAMELS-specific), as well as corresponding column "Protocol" (not shown by default)
2. Sample ID (refers to data.ENTRY.SAMPLE.sample_id__field)
3. For some app defs inheriting from NXentry (such as NXem), autors are stored not under data.ENTRY.USER.name_field, but under data.ENTRY.userID.name__field. I can put the info from both into the same column, but it seems that I can not set up a filter for both simultaneously
 
Everything else is rather generic for NXentry

RubelMozumder:
You can extend the nexus_app instead of having another app in pynxtools. I think that is what you wanted to do.

GinzburgLev:
Updated layouts for smaller screen resolutions now. It still does not look very good at 1368x768, scale 100% (there I really can not do much, but also such screens are not that common for modern PCs). It is reasonable for everything larger, in my opinion. Attached are screenshots for 1920x1080, 1600x900 and 1368x768, all at 100% scale. For larger screens, I did not change anything from before.

Screenshot from 2025-03-12 10-25-33
Screenshot from 2025-03-12 10-26-07
Screenshot from 2025-03-12 10-26-55

Regarding AND / OR condition for multiple choice in Menu / Widget - a known issue of nomad, here is a relevant issue: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/2278

I do not think I can improve much more at the moment.

sanbrock (approved):
LGTM

GinzburgLev:
> @GinzburgLev So the histogram of nexus terms in the menu does work for you? Did you use some special branch of Nomad for this? Maybe I did something wrong in my tests...

I did not use anything special. Maybe this recent fix is relevant? https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2367

PR #584, opened 2025-03-13
Nexus identifierNAME implementation

RubelMozumder:


src/pynxtools/nomad/schema.py
```diff
@@ -137,6 +138,82 @@ class NexusActivityStep(ActivityStep):
...
+            upload_id=archive.metadata.upload_id, mainfile=f_name
+        ).first()
+        if not entry:
+            create_Entity(self.lab_id, archive, f_name, self.name)
+        self.reference = get_entry_reference(archive, f_name)
+        super().normalize(archive, logger)
+
+
+class NexusReferences(ArchiveSection):
+    NXidentifiers = SubSection(
```
lukaspie, line 186: I suggest to call this something else because it makes it seems like there is still a class called `NXidentifier`. Maybe we just call it `nexus_identifiers`?

src/pynxtools/nomad/schema.py
```diff
@@ -137,6 +138,82 @@ class NexusActivityStep(ActivityStep):
...
+        section_def=NexusIdentifier,
+        repeats=True,
+        description="These are the NOMAD references correspond to NeXus identifierNAME fields.",
+    )
+
+    def normalize(self, archive, logger):
+        # Consider multiple identifiers exists in the same group/section
+        identifiers = [
```
sanbrock, line 194: One could get all identifiers by simply getting `identifierNAME` which is a dict of them.

src/pynxtools/nomad/schema.py
```diff
@@ -137,6 +138,82 @@ class NexusActivityStep(ActivityStep):
...
+                continue
+            lab_id = (
+                re.split("([0-9a-zA-Z.]+)", val)[1]
+                + hashlib.md5(val.encode()).hexdigest()
+            )
+            identifier_path = f"{self.m_def.name}_{identifier.split('__field')[0]}"
+            logger.info(f"{lab_id} to be created")
+            logger.info(f"Identifier {val} refers to {lab_id}.")
+            nx_id = NexusIdentifier(lab_id=lab_id, name=identifier_path)
+            nx_id.normalize(archive, logger)
```
sanbrock, line 213: will be called automatically, nothing to do here.

src/pynxtools/nomad/schema.py
RubelMozumder: This refereing a group, `NXidentifiers` with `NX` to make that unique, because currently there are no groups or sections starting with `NX`. So no chance to collide with groups that may come with the same name. Note that `NXindentifiers` is comming through `NXobject`.

src/pynxtools/nomad/schema.py
RubelMozumder: I am not sure section obeject has a dict for variadic name like `identifierNAME__field`,
```debug
>> self.__dict__["identifierNAME__field"]
Traceback (most recent call last):
  File "<string>", line 1, in <module>
KeyError: 'identifierNAME__field'

>>getattr(self, "identifierNAME__field")
```debug
or not with 
```
>> self.__dict__.get("identifierNAME")

>> getattr(self, "identifierNAME")

```

File: src/pynxtools/nomad/schema.py (hunk 3 above)
Comment (RubelMozumder) [reply src/pynxtools/nomad/schema.py:213]: I also assumed nomad will automatically run nomalizer for NexusIdentifier, But seems, that is not the case.

File: src/pynxtools/nomad/schema.py [hunk 4]
```diff
@@ -137,6 +138,82 @@ class NexusActivityStep(ActivityStep):
...
+        def get_entry_reference(archive, f_name):
+            """Returns a reference to data from entry."""
+            from nomad.utils import hash
+
+            upload_id = archive.metadata.upload_id
+            entry_id = hash(upload_id, f_name)
+            return f"../uploads/{upload_id}/archive/{entry_id}#data"
+            # return f"/entries/{entry_id}/archive#/data"
+
+        f_name = f"{self.name}_{self.lab_id}.archive.json"
```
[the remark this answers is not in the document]
RubelMozumder: addresed!

src/pynxtools/nomad/schema.py
```diff
@@ -137,6 +138,82 @@ class NexusActivityStep(ActivityStep):
...
+            for key in self.__dict__.keys()
+            if key.startswith("identifier") and key.endswith("__field")
+        ]
+        if not identifiers:
+            return
+        self.NXidentifiers = []
+        for identifier in identifiers:
+            if not (val := getattr(self, identifier)):
+                continue
+            lab_id = (
```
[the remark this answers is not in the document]
RubelMozumder: Kept as found in nexus file.

src/pynxtools/nomad/schema.py
lukaspie: I get that it is a new concept, still I would call it something else because the `NX..` prefix is very confusing if you know NeXus

src/pynxtools/nomad/schema.py
```diff
@@ -1392,7 +1413,7 @@ def build_nomad_topology(archive):
     __rename_nx_for_nomad("NXfabrication"): normalize_fabrication,
     __rename_nx_for_nomad("NXsample"): normalize_sample,
     __rename_nx_for_nomad("NXsample_component"): normalize_sample_component,
-    __rename_nx_for_nomad("NXidentifier"): normalize_identifier,
+    # __rename_nx_for_nomad("NXidentifier"): normalize_identifier,
```
sanbrock, line 1416: this can be removed

src/pynxtools/nomad/schema.py
```diff
@@ -137,6 +138,74 @@ class NexusActivityStep(ActivityStep):
...
+
+            upload_id = archive.metadata.upload_id
+            entry_id = hash(upload_id, f_name)
+            return f"../uploads/{upload_id}/archive/{entry_id}#data"
+
+        super().normalize(archive, logger)
+        if not self.reference:
+            lab_hash = hashlib.md5(self.lab_id.encode()).hexdigest()
+
+            f_name = f"{self.name}_{lab_hash}.archive.json"
```
sanbrock, line 177: class_name could be used as prefix. Is `name` always valid for filename, too?

RubelMozumder: The `name` is constructed from class_name and field_name field to infer path of identifier. 
I think, `name` should be always valid if an identifierNAME is initialized, otherwise it would not initialize the NexusIdentifier.

src/pynxtools/nomad/schema.py
RubelMozumder: I modified the name `NXidentifiers` to `NexusIdenfiers` indicating that it is a list. I think this might work.

sanbrock (approved):
LGTM

PR #585, opened 2025-03-13
Fix validation and add more tests

rettigl:
This replaces #554 with properly keeping the git history. File contents are the same.


lukaspie:
For some reason, in the pytest workflow, https://github.com/FAIRmat-NFDI/pynxtools/commit/802fc31 is checked out. This is the commit that you would usually get _after_ you have merged this PR into master. I am quite confused, why does this commit exist at all already?

In addition, it checks out https://github.com/FAIRmat-NFDI/nexus_definitions/commit/3f66054d6b1651617fdfbb24d4b2bfa33f75de66. But for [branch here](https://github.com/FAIRmat-NFDI/pynxtools/tree/fix_validation_and_add_more_tests/src/pynxtools), the definitions are supposedly at https://github.com/FAIRmat-NFDI/nexus_definitions/tree/6e3134567d715198a03499d2ff1811389b2e6462, aka the old definitions we had for some time now.

I am not sure what's going on.

rettigl:
pynxtools tests pass now, but plugin tests have more severe problems. Need to investigate that further.

PR #586, opened 2025-03-13
fix short "*{}" notation

rettigl:
Fixes the *{} notation for values (it worked for keys before, but for values always used the first value).


lukaspie (approved):
lgtm

PR #588, opened 2025-03-17
Update nexus def and  add toposort package in pyproject.toml.

RubelMozumder:
This PR take care of

- [x] Missing `toposort` python package


lukaspie:
We cannot update the definitons just like that, this is why we have #578. Please revert the definitions update here. 

toposort seems to be missing from the toml indeed, so that's a good addition.

lukaspie (approved):
LGTM

PR #589, opened 2025-03-17
update CI/CD workflows

lukaspie:
NOMAD replaced `default_plugins.txt` by `test_plugins.txt` in https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2383 and removed `pynxtools[convert]` from there. Therefore, we needed to change the `nomad-requirements` workflow. 

Some more updates in the other workflows:
- Use `setup-uv` action everywhere
- Use python 3.11 consistently -> shall we use 3.12 at some point?

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -3,10 +3,7 @@ name: build_and_deploy_mkdocs
 on:
   push:
     branches:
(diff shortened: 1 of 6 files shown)


sanbrock (approved):
LGTM

PR #590, opened 2025-03-19
add bool to NX_CHAR_OR_NUMBER

rettigl:
Nexus states "Any valid character string or NeXus number representation", and thus I think it makes sense to also allow bool here.


lukaspie:
I am not sure this is correct, since neither `NX_CHAR` nor `NX_NUMBER` only booleans. Then why would their union support it?

rettigl:
My reasoning was that currently, there is no way in nexus that a field can accept any kind of data. I would expect that NX_CHAR_OR_NUMBER should be precisely that.

rettigl:
> I am not sure this is correct, since neither `NX_CHAR` nor `NX_NUMBER` only booleans. Then why would their union support it?

Technically, we could include booleans also in NX_INT, then they would be contained in NX_NUMBER

lukaspie:
> My reasoning was that currently, there is no way in nexus that a field can accept any kind of data. I would expect that NX_CHAR_OR_NUMBER should be precisely that.

https://github.com/nexusformat/definitions/pull/1246#issue-1622383164 introduced `NX_CHAR_OR_NUMBER` as the union of `NX_CHAR` and `NX_NUMBER`. IMO, this does not include `NX_BOOLEAN`. 

I see that it would be nice to be able to define a field can accept any kind of data. However, with the way the types are currently implemented, I don't think it is really possible. I think the best solution would actually be to be able to write unions of types, like type="NX_NUMBER|NX_CHAR" instead of inventing workarounds like `NX_CHAR_OR_NUMBER`. But this requires again a lengthy NIAC discussion.

The question is, what we do we do for now in pynxtools? Do we go with the literal interpretation of `NX_CHAR_OR_NUMBER` as it was defined or do we allow booleans as well to be more flexible? I think either one is fine, but other validators will of course make difference choices.

rettigl:
This is a bit inconsistent:
```
print(isinstance(value.item(), int))
print(isinstance(np.bool_, int))
print(isinstance(np.bool_, np.integer))

True
False
False
```

rettigl:
From https://joergdietrich.github.io/python-numpy-bool-types.html:

> The bool_ type is not a subclass of the int_ type (the bool_ is not even a number type). This is different than Python’s default implementation of bool as a sub-class of int.

lukaspie (approved):
LGTM, but linting needs to be fixed

PR #594, opened 2025-03-20
Fix slow varification of arrays

rettigl:
Split array dtype and sign verification into two functions


lukaspie (approved):
LGTM.

Thanks for the fix, this indeed speeds up the validation for XPS by quite a lot!

PR #596, opened 2025-03-22
Fix Coverage reports via coveralls

### LINKED ISSUE #595: Coverage reports don't work
The coverage reports currently do not seem to work properly. 
First, for PRs it sometimes creates errors (see e.g. https://github.com/FAIRmat-NFDI/pynxtools/actions/runs/13987769285).
Second, it appears to only check coverage in the test files, not the actual source code itself: 
Image

Third, coverage reports are issued for all python versions. This leads to inconsistencies depending on whether nomad is present or not, and corresponding changes in coverage reports. I suggest to report coverage only for one python version.

rettigl:
run coveralls only for pyton3.12, and use coveralls github action

Closes #595


lukaspie (approved):
LGTM

PR #598, opened 2025-03-24
Update for nexus app after pynxtools update

GinzburgLev:
Updated instrument name(s), sample_id and start_time histogram


lukaspie (approved):
LGTM

PR #607, opened 2025-03-31
Remodel the `eln_mapper` tool to use the NeXusTree

lukaspie:
I have started using the automatic ELN creation because I would like to have a parser in `pynxtools-xps` that automatically creates a matching ELN for a given input file. In that progess, it became obvious that the current `eln_mapper` tool can be improved upon. Therefore, I completely remodelled it.

- ELN creation based on the existing NexusTree structure used for the validation
- Both the reader and schema ELN generators are now classes that build on top of a common abstract `ElnGenerator` class.
- Two new options for the CLI tool:
  - `--optionality`: choice between writing those concepts to the ELN that are either "required", "recommended", "optional".
  - `--filter-file`:  Possibility to pass a list of filters (each of which is a template key) to the ELN generation. Only those concepts that are in the list will be written. For the CLI tool, we allow passing files that are used as config files in the MultiFormatReader to define such filters. Only those concepts that are supposed to come from the ELN file (annotated by the `@eln` prefix) are selected for the generated ELN file.

Can ideally be tested by running `generate_eln --nxdl NXxps --eln-type schema --optionality optional --filter-file src/pynxtools_xps/config/config_vms.json` from the root of the `pynxtools-xps` repo.


src/pynxtools/eln_mapper/schema_eln.py
```diff
@@ -0,0 +1,304 @@
...
+    },
+    "NX_INT": {
+        "convert_type": "int",
+        "component_name": "NumberEditQuantity",
+    },
+    "NX_NUMBER": {
+        "convert_type": "np.float64",
+        "component_name": "NumberEditQuantity",
+    },
+    "<NO FIELD TYPE>": {
```
sanbrock, line 91: UINT, POSINT, BINARY, CHAR_OR_NUMBER?

tests/data/eln_mapper/scan.scheme.archive.yaml
```diff
@@ -17,77 +17,282 @@ definitions:
           m_annotations:
             eln:
               component: StringEditQuantity
-              defaultDisplayUnit: <No Default unit>
-          description: ''
+          description: Extended title for entry
```
sanbrock, line 20: Why do not we add the link to the actual definition?

src/pynxtools/eln_mapper/schema_eln.py
lukaspie: Yes, these were not implemented so far. I have extended the mapping and now I am using ("str", "StringEditQuantity") as defaults (for NX_CHAR as default in NeXus).

tests/data/eln_mapper/eln.yaml
```diff
@@ -1,14 +1,17 @@
...
+      unit: null
+sample:
+  rotation_angle:
+    value: null
+    unit: null
+monitor:
+  data:
+    value: null
+    unit: null
+data: null
```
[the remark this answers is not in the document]
lukaspie: We weren't handling links so far, presumably these should never be filled from the ELN? I would keep it like that by default.  We can maybe add another flag that would write out all concepts, including links. Not sure.

tests/data/eln_mapper/scan.scheme.archive.yaml
lukaspie: These are the auto-generated docstrings (basically walking down the inheritance chain until a valid doc is found). We could also automatically add links here, but it may get too crowded? Will try though.

src/pynxtools/dataconverter/nexus_tree.py
```diff
@@ -985,10 +987,11 @@ def add_children_to(parent: NexusNode, xml_elem: ET._Element) -> None:
...
     # Set root attributes
-    nx_root, _ = get_nxdl_root_and_path("NXroot")
-    for root_attrib in nx_root.findall("nx:attribute", namespaces=namespaces):
-        child = tree.add_node_from(root_attrib)
-        child.optionality = "optional"
+    if set_root_attr:
+        nx_root, _ = get_nxdl_root_and_path("NXroot")
+        for root_attrib in nx_root.findall("nx:attribute", namespaces=namespaces):
+            child = tree.add_node_from(root_attrib)
+            child.optionality = "optional"
```
RubelMozumder, line 994: Not sure why we need root_attr in nexus group, even it is requested.

src/pynxtools/eln_mapper/eln.py
```diff
@@ -19,177 +17,199 @@
...
+                "are skipped, is this intentional?"
+            )
+
+    def _generate_output_file_name(self, output_file: str):
+        """
+        Generate the output file name of the schema ELN generator.
+
+        To be implemented by the different subclasses of ElnGenerator.
+        """
+        return ""
```
RubelMozumder, line 101: Is it for future implementation?

lukaspie: No, this whole class is an abstract base class, so this shall be an abstract method that a subclass must implement. I added a decorator to make it clearer.

src/pynxtools/dataconverter/nexus_tree.py
lukaspie: As far as I know, Florian decided to build the tree incl. all the root attributes so we can check for them in the validation. They shouldn't really need to be there, it is just a limitation of how the validation works.

tests/data/eln_mapper/scan.scheme.archive.yaml
lukaspie: added the links as discussed

RubelMozumder (approved):
LGTM!

PR #609, opened 2025-04-02
Export feature

sanbrock:


src/pynxtools/nomad/dataconverter.py
```diff
@@ -203,6 +203,12 @@ class NexusDataConverter(EntryData):
         a_browser=dict(adaptor="RawFileAdaptor"),
     )
 
+    filter = Quantity(
```
lukaspie, line 206: I don't understand how this filter option works? Is that a regex? Because `input_files` already supports e.g. wildcard notation like `*.txt*`.

lukaspie (commented):
Mostly I have questions about the `filter`option? Is that used in any of our examples yet?

src/pynxtools/nomad/dataconverter.py
```diff
@@ -236,6 +244,9 @@ def normalize(self, archive, logger):
             write_yaml(archive, eln_fname, eln_dict)
             input_file_list.append(eln_fname)
 
+        if not self.export:
+            return
+
```
RubelMozumder, line 249: We can also think of writing the nexus file but without creating the intermediate `eln.yaml` file by flag. In this case we need to nexus converter interface.

sanbrock: This in only about running or not the dataconverter and exporting the content to NeXus data exchange format (which also results in a new Nomad entry with the corresponding Nexus view in it).

src/pynxtools/nomad/dataconverter.py
sanbrock: Yes, filter is already in master, here it is only a reordering of the Quantities. The functionality is to enable selecting multiple files and running dataconverter on them one by one. This allows converting multiple vendor files from a single ELN.

src/pynxtools/nomad/dataconverter.py
sanbrock: Note that if the input files explicitly contain an eln_data.yaml file, no new eln_data.yaml file is generated. This is the built in functionality.

lukaspie: We can still think about just allowing the filled ELN dict from the archive.json as an input to the dataconverter at some point, without even creating the yaml file. I think that was Rubel's point? But can be done later.

lukaspie (approved):
LGTM

PR #610, opened 2025-04-03
remove name mangling in NOMAD integration

### LINKED ISSUE #606: Unconventional function naming
I noticed some function names are using **double** leading underscores which is a bit confusing, and I don't think this is [PEP-8 compliant](https://peps.python.org/pep-0008/#descriptive-naming-styles):
> In addition, the following special forms using leading or trailing underscores are recognized (these can generally be combined with any case convention):

> _single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g. :
tkinter.Toplevel(master, class_='ClassName')

> __double_leading_underscore: when naming a class attribute, invokes name mangling (inside class FooBar, __boo becomes _FooBar__boo; see below).

> __double_leading_and_trailing_underscore__: “magic” objects or attributes that live in user-controlled namespaces. E.g. __init__, __import__ or __file__. Never invent such names; only use them as documented.

Could be found with `def\s+__\w+[^_]\(`

I don't think the following is intended for name mangling:
https://github.com/FAIRmat-NFDI/pynxtools/blob/beedda5d018974185cf0e044acc05b0823e846af/src/pynxtools/nomad/parser.py#L372

And in `nomad.utils` and `nomad.schema`

lukaspie:
Fixes #606

### DIFF
--- a/src/pynxtools/__init__.py
+++ b/src/pynxtools/__init__.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+from typing import Dict
 import logging
 import os
 import re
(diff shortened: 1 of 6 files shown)


sanbrock (approved):
LGTM

PR #613, opened 2025-04-04
use orjson to store metainfo package

lukaspie:


lukaspie:
When loading the pynxtools metainfo package from the stored file `src/pynxtools/nomad/nxs_metainfo_package.json` (which is autogenerated from running the schema without re-loading), it looks like the metainfo is correctly instantiated.

However, in parsing, there is an error along the lines of `There is no subsection to hold a pynxtools.nomad.schema.<section-name>:Section in pynxtools.nomad.schema.<section-parent-name>:Section`. For the specific debugging I am using (see [`launch.json`](https://github.com/lukaspie/nomad-distro-dev/blob/main/.vscode/launch.json#L65), I get the error `There is no subsection to hold a pynxtools.nomad.schema.Arpes__ENTRY:Section in pynxtools.nomad.schema.Arpes:Section`. This is the first node in the NeXus files that's parsed. Interestingly, additional attributes that are in the NeXus file (in the NeXus base class `NXroot`) are properly parsed, so _something is_ working.

For the error above, it basically comes from `m_create` function call in [`src/pynxtools/nomad/parser.py`](https://github.com/FAIRmat-NFDI/pynxtools/blob/metainfo-pkg/src/pynxtools/nomad/parser.py#L112), specifically from [this line](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/nomad/metainfo/metainfo.py?ref_type=heads#L1737). Interestingly, in debug mode, `self.m_def.all_sub_sections_by_section` does contain the value of `section_def` (which is `pynxtools.nomad.schema.Arpes__ENTRY:Section`) as a key, see the screenshot.

So basically, my main questions are:
1) Why does the `get` call not resolve even though the section is in the `all_sub_sections_by_section`?
2) What's different about instantiating the schema from scratch vs. loading the package and initializing the metainfo afterwards? If I do the former, everything looks exactly the same in debug mode, but the `get` call keeps properly resolved.

image

@lauri-codes @blueraft @DanielYang59 any suggestions why this may be the case? cc @sanbrock

lukaspie:
Problem is that these two objects are not exactly the same. In debug mode,
```
id(list(self.m_def.all_sub_sections_by_section.keys())[-1])
>> 140248739911328
id(section_def)
>> 140248775424816
```

blueraft:
I can also reproduce this! cc: @TLCFEM since this is something related to the metainfo.

TLCFEM:
Try set `cached=False` in this line: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blame/develop/nomad/metainfo/metainfo.py?ref_type=heads#L4394 
Report back the result to proceed further.

lukaspie:
> Try set `cached=False` in this line: [gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blame/develop/nomad/metainfo/metainfo.py?ref_type=heads#L4394](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blame/develop/nomad/metainfo/metainfo.py?ref_type=heads#L4394) Report back the result to proceed further.

I still get the same issue, the ids do not match. I also tried `cached=False` for `all_sub_sections` since that is called in `all_sub_sections_by_section`, but the result was the same.

TLCFEM:
@lukaspie Please apply and check if everything works.
https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2430

TLCFEM:
Long story short, `m_from_dict` utilise the `SectionReference` to resolve the references section, this yields a `MProxy` that mimics the behaviour of a `Section`.
It contains everything the target `Section` contains but they are different objects in memory.

src/pynxtools/nomad/nxs_metainfo_package.json
TLCFEM, line 1: I do not think this file needs to be check-in and shipped with the package.

src/pynxtools/nomad/schema.py
```diff
@@ -1074,368 +1332,103 @@ def create_package_from_nxdl_directories() -> Package:
...
-def normalize_process(self, archive, logger):
-    """Normalizer for Process section."""
-    current_cls = section_definitions[_rename_nx_for_nomad("NXprocess")].section_cls
-    if self.date__field:
-        self.start_time = self.date__field
-    self.name = self.__dict__["nx_name"]
-    # one could also copy local ids to identifier for search purposes
-    super(current_cls, self).normalize(archive, logger)
+    local_dir = os.path.abspath(os.path.dirname(__file__))
+    nxs_filepath = os.path.join(local_dir, "nxs_metainfo_package.json")
```
TLCFEM, line 1348: Depending on where and who installed this package, and who executes this script, the write permission to this location may not always be possible.
Maybe choose some standard location such as `/var/cache/`.

src/pynxtools/nomad/nxs_metainfo_package.json
blueraft: This saves more than 10 seconds, and the XML definitions don't change often so I don't see why we can't ship this and avoid having to do the calculations. For most consumers who just install the package, it'll be an easy win.

The file will be compressed when the package is built so we don't need to do anything on that front either.

lukaspie: The disadvantage of not shipping it is that people installing the package in their oasis or just locally would have to rerun the package generation once. I guess this can be a problem for those people that want to e.g. install `pynxtools` in `nomad-distro-dev`, but don't actually develop it locally. I guess this is the use case that @blueraft is talking about. The same would of course happen for any NOMAD (OASIS) deployment, even though there an additional 10 seconds probably won't hurt too much. 

What's nice about having the json file checked into git is that allows us to have control that the XML definitions and the schema package match (see the new workflow). This is the one of the highest priority for me because I don't want to have several mismatching representations of the same concepts.

Another option would be that we don't check it in / ship it, but rather, locally, you need to generate it once and it is stored in e.g. `/var/cache`. We could maybe make the filename of the json file contain the git versions of pynxtools and the definitions, to keep the two local versions (definitions submodule + schema creation python file vs. the local package) in check. So if you update the definitions or `schema.py` locally, you always automatically regenerate the package when restarting NOMAD.

So, it sort of depends on the use case a bit. @blueraft, since you initiated this discussion, do you see the first use case as something that many people do? You mentioned that you have been telling lots of people to just deactivate pynxtools if it takes too long to start NOMAD, presumably you were talking exactly about just having `pynxtools` in `nomad-distro-dev` (as it is by default when forking the repo btw) without actually developing it?

lukaspie: Even if we check it in, we chould compress it and then decompress in the workflow so that the repository doesn't get blown up.

blueraft: >  just having pynxtools in nomad-distro-dev (as it is by default when forking the repo btw) without actually developing it?
Yes, for this case it'd be helpful to also not have to run this extra computation at the beginning.

The extra 10s for deployment is manageable but the bigger issue stems from the extra memory usage since each `gunicorn` worker ends up having to do the expensive calculation which spikes up the memory usage.

lukaspie: Okay, I see.

I think one concern is the repo size though, everytime we change the package we would add 1 MB of binary content (assuming it's zipped already), which would grow the repo over time. So eventually, we end up with 100s of MB slowing down the cloning, etc.

Maybe with uncompressed json it is different, those are not binary, so the diff that git stores would be smaller.

Another option is that we don't check it in the files, but we add it only for Github releases (and PyPI packages). So basically, in the `publish.yaml` pipeline that gets triggered on release, we add it dynamically as an asset to that release and then we pack it together with the python package. Maybe something like this:

```publish.yml
  build:
    name: Build Package
    needs: [compare-versions, test]
    runs-on: ubuntu-latest
    steps:
      [...]
      - run: |
          git reset --hard HEAD
          python -m build
      - name: Generate ZIP file
        run: |
          mkdir -p release
          # Here, we run the script to create the ZIP file of the package
          zip -r release/nxs-package.zip dist/
      - name: Upload ZIP file to GitHub Release
        uses: softprops/action-gh-release@v1
        with:
          files: release/nxs-package.zip
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      - name: Upload dist artifacts
        uses: actions/upload-artifact@v4
        with:
          name: dist-artifacts
          path: dist/
```

But honestly, I am not sure if that works though and whether it is a good idea in the first place.

lukaspie:
> With this PR, the next `pynxtools` release would be incompatible with earlier versions of `nomad` right? Since `pynxtools` doesn't declare `nomad` as a direct dependency (not sure why?), this might result in some broken installations if they're not using the latest `nomad` version.

I think the original idea was that pynxtools can be used standalone (for data conversion), so it was meant to be a lightweight install outside of NOMAD. I think after pluginization, we should have probably added `nomad` as explicit dependency though.

In any case, we can probably only publish after the next nomad release anyway, right?

rettigl:
> > With this PR, the next `pynxtools` release would be incompatible with earlier versions of `nomad` right? Since `pynxtools` doesn't declare `nomad` as a direct dependency (not sure why?), this might result in some broken installations if they're not using the latest `nomad` version.
> 
> I think the original idea was that pynxtools can be used standalone (for data conversion), so it was meant to be a lightweight install outside of NOMAD. I think after pluginization, we should have probably added `nomad` as explicit dependency though.
> 
> In any case, we can probably only publish after the next nomad release anyway, right?

I would argue we still should keep this independence. There should definitely be a standalone installation without nomad possible. We could create an optional dependency group, e.g. [nomad] that defines the required nomad version, though.

src/pynxtools/nomad/nxs_metainfo_package.json
rettigl: Could we run this during the build / wheel process, so it becomes part of the package, and is also being built if you install the package in editable mode?

lukaspie:
> > > With this PR, the next `pynxtools` release would be incompatible with earlier versions of `nomad` right? Since `pynxtools` doesn't declare `nomad` as a direct dependency (not sure why?), this might result in some broken installations if they're not using the latest `nomad` version.
> > 
> > 
> > I think the original idea was that pynxtools can be used standalone (for data conversion), so it was meant to be a lightweight install outside of NOMAD. I think after pluginization, we should have probably added `nomad` as explicit dependency though.
> > In any case, we can probably only publish after the next nomad release anyway, right?
> 
> I would argue we still should keep this independence. There should definitely be a standalone installation without nomad possible. We could create an optional dependency group, e.g. [nomad] that defines the required nomad version, though.

So, we could have a `[nomad]` extra in pynxtools pyproject.toml and then [in nomad-distro](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/pyproject.toml?ref_type=heads#L43), we would use `pynxtools[nomad]==<some-version>` to ensure compatibility?

If I remember correctly, there were problems when using optional dependencies (like we had before with `[convert]`) and then running using `uv pip compile` like [here](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/scripts/generate_python_dependencies.sh?ref_type=heads#L11), but maybe this is different.

blueraft:
> > > > With this PR, the next `pynxtools` release would be incompatible with earlier versions of `nomad` right? Since `pynxtools` doesn't declare `nomad` as a direct dependency (not sure why?), this might result in some broken installations if they're not using the latest `nomad` version.
> > > 
> > > 
> > > I think the original idea was that pynxtools can be used standalone (for data conversion), so it was meant to be a lightweight install outside of NOMAD. I think after pluginization, we should have probably added `nomad` as explicit dependency though.
> > > In any case, we can probably only publish after the next nomad release anyway, right?
> > 
> > 
> > I would argue we still should keep this independence. There should definitely be a standalone installation without nomad possible. We could create an optional dependency group, e.g. [nomad] that defines the required nomad version, though.
> 
> So, we could have a `[nomad]` extra in pynxtools pyproject.toml and then [in nomad-distro](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/pyproject.toml?ref_type=heads#L43), we would use `pynxtools[nomad]==<some-version>` to ensure compatibility?
> 
> If I remember correctly, there were problems when using optional dependencies (like we had before with `[convert]`) and then running using `uv pip compile` like [here](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/scripts/generate_python_dependencies.sh?ref_type=heads#L11), but maybe this is different.

In `pynxtools[nomad]`, `nomad-lab >= 1.3.16` should work though, having a fixed version there would instead complicate things.

rettigl:
> In `pynxtools[nomad]`, `nomad-lab >= 1.3.16` should work though, having a fixed version there would instead complicate things.

That's more or less what I mean, yes. Not sure about the uv pip compile, though. I guess we need to test it.

src/pynxtools/nomad/nxs_metainfo_package.json
blueraft: > Another option is that we don't check it in the files, but we add it only for Github releases (and PyPI packages).

In nomad-lab, we have a similar setup, where the docs and gui are only included in the `sdist`. In the gitlab pipeline, we build the gui, copy the files over to the correct folder and run `uv build`, so the final `sdist` has the files for the gui. I think a similar thing could be done here, in CI you create the json file and then run `uv build,` so the resulting package will have the json file but it's not part of the `git` repo.

lukaspie: I guess we could add something like this to our existing build wrappers (in `src/pynxtools/_build_wrapper.py`):
```python 
import orjson
from pynxtools.nomad.schema import create_metainfo_package

def _create_and_save_nexus_schema(target_dir):
    """Creates and saves the nexus schema JSON file in the target directory."""
    nxs_filename = "nxs_metainfo_package.json"
    nxs_filepath = os.path.join(target_dir, nxs_filename)

    if not os.path.exists(nxs_filepath):
        print(f"Creating {nxs_filename} during build process...")
        schema_dict = create_metainfo_package().m_to_dict()
        with open(nxs_filepath, "wb") as f:
            f.write(orjson.dumps(schema_dict, option=orjson.OPT_INDENT_2))
        print(f"{nxs_filename} created successfully.")
    else:
        print(f"{nxs_filename} already exists, skipping creation.")

def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
    """PEP 517 compliant build wheel hook."""
    _write_version_to_metadata()
    _write_definitions_remote_url()
    build_temp_dir = pathlib.Path(wheel_directory) / "build_temp"
    build_temp_dir.mkdir(parents=True, exist_ok=True)
    _create_and_save_nexus_schema(str(build_temp_dir))
    return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)

def build_sdist(sdist_directory, config_settings=None):
    """PEP 517 compliant build sdist hook."""
    _write_version_to_metadata()
    _write_definitions_remote_url()
    _create_and_save_nexus_schema(sdist_directory)
    return _orig.build_sdist(sdist_directory, config_settings)
```

This would presumably work for package releases. However, then the editable package installation takes 10 seconds longer, which is also not ideal.

blueraft: It's a one time thing right? Unless the package metadata changes, the installation step won't be run during local development

lukaspie:
> > > > > With this PR, the next `pynxtools` release would be incompatible with earlier versions of `nomad` right? Since `pynxtools` doesn't declare `nomad` as a direct dependency (not sure why?), this might result in some broken installations if they're not using the latest `nomad` version.
> > > > 
> > > > 
> > > > I think the original idea was that pynxtools can be used standalone (for data conversion), so it was meant to be a lightweight install outside of NOMAD. I think after pluginization, we should have probably added `nomad` as explicit dependency though.
> > > > In any case, we can probably only publish after the next nomad release anyway, right?
> > > 
> > > 
> > > I would argue we still should keep this independence. There should definitely be a standalone installation without nomad possible. We could create an optional dependency group, e.g. [nomad] that defines the required nomad version, though.
> > 
> > 
> > So, we could have a `[nomad]` extra in pynxtools pyproject.toml and then [in nomad-distro](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/pyproject.toml?ref_type=heads#L43), we would use `pynxtools[nomad]==<some-version>` to ensure compatibility?
> > If I remember correctly, there were problems when using optional dependencies (like we had before with `[convert]`) and then running using `uv pip compile` like [here](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-distro/-/blob/main/scripts/generate_python_dependencies.sh?ref_type=heads#L11), but maybe this is different.
> 
> In `pynxtools[nomad]`, `nomad-lab >= 1.3.16` should work though, having a fixed version there would instead complicate things.

Yeah sorry, this is what I meant, just a lower bound.

src/pynxtools/nomad/nxs_metainfo_package.json
lukaspie: This will also not work for editable installs because the paths will be wrong...

lukaspie:
@rettigl I have implemented it now that the json package file is created in the package before the build step is run. For editable installs, on the first run, we generate the package once and then afterwards, we can just reload it. The filename also contains the NeXus version, so if you change the definition, the schema package is automatically recreated. Are you happy with that? We also have `nomad` in the optional-dependencies now.

rettigl:
> @rettigl I have implemented it now that the json file is created before the `uv build` step is run, so it gets shipped with the package to PyPI and is available after install. For editable installs, on the first run, we generate the package once and then afterwards, we can just reload it. The filename also contains the NeXus version, so if you change the definition, the schema package is automatically recreated. Are you happy with that? We also have `nomad` in the optional-dependencies now.

This is totally fine. I only think it would be good to avoid doing every time you start up nomad, which already takes quite long for building the nexus metainfo.

.github/workflows/publish.yml
```diff
@@ -77,22 +75,54 @@ jobs:
...
+          uv pip install dist/*.tar.gz
+      - name: Check that package JSON file exists
+        run: |
+          SITE_PACKAGES=$(python -c "import site; print(site.getsitepackages()[0])")
+          FILE_COUNT=$(find "$SITE_PACKAGES/pynxtools/nomad" -name 'nxs_metainfo_package_*.json' | wc -l)
+          if [ "$FILE_COUNT" -eq 0 ]; then
+            echo "ERROR: nxs_metainfo_package_*.json not found in installed package."
+            exit 1
+          else
+            echo "Found $FILE_COUNT JSON file(s) in package:"
```
sanbrock, line 119: shouldn't we test if the found package(s) correspond(s) to the actual definitions submodule?

scripts/update_definitions.sh
```diff
@@ -0,0 +1,23 @@
+# Updates the definitions and generates the metainfo package.
+# Only works if you have pynxtools installed in editable mode.
+
+update_nexus_version() {
```
sanbrock, line 4: Cannot we either put the common functions to a common file being sourced, or make a generic script which behaves differently based on a parameter (update vs. reset)?

src/pynxtools/nomad/schema.py
```diff
@@ -1076,366 +1336,100 @@ def create_package_from_nxdl_directories() -> Package:
...
+                    visited_definitions.add(section)
+                    sections.append(section)
 
-    # plotly figure
-    fig = plot_3d_plotly(df_sampled)
-    # find figures hosting subsesction
-    figure_host = archive.data
-    # apend the figure to the figures list
-    figure_host.figures = [PlotlyFigure(figure=fig.to_plotly_json())]
+    for section in sections:
```
sanbrock, line 1387: would we gain additional time, if these attributes are also saved/loaded?

scripts/update_definitions.sh
lukaspie: changed to a single script with a parameter

src/pynxtools/nomad/schema.py
lukaspie: Thy are saved/loaded with this approach.

rettigl:
> I spent some time this week profiling the issue.
> 

The question is if one can still improve something on the Nomad side on why the creation of these entities takes so long.

>     * Currently, we add field statistics for all float quantities (mean, min, max, ...). Sometimes these are targetted in apps (e.g. https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/main/src/pynxtools_mpes/nomad/entrypoints.py#L337) -> these increase the number of quantities by **a lot.** Can we stop adding them blindly everywhere? Initial idea: only add them for the quantities in `data.ENTRY.DATA`. @rettigl you are currently the one that's using these statistics the most, would this approach be fine with you?

the idea is to find data sets that cover certain data ranges. As this mostly indeed makes sense for the axes in data, this would certainly be an option. This however sounds like a pretty "hacky" solution.

.github/workflows/publish.yml
mkuehbach: Should be tested if works as expected in production

src/pynxtools/nomad/schema.py
mkuehbach: Moving several of the def normalize_* code makes comparing the changes here unnecessary involved.
Copied these functions into a text comparison tool and there was able to confirm the these functions have not been modified but their text block just moved within the py file.

src/pynxtools/nomad/utils.py
```diff
@@ -128,3 +131,10 @@ def get_quantity_base_name(quantity_name):
         if quantity_name.endswith("__field") and quantity_name[-8] != "_"
         else quantity_name
     )
+
+
+def get_package_filepath():
+    local_dir = os.path.abspath(os.path.dirname(__file__))
+    filename = f"nxs_metainfo_package_{get_nexus_version()}.json"
```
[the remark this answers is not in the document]
mkuehbach: You follow a single source principle by calling get_package_filepath() so yeah probably fine.

PR #614, opened 2025-04-07
install NOMAD release and test before publishing

lukaspie:


.github/workflows/publish.yml
```diff
@@ -8,49 +8,104 @@ env:
...
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Install uv and set the python version
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: ${{ env.python-version }}
+
+      - name: Install dependencies
+        run: uv pip install build
```
RubelMozumder, line 29: Do we need that here? We are getting the versions here.

lukaspie: This is the setup for all following scripts. We can also do it later, probably is a very small difference.

lukaspie: I moved it to the `build` job.

RubelMozumder (approved):

PR #616, opened 2025-04-07
streamline build artifacts to upload job

lukaspie:


.github/workflows/publish.yml
```diff
@@ -3,6 +3,7 @@ name: Upload Python Package
 on:
   release:
     types: [published]
+  workflow_dispatch:
```
RubelMozumder, line 6: IS it intended!

RubelMozumder (commented):
I do not know it will work. Because, I think the error say there is not dist dir and probably build did not work properly.

lukaspie:
> I do not know it will work. Because, I think the error say there is not dist dir and probably build did not work properly.

This is because the built package is not available in the separate deploy step. You need to pass it along from the build to the deploy step using the artifacts upload.

RubelMozumder (approved):
Let's see how it goes.

RubelMozumder:
> > I do not know it will work. Because, I think the error say there is not dist dir and probably build did not work properly.
> 
> This is because the built package is not available in the separate deploy step. You need to pass it along from the build to the deploy step using the artifacts upload.

I see the output was not passed to the deploy step.

PR #620, opened 2025-04-09
Testing for generating template from NeXusTree

RubelMozumder:
In the current implementation how the Template instance is populated, not all the inheritance fields can be resolved specially some fields from inheritance chain are missing. Taking leverage from NeXus tree would be a good option for that.
 
Beside that, the template only stores the path of the nexus concept path as a plain string. And in the reader plugin, the reader developer somehow needs to check the proper type of the data, and most of the time, the information of the data type or unit type for a field or attribute is noted in reader plugin manually. Which is error prone and needs extra effort. This PR is intended to handle.

- [ ] Generate the template from NeXus tree
- [ ] Each template key should be an object that would carry other information such as `data type`, `unit type`, `nameType` info which will be carry exact information as an nxdl node contains. This method would justify `nxdl` as a single source of truth. 
- [ ] Add the features of template to curate the data fields and attributes according to the nxdl.
- [ ] Do not alter the current interface of the Template but implement the feature(2 & 3) as an extension of the current feature.

### DIFF
--- a/src/pynxtools/debug.py
+++ b/src/pynxtools/debug.py
@@ -1,0 +1,59 @@
+from pynxtools.dataconverter.helpers import generate_template_from_nxdl
+from pynxtools.dataconverter.nexus_tree import *
+from pynxtools.dataconverter.template import Template
+
+spm_appdef = "NXspm"
+spmT = generate_tree_from(spm_appdef)
+
+
+def add_template_key_from(nx_node,
+                       parent_path = ""):
+    key = None
+    unit = None
+    if nx_node.type == "attribute":
+        leaf_part = (f"{nx_node.name}[{nx_node.name}]"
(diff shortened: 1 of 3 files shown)


src/pynxtools/debug.py
```diff
@@ -0,0 +1,59 @@
...
+spm_appdef = "NXspm"
+spmT = generate_tree_from(spm_appdef)
+
+
+def add_template_key_from(nx_node,
+                       parent_path = ""):
+    key = None
+    unit = None
+    if nx_node.type == "attribute":
+        leaf_part = (f"{nx_node.name}[{nx_node.name}]"
```
RubelMozumder, line 14: Delete this file.

lukaspie:
@RubelMozumder can this be closed? The template is now automatically generated from the NeXus tree (as of #791).

Issue #599, opened 2025-03-25
Undescriptive error message


rettigl:
If one has "old style" identifiers in the dictionary for the converter, it fails with an undescriptive error message:

```
File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/convert.py:246, in convert(input_file, reader, nxdl, output, skip_verify, **kwargs)
    236 data = transfer_data_into_template(
    237     input_file=input_file,
    238     reader=reader,
   (...)
    242     **kwargs,
    243 )
    245 helpers.add_default_root_attributes(data=data, filename=os.path.basename(output))
--> 246 Writer(data=data, nxdl_f_path=nxdl_f_path, output_path=output).write()
    248 logger.info(f"The output file generated: {output}.")

File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/writer.py:340, in Writer.write(self)
    338 """Writes the NeXus file with previously validated data from the reader with NXDL attrs."""
    339 try:
--> 340     self._put_data_into_hdf5()
    341 finally:
    342     self.output_nexus.close()

File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/writer.py:296, in Writer._put_data_into_hdf5(self)
    294         print(str(exc))
    295     except Exception as exc:
--> 296         raise IOError(
    297             f"Unknown error occured writing the path: {path} "
    298             f"with the following message: {str(exc)}"
    299         ) from exc
    301 for links in hdf5_links_for_later:
    302     dataset = handle_dicts_entries(*links)

OSError: Unknown error occured writing the path: /ENTRY[entry]/entry_identifier/identifier with the following message: 'type'
```

PR #623, opened 2025-04-10
remove unused import.

RubelMozumder:


RubelMozumder:
@rettigl, you can have a quick look at this PR. The current implementation will raise an error

```
OSError: Unknown error occurred writing the path: /ENTRY[entry]/entry_identifier/identifier with the following message: NXDL attribute `type` not found for group /ENTRY[entry]/entry_identifier.
Hint: Follow the convention `fixednameVARIACKPART[fixedname_given_name]` with exact group name `fixednameVARIACKPART` in NXDL file.
```

src/pynxtools/dataconverter/writer.py
```diff
@@ -117,11 +116,13 @@ def handle_dicts_entries(data, grp, entry_name, output_path, path):
...
     - External links
-    - compression label"""
+    - compression label
+    """
+    file = None
     if "link" in data:
         file, path = split_link(data, output_path)
     # generate virtual datasets from slices
-    if "shape" in data.keys():
+    if file is not None and "shape" in data.keys():
```
rettigl, line 125: I don't understand, how is this related to the issue?

src/pynxtools/dataconverter/writer.py
```diff
@@ -246,9 +247,16 @@ def ensure_and_get_parent_node(self, path: str, undocumented_paths) -> h5py.Grou
...
-                grp.attrs["NX_class"] = attrs["type"]
+                try:
+                    grp.attrs["NX_class"] = attrs["type"]
+                except KeyError as e:
+                    raise NxdlAttributeNotFoundError(
+                        f"NXDL attribute `type` not found for group {parent_path}.\n"
+                        f"Hint: Follow the convention `fixednameVARIACKPART[fixedname_given_name]` "
+                        f"with exact group name `fixednameVARIACKPART`in NXDL file."
+                    ) from e
+
```
rettigl, line 259: I still don't find this error message helpful. Why is an error raised in the first place?
entry_identifier is now defined as a field in NXentry, while the writer wants to create a group. 
This should ideally already be marked as definition violation by the validation routine (I think #621 fixes this), but the writer should just write what it gets. If the user wants to write a group with a name of a defined field, the writer should still do it, I would say, and not fail. 
Your error message still does not help me solve the underlying problem as user.

src/pynxtools/dataconverter/writer.py
```diff
@@ -304,35 +309,31 @@ def add_units_key(dataset, path):
...
+                logger.warning(
+                    "No path '%s' available to attached Attribute.", hdf5_path
+                )
+                dataset_or_grp = None
+            if dataset_or_grp is None:
+                continue
+            if path.endswith("/@units"):
+                dataset_or_grp.attrs["units"] = d_value
+            else:
+                dataset_or_grp.attrs[entry_name[1:]] = d_value
```
rettigl, line 336: What issue do these changes solve? I don't understand it.

rettigl (changes requested):
This unfortunately still don't fix the problem, that the writer fails.
In addition, the provided error message does not help to solve the problem.

src/pynxtools/dataconverter/writer.py
RubelMozumder: That also came in my mind, writer will write the nexus file even the fields and groups are not defined at all.

If this is the case, we may write `GROUP_NOT_FOUND` something like that.

src/pynxtools/dataconverter/writer.py
RubelMozumder: It does not solve anything. This separates the concept handling of `group and field` from that of `attributes and units`.

RubelMozumder: I also mentioned that in this PR description.

src/pynxtools/dataconverter/writer.py
RubelMozumder: Removed!

src/pynxtools/dataconverter/writer.py
lukaspie: > entry_identifier is now defined as a field in NXentry, while the writer wants to create a group. This should ideally already be marked as definition violation by the validation routine (I think #621 fixes this), but the writer should just write what it gets.

Fully agree on this, the writer should not need to have this sort of information. In the validation, we really shouldn't allow groups to be called the same as another non-variadic field, this will also be very tricky to handle in NOMAD (see https://github.com/FAIRmat-NFDI/pynxtools/pull/437). This should be caught and at least a warning should be raised. The question is then if the conversion should fail with an error or if we just write an undocumented group instead. For such a group, we can't define an `NX_CLASS` attribute (since this new group doesn't fit to any defined class). You _can_ write valid NeXus files without the `NX_CLASS` attribute though (see https://github.com/nexusformat/definitions/issues/1493), so we should maybe just warn in the writer that this attribute doesn't exist, but continue writing the file.

RubelMozumder: Sure. I converted that error in warning massage. So, It will neither write the `type` attribute in nexus file nor stop writing the nexus file.

lukaspie: The message you are throwing still doesn't make any sense for what is the actual problem though.

RubelMozumder: What I understand, the actual problem is the `entry_identifier` (or any other group) does not have any group type, whether it is a group of variadic name type or an undefined group.

E.g. if someone writes 
`.../OBJECT[entry_identifier]/...` or `.../COLLECTION[entry_identifier]/...` which follows the rule of annotating a group of `NXobject` OR `NXcollection` type and correct convention.

The issue #599 is an edge case that uses an old format of `template` key without any group type. Although, all proper warnings come from the validation step.

I see this warning is an extra warning in any case if the `validator` is unable to detect it.
Or are there some extra points currently missing?

rettigl: The actual issue is that entry_identifier is defined here https://github.com/FAIRmat-NFDI/nexus_definitions/blob/297e81c8188434403750f7e1cb98fbbfec6c5d8f/base_classes/NXentry.nxdl.xml#L130 with a deprication attribute. Thus, attributes is defined, yet does not contain type, thus the error.
Properly, we should check for the existence of type in attribute.keys before accessing it. Then we also don't need an error or a try/catch block.

lukaspie: Oh, I see what you mean now. This will indeed be solved by #621

src/pynxtools/dataconverter/writer.py
RubelMozumder: @rettigl, here the `file` variable is created because logically these lines of code are not completed.

src/pynxtools/dataconverter/writer.py
RubelMozumder: Addressed!

RubelMozumder:
Encountering an issue related with @lur_cache at [get_inherited_node](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/dev_tools/utils/nxdl_utils.py#L909). 
This function returns some old stored values rather than recalculating it. Beause of this behavior, `get_inherited_node` does not return full list of the base class (skips NXresolution).

```Error
'WARNING: NXDL attribute `type` not found for group /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[wavelength_resolution].\nHint: Follow the convention `fixednameVARIACKPART[fixedname_given_name]` with exact group name `fixednameVARIACKPART`in NXDL file.'
```
The error is from raman reader. Though the group is available in [NXoptical_spcetroscopy](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXoptical_spectroscopy.html#nxoptical-spectroscopy) and the group was not been written in [raman nexus file](https://github.com/FAIRmat-NFDI/pynxtools-raman/blob/main/tests/data/rod/example.nxs). I have seen such misbehavior from `lru_cache` earlier.

@lukaspie, currently I have a few solutions in my mind.
1. To liberal `convert_nexus` caplog level and set it to `Error`
2. Remove the warning from the writer
3. Deactivate the `@lru_cache` in `dev_tools` in `nexus_definitions` (I will never go for that)

Both solutions 1 and 2 look more or less similar. But I would prefer solution 1, because tests are only for us to prevent unwanted changes from developers. But, warning is important for us and users equally. 

Let's discuss it tomorrow.

rettigl:
> Encountering an issue related with @lur_cache at [get_inherited_node](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/dev_tools/utils/nxdl_utils.py#L909). This function returns some old stored values rather than recalculating it. Beause of this behavior, `get_inherited_node` does not return full list of the base class (skips NXresolution).
> 
> ```
> 'WARNING: NXDL attribute `type` not found for group /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[wavelength_resolution].\nHint: Follow the convention `fixednameVARIACKPART[fixedname_given_name]` with exact group name `fixednameVARIACKPART`in NXDL file.'
> ```
> 
> The error is from raman reader. Though the group is available in [NXoptical_spcetroscopy](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXoptical_spectroscopy.html#nxoptical-spectroscopy) and the group was not been written in [raman nexus file](https://github.com/FAIRmat-NFDI/pynxtools-raman/blob/main/tests/data/rod/example.nxs). I have seen such misbehavior from `lru_cache` earlier.
> 
> @lukaspie, currently I have a few solutions in my mind.
> 
>     1. To liberal `convert_nexus` caplog level and set it to `Error`
> 
>     2. Remove the warning from the writer
> 
>     3. Deactivate the `@lru_cache` in `dev_tools` in `nexus_definitions` (I will never go for that)
> 
> 
> Both solutions 1 and 2 look more or less similar. But I would prefer solution 1, because tests are only for us to prevent unwanted changes from developers. But, warning is important for us and users equally.
> 
> Let's discuss it tomorrow.

I believe this is not related to the lru_cache. Even if I disable it, I get the same behavior. 
The issue arises, because NXresolution is not defined as a variadic name concept anywhere, but only the named concept wavelengh resolution is defined in NXoptical_spectroscopy. 
So (https://github.com/FAIRmat-NFDI/pynxtools/blob/599-undescriptive-error-message/src/pynxtools/dataconverter/writer.py#L222) returns /ENTRY/INSTRUMENT/RESOLUTION, which indeed is not defined in the tree.
This is quite related to #621 and will probably turn up in similar cases, where named concepts are written with concept definition in the template.
Changing RESOLUTION[wavelength_resolution] to wavelength_resolution in the raman rod reader fixes the conversion issue (validation fails then, though).
Note that the reference file does not contain a class definition for this group:
grafik
So the problem is indeed one we need to fix @lukaspie

RubelMozumder:
> > Encountering an issue related with @lur_cache at [get_inherited_node](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/dev_tools/utils/nxdl_utils.py#L909). This function returns some old stored values rather than recalculating it. Beause of this behavior, `get_inherited_node` does not return full list of the base class (skips NXresolution).
> > ```
> > 'WARNING: NXDL attribute `type` not found for group /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[wavelength_resolution].\nHint: Follow the convention `fixednameVARIACKPART[fixedname_given_name]` with exact group name `fixednameVARIACKPART`in NXDL file.'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The error is from raman reader. Though the group is available in [NXoptical_spcetroscopy](https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXoptical_spectroscopy.html#nxoptical-spectroscopy) and the group was not been written in [raman nexus file](https://github.com/FAIRmat-NFDI/pynxtools-raman/blob/main/tests/data/rod/example.nxs). I have seen such misbehavior from `lru_cache` earlier.
> > @lukaspie, currently I have a few solutions in my mind.
> > ```
> > 1. To liberal `convert_nexus` caplog level and set it to `Error`
> > 
> > 2. Remove the warning from the writer
> > 
> > 3. Deactivate the `@lru_cache` in `dev_tools` in `nexus_definitions` (I will never go for that)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Both solutions 1 and 2 look more or less similar. But I would prefer solution 1, because tests are only for us to prevent unwanted changes from developers. But, warning is important for us and users equally.
> > Let's discuss it tomorrow.
> 
> I believe this is not related to the lru_cache. Even if I disable it, I get the same behavior. The issue arises, because NXresolution is not defined as a variadic name concept anywhere, but only the named concept wavelengh_resolution is defined in NXoptical_spectroscopy. So (https://github.com/FAIRmat-NFDI/pynxtools/blob/599-undescriptive-error-message/src/pynxtools/dataconverter/writer.py#L222) returns /ENTRY/INSTRUMENT/RESOLUTION, which indeed is not defined in the tree. This is quite related to #621 and will probably turn up in similar cases, where named concepts are written with concept definition in the template. Changing RESOLUTION[wavelength_resolution] to wavelength_resolution in the raman rod reader fixes the conversion issue (validation fails then, though). Note that the reference file does not contain a class definition for this group: grafik So the problem is indeed one we need to fix @lukaspie

Yeah, this error is coming from group `RESOLUTION[wavelength_resolution]`. Unfortunately, while debugging, I logically found the issue is coming from `lru_cache`, though I did not check it. I fixed the test data for raman reader.

src/pynxtools/dataconverter/writer.py
lukaspie: Why do they need to be separated? What do we gain from this? If we have so many changes to the code, it would be good to understand the reasoning for why these changes are made.

src/pynxtools/dataconverter/writer.py
```diff
@@ -246,11 +247,19 @@ def ensure_and_get_parent_node(self, path: str, undocumented_paths) -> h5py.Grou
...
+                grp.attrs["NX_class"] = nx_class
+            else:
+                logger.warning(
+                    "NXDL attribute `type` not found for group %s.\n"
+                    "Hint: Follow the convention `fixednameVARIADICPART[fixedname_given_name]` "
+                    "or `non_variadic_group` where `fixednameVARIADICPART` is a variadic name "
+                    "and `non_variadic_group` is a fixed name of the group defined in NXDL file "
+                    "Or  no such group exists in NXDL file.",
+                    parent_path,
+                )
```
lukaspie, line 261: I still don't think that error message makes that much sense. The first case (i.e., not using `fixednameVARIADICPART[fixedname_given_name]` for a concept that has a renameable part) is now fixed by #621, so I don't think we will get to this point if you don't use it properly. Thus, that part of the error message can be removed.

The actual issue is if you try to attach a field for a group that doesn't exist, but for which the same name is used as a field. For example, using `"/ENTRY/entry_identifier/identifier"` as a key. `entry_identifier` is a **field** with XML attribute `deprecated` in `NXentry`, but with this notation we are using it as if it were a named group. Obviously, a field cannot have a sub-field itself, so that will lead to problems. We should instead raise an error that we are trying to write a group with a name that is reserved for a field instead.

This is something that should be caught by the validation instead of in the writer IMO.

lukaspie: We should not allow giving a group some name that is defined as the name of an field/attribute on the same level in general. This may be technically allowed in NeXus, but will break the whole NOMAD integration (since groups are (sub)sections and fields are quantities).

RubelMozumder: Yeah, we can remove the error massage :+1:, as this message mainly says the `invalid` group or `misdefined` group. This sort of wanrinig is part of the validation. Here, writer will the nexus file, if anything is wrong with the writing process that should be handeled here.

However, in one point I am a bit confused that is raising error wrting a group with the same name of `attribute` or `field` (which may come from mistyping, or just a regular error). In that case, `group` would not get proper `nx_class` name in such case NOMAD just will overlook that group in registering. Because, nomad can not match that group with any of the subsection in nomad metainfo.

rettigl: > We should not allow giving a group some name that is defined as the name of an field/attribute on the same level in general. This may be technically allowed in NeXus, but will break the whole NOMAD integration (since groups are (sub)sections and fields are quantities).

I agree, this creates the whole problem in the first place. We could consider to allow it to pass, but raise a strong and descriptive warning message.
With the current master, this still gets through validation with 
`WARNING: Field /ENTRY[entry]/entry_identifier/identifier written without documentation.`
And no hint to the conflicting concept, and then breaks in the writer. 

I would suggest to do two things:
- [ ] Add a check for conflicting concepts to the validation with clear and descriptive warning messages
- [ ] Catch any issue this might create in the writer (consider e.g. also a field of a name being defined, and the same name also as a group), also with clear and descriptive error messages

Any other changes to the code that don't relate to this specific problem I suggest to move to a different PR. It's very hard to review this way, because most changes you made don't relate to the problem as far as I can tell.

lukaspie: `WARNING: Given field name 'COLLECTION' conflicts with the non-variadic name 'entry_identifier (opt)'`

This is already correctly throwing a warning (should be an error though), but it should probably tell you that one is a field (defined and deprcated in `NXentry´) and the other is a new group.

RubelMozumder: > > We should not allow giving a group some name that is defined as the name of an field/attribute on the same level in general. This may be technically allowed in NeXus, but will break the whole NOMAD integration (since groups are (sub)sections and fields are quantities).
> 
> I agree, this creates the whole problem in the first place. We could consider to allow it to pass, but raise a strong and descriptive warning message. With the current master, this still gets through validation with `WARNING: Field /ENTRY[entry]/entry_identifier/identifier written without documentation.` And no hint to the conflicting concept, and then breaks in the writer.
> 
> I would suggest to do three things:
> 
> * [ ]  Add a check for conflicting concepts to the validation with clear and descriptive warning messages
> * [ ]  Catch any issue this might create in the writer (consider e.g. also a field of a name being defined, and the same name also as a group), also with clear and descriptive error messages
> * [ ]  Any other changes to the code that don't relate to this specific problem I suggest to move to a different PR. It's very hard to review this way, because most changes you made don't relate to the problem as far as I can tell.

Yeah, I can create two PRs referring to this PR.

I think both of the warnings are properly described the issue here.
The warning `WARNING: Field /ENTRY[entry]/entry_identifier/identifier written without documentation` says the correct warning. Indeed, there is not such group `entry_identifier` and even no such field.

lukaspie: We should not allow this in the first place if there is also a field called `entry_identifier` defined at the same level.

RubelMozumder:
Handled in PR #638.

PR #625, opened 2025-04-17
update definitions

lukaspie:
Needed to change a small part in the schema creation because `NXroot` doesn`t extend `NXobject` anymore.


rettigl:
@sanbrock @lukaspie I added now another special case for the schema generation for the Root schema, such that it also adds the NexusMeasurement base section. This fixes the issue with the search apps for me. Please have a look.

.github/workflows/plugin_test.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
             branch: main
             tests_to_run: tests/.    
           - plugin: pynxtools-mpes
-            branch: main
+            branch: update-defs
```
rettigl, line 36: Revert back bevore merging

lukaspie: All plugin tests are passing now, so I am reverting before we merge here (and then we should merge in the plugins afterwards).

Issue #575, opened 2025-03-05
Tests race condition


rettigl:
If tests are run in parallel, test_convert::test_compression and test_convert::test_links_and_virtual_fields have a race condition, as both copy the xarray h5 file, and restore it afterwards.

PR #626, opened 2025-04-23
Goup tests that use the same resources.

RubelMozumder:
**Problem:** Both tests `test_convert::test_compression and test_convert::test_links_and_virtual_fields` are writing the same file `xarray_saved_small_calibration.h5` from different tests. So, if the tests are run in parallel, two workers fall into a race condition.

This PR serializes the tests `test_convert::test_compression` and `test_convert::test_links_and_virtual_fields` using the grouping feature `xdist-group` from `pytest-xdist`.

In parallel testing, this feature allows scheduling tests over the workers but keeps the grouped test under the same worker, which resolves the classic race condition.

To use this feature, we need to tell pytest to run the grouped test under the same worker using `--dist loadgroup` flag, e.g. 

`$: python -m pytest -n auto --dist loadgroup`.

This resolves the issue [PR 575: Tests race condition](https://github.com/FAIRmat-NFDI/pynxtools/issues/575)  

- [ ] Update documentation on how to run tests in parallel conditions.

### DIFF
--- a/docs/how-tos/run-tests-in-parallel.md
+++ b/docs/how-tos/run-tests-in-parallel.md
@@ -1,5 +1,27 @@
-# Run `pynxtools` Tests in Parallel
-Pytest test framework also allows tests to run in parallel with some other third party pytest plugin, e.g. `pytest-xdist`. In our pytest for pynxtools we are using `[pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/)` to execute the test in prallel. In case of sheared resources among multiple test, the tests are being grouped by fixtures by `pytest.mark.xdist_group` to prevent any classic race coditions.
+# Running `pynxtools` Tests in Parallel
 
+The `pytest` framework allows tests to run in sequential and parallel using third-party plugins such as [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/). In our `pytest` setup for `pynxtools`, we use `pytest-xdist` to execute tests in parallel. To handle shared resources among multiple tests tests are grouped using the `@pytest.mark.xdist_group` fixture. This prevents classic race conditions by ensuring that tests sharing the same resources are executed serially.
 
+## Running Tests Sequentially
 
+In a local setup, tests can be run sequentially using the following command:
+
+```console
+$ python -m pytest tests
+```
+
+This will execute all tests in a sequential manner. For more details, refer to the official documentation: [How to invoke pytest](https://docs.pytest.org/en/stable/how-to/usage.html).
+
+## Running Tests in Parallel
+
+The `pytest-xdist` plugin can be used to speed up test execution by distributing tests among available workers. To prevent race conditions, tests that share the same resources are grouped using the `@pytest.mark.xdist_group(name="group_name")` fixture. These grouped tests must be run with the `--dist loadgroup` flag. For example:
+
+```console
+$ python -m pytest tests -n auto --dist loadgroup
+```
+
+Here:
+- The `-n auto` flag tells `pytest` to automatically distribute tests among all available workers.
+- The `--dist loadgroup` flag ensures that tests marked with the same @pytest.mark.xdist_group(name="...") are executed serially.
+
+This setup allows for efficient parallel test execution while maintaining the integrity of tests that depend on shared resources.
--- a/docs/how-tos/run-tests-in-parallel.md
+++ b/docs/how-tos/run-tests-in-parallel.md
@@ -1,6 +1,6 @@
 # Running `pynxtools` Tests in Parallel
 
-The `pytest` framework allows tests to run in sequential and parallel using third-party plugins such as [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/). In our `pytest` setup for `pynxtools`, we use `pytest-xdist` to execute tests in parallel. To handle shared resources among multiple tests tests are grouped using the `@pytest.mark.xdist_group` fixture. This prevents classic race conditions by ensuring that tests sharing the same resources are executed serially.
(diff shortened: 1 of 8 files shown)


lukaspie:
> To use this feature, we need to tell pytest to run the grouped test under the same worker using `--dist loadgroup` flag, e.g.
> 
> `$: python -m pytest -n auto --dist loadgroup`.

I don't see this in the CI/CD here. If it is really necessary, why don't we use it?

RubelMozumder:
> > To use this feature, we need to tell pytest to run the grouped test under the same worker using `--dist loadgroup` flag, e.g.
> > `$: python -m pytest -n auto --dist loadgroup`.
> 
> I don't see this in the CI/CD here. If it is really necessary, why don't we use it?


Locally, there are no complaints. Though in CI/CD jobs are already shared through the workers (4-workers are available). Using `-n auto --dist loadgroup` under the GH parallel processing hook, I see that there are many race conditions in the other tests which open and write files.~~only keeping `test_convert::test_compression and test_convert::test_links_and_virtual_fields` under the same worker, so no issues with that tests~~.

rettigl (changes requested):
This seems to work, but requires the addition of the --dist flag, which we should document in the readme, and add to the workflows. test_params_file() still gets in conflict, and needs the same flag.

rettigl:
> > > To use this feature, we need to tell pytest to run the grouped test under the same worker using `--dist loadgroup` flag, e.g.
> > > `$: python -m pytest -n auto --dist loadgroup`.
> > 
> > 
> > I don't see this in the CI/CD here. If it is really necessary, why don't we use it?
> 
> Locally, there are no complaints. Though in CI/CD jobs are already shared through the workers (4-workers are available). Using `-n auto --dist loadgroup` under the GH parallel processing hook, I see that there are many race conditions in the other tests which open and write files.~only keeping `test_convert::test_compression and test_convert::test_links_and_virtual_fields` under the same worker, so no issues with that tests~.

I dont quite understand what you mean here. If you parallelize jobs on the GH workers, you should get the same issues as locally, however maybe less likely if you have more/different kinds of cores locally.

RubelMozumder:
> > > > To use this feature, we need to tell pytest to run the grouped test under the same worker using `--dist loadgroup` flag, e.g.
> > > > `$: python -m pytest -n auto --dist loadgroup`.
> > > 
> > > 
> > > I don't see this in the CI/CD here. If it is really necessary, why don't we use it?
> > 
> > 
> > Locally, there are no complaints. Though in CI/CD jobs are already shared through the workers (4-workers are available). Using `-n auto --dist loadgroup` under the GH parallel processing hook, I see that there are many race conditions in the other tests which open and write files.~only keeping `test_convert::test_compression and test_convert::test_links_and_virtual_fields` under the same worker, so no issues with that tests~.
> 
> I dont quite understand what you mean here. If you parallelize jobs on the GH workers, you should get the same issues as locally, however maybe less likely if you have more/different kinds of cores locally.

Like you, I also expected the same. But this is not the case.
With the `pytest-xdist`, the `tests` or `groups of tests` will be distributed to the available workers in the host system. On the other side, in GH ci/cd the pytest workflow also has 5 `pytest jobs` (pytest with Python 3.8, 3.9, ..., 3.12) to run in parallel. I guess if we use parallelization for each `pytest job` (e.g. pytest with Python 3.8, 3.9, ..., 3.12). The tests from each job would be redistributed among the workers. This has created the classic race condition again.

I reached this conclusion, seeing how 5 `pytest jobs` were failing. Each of the 5 jobs failed because of the attempt to open a file synchronously (that means more than one worker trying to open the same file) in random tests associated with writing a file. But, I run each of the `pytest job` manually in a separate there were no issues.

RubelMozumder:
> This seems to work, but requires the addition of the --dist flag, which we should document in the readme, and add to the workflows. test_params_file() still gets in conflict, and needs the same flag.

@lukaspie and @rettigl 

Test `test_convert:test_params_file` was failing because `testdata.json` file is being used in several tests. The test `test_convert:test_params_file` is also included in the `shared_resouce` group. I think there should not be an issue to run the test in parallel using `-n auto --dist loadgroup` in ci/cd workflow and I added it.
I also added the docs referring flag `--dist loadgroup`.

docs/how-tos/run-tests-in-parallel.md
```diff
@@ -0,0 +1,27 @@
+# Running `pynxtools` Tests in Parallel
+
+The `pytest` framework allows tests to run in sequential and parallel using third-party plugins such as [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/). In our `pytest` setup for `pynxtools`, we use `pytest-xdist` to execute tests in parallel. To handle shared resources among multiple tests tests are grouped using the `@pytest.mark.xdist_group` fixture. This prevents classic race conditions by ensuring that tests sharing the same resources are executed serially.
```
rettigl, line 3:
```suggestion
The `pytest` framework allows tests to run in sequential and parallel using third-party plugins such as [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/). In our `pytest` setup for `pynxtools`, we use `pytest-xdist` to execute tests in parallel. To handle shared resources among multiple tests, tests are grouped using the `@pytest.mark.xdist_group` fixture. This prevents classic race conditions by ensuring that tests sharing the same resources are executed sequentially.
```

rettigl (approved):
LGTM now

rettigl:
test_nexus apparently is affected as well.

RubelMozumder:
> test_nexus apparently is affected as well.

Addressed!

rettigl:
@RubelMozumder Merging this destroyed coverage reports. Either revert, or fix.

tests/nexus/test_nexus.py
```diff
@@ -19,7 +19,6 @@
 
 import logging
 import os
-import difflib
```
rettigl, line 22: Why did you remove this line? It breaks the test for me...

RubelMozumder: Sorry, that was not intentionally removed, but I see that somehow erased. 
Thanks @lukaspie.

RubelMozumder:
> @RubelMozumder Merging this destroyed coverage reports. Either revert, or fix.

The overall result is still there only for `3.12`, and it is not expected for `python<3.11` or later according to githab pytest workflow.
https://github.com/FAIRmat-NFDI/pynxtools/actions/runs/14660792748/job/41144610789

PR #627, opened 2025-04-23
update definitions, allow `target` and `reference` attributes

lukaspie:
Intentionally only implemented a very shallow check for the two attributes because this is likely to change after one of the next Telcos.

Note that we still have `@reference` in our current version of NXdata, so this is anyway allowed at the moment. But, this is also part of the unresolved discussions.


rettigl (approved):
Fine with me as a temporary solution

PR #633, opened 2025-04-29
add datetime as minimum of start times from all entries

GinzburgLev:
Making "Start Time" column in NeXus app sortable by assigning min(start_times) from all entries to data.datetime


sanbrock (approved):
LGTM

PR #635, opened 2025-05-07
update definitions, remove `uv` from dev dependencies

lukaspie:
`uv` should be installed independent of the package and not as a package dependency.


.github/workflows/plugin_test.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-ellips
-            branch: main
+            branch: update-example-and-ci
```
lukaspie, line 27: reminder to revert

pyproject.toml
```diff
@@ -63,7 +63,6 @@ dev = [
     "types-pyyaml",
     "types-pytz",
     "types-requests",
-    "uv",
```
RubelMozumder, line 66: What is the purpose of removing the `uv` from dev?

lukaspie: One should install `uv` independently from the package and not as a package dependency. For other plugins that depend on `uv` as a dev dependency, there were issues that people suddenly had multiple versions of uv installed.

RubelMozumder: There are two things:
1. In pynxtools `uv` does not have any specific version, package manager can collect the pinned one wherever `uv` is pinned.
2. Plugin developer should not install `pynxtools` in dev mode, if still they install `pynxtools` in dev mode they should follow how `dev` dependency from `pynxtools`. Definitely `pynxtools` should have precedence upon its' plugins.

lukaspie: This is not about pynxtools plugins, but I was talking about other NOMAD plugins.

The point is that for nomad development, it is encouraged to install uv as standalone, independent from the packages you are working on. It has also been remove from the nomad-plugin-template, we are just updating to that here.

RubelMozumder (approved):
Once the plugin `pynxtools-xps` passes the ci/cd, you can merge. I  approve it.

.github/workflows/plugin_test.yaml
lukaspie: changed

PR #637, opened 2025-05-09
convert int-like values silently into float-like for NX_FLOAT

### LINKED ISSUE #523: [Bug]: Dataconverter does not change type of a field with the type different but compatible to the expected one
When the field of a certain type is passed a value of a different, but compatible type, the correct verification is done, however the resulting .nxs file still has the original (non-converted) type.

### Example

Using the pynxtools-mpes example upload files, add a field at "metadata/energy_calibration/tof" path to MoTe2.h5. It is supposed to be NX_FLOAT, add instead integer there. The output of dataconverter (`dataconverter --reader mpes --nxdl NXmpes path/config_file.json path/MoTe2_with_int.h5`) will still have integer saved under the field. Same procedure with string under the field will (correctly) give a warning.

### Likely reason

Function `is_valid_data_field()` ("pynxtools/src/pynxtools/dataconverter/helpers.py") checks the possibility of conversion and the conversion itself, however it is later used only in validation, and not for the conversion itself.

rettigl:
As discussed with @lukaspie, I believe we should not be strict with NX_FLOAT and convert int-like values into float-like values silently. I don't see any case where this would cause an issue, but if NX_FLOAT values are encoded as ints in some source file, it can be difficult to change this in the writing process. One example is the example E1 from pynxtools_mpes https://github.com/FAIRmat-NFDI/pynxtools-mpes/blob/60aa3d5724f7a1578287cb1331008a47c95c91ae/src/pynxtools_mpes/nomad/examples/E1%20Convert%20to%20NeXus.ipynb where one NX_FLOAT value is encoded as int in the hdf5 source file.

This PR adds a conversion function and respective tests.

Closes #523

### DIFF
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -672,6 +672,10 @@ def convert_int_to_float(value):
         return float(value)
     elif isinstance(value, list):
         return [convert_int_to_float(v) for v in value]
+    elif isinstance(value, tuple):
+        return tuple(convert_int_to_float(v) for v in value)
+    elif isinstance(value, set):
+        return {convert_int_to_float(v) for v in value}
     elif isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.integer):
         return value.astype(float)
     else:
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -658,6 +658,26 @@ def convert_str_to_bool_safe(value: str) -> Optional[bool]:
     raise ValueError(f"Could not interpret string '{value}' as boolean.")
 
 
+def convert_int_to_float(value):
+    """
+    Converts int-like values to float, including values in arrays, and lists
+
+    Args:
+        value: The input value, which can be a single value, list, or numpy array.
+
+    Returns:
+        The input value with all int-like values converted to float.
+    """
+    if isinstance(value, int):
+        return float(value)
+    elif isinstance(value, list):
+        return [convert_int_to_float(v) for v in value]
+    elif isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.integer):
+        return value.astype(float)
+    else:
+        return value
+
+
 def is_valid_data_field(
     value: Any, nxdl_type: str, nxdl_enum: list, nxdl_enum_open: bool, path: str
 ) -> Any:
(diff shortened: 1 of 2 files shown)


src/pynxtools/dataconverter/helpers.py
```diff
@@ -658,6 +658,26 @@ def convert_str_to_bool_safe(value: str) -> Optional[bool]:
...
+    Args:
+        value: The input value, which can be a single value, list, or numpy array.
+
+    Returns:
+        The input value with all int-like values converted to float.
+    """
+    if isinstance(value, int):
+        return float(value)
+    elif isinstance(value, list):
+        return [convert_int_to_float(v) for v in value]
```
lukaspie, line 674: Not sure if this ever comes up, but just for the sake of being comprehensive, we can add support for sets and tuples.
```suggestion
    elif isinstance(value, list):
        return [convert_int_to_float(v) for v in value]
    elif isinstance(value, tuple):
        return tuple(convert_int_to_float(v) for v in value)
    elif isinstance(value, set):
        return {convert_int_to_float(v) for v in value}
```

src/pynxtools/dataconverter/helpers.py
```diff
@@ -683,6 +703,12 @@ def is_valid_data_field(
                 collector.collect_and_log(
                     path, ValidationProblem.InvalidType, accepted_types, nxdl_type
                 )
+        elif accepted_types[0] is float:
```
lukaspie, line 706: That also converts for the validation, right? Not in the actual template that is passed to the writer.

src/pynxtools/dataconverter/helpers.py
rettigl: That's a good idea, I will add this

src/pynxtools/dataconverter/helpers.py
rettigl: It does change the value in the dict (through the return value), and thus also in the hdf file, similarly to the bools. I checked that it works, but have no explicit unit test for this.

lukaspie (approved):
LGTM

src/pynxtools/dataconverter/helpers.py
rettigl: The (possibly modified) value is returned here: https://github.com/FAIRmat-NFDI/pynxtools/blob/215779d426b3caffc46c69a7ce8a910822fa714b/src/pynxtools/dataconverter/helpers.py#L748
And replaces the old one here: https://github.com/FAIRmat-NFDI/pynxtools/blob/215779d426b3caffc46c69a7ce8a910822fa714b/src/pynxtools/dataconverter/validation.py#L510
https://github.com/FAIRmat-NFDI/pynxtools/blob/215779d426b3caffc46c69a7ce8a910822fa714b/src/pynxtools/dataconverter/validation.py#L552
https://github.com/FAIRmat-NFDI/pynxtools/blob/215779d426b3caffc46c69a7ce8a910822fa714b/src/pynxtools/dataconverter/validation.py#L643

lukaspie: I see, that makes sense

PR #638, opened 2025-05-14
add a separate error for names with wrong type and remove such keys

lukaspie:
The idea is to catch the following situation: a field/group is given that has a name that another group/field at the same level has (and which is therefore reserved).

Approach:
1) Throw and log an error on such cases
2) this error should be separate to the existing checks for concept names
3) if this error is thrown, it should result in that field not to be written at all (i.e., it shall be removed from the template)

### DIFF
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -523,7 +523,7 @@ def validate_dict_against(
 
         resolved_keys = copy.deepcopy(keys)
         for key, value in keys.copy().items():
-            if isinstance(value, dict) and len(value) == 1 and "link" in value:
+            if isinstance(value, dict) and "link" in value:
                 key_path = f"{prev_path}/{key}" if prev_path else key
                 current_keys = nested_keys
                 link_key = None
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -629,7 +629,7 @@ def validate_dict_against(
 
         key_len = len(key_components)
 
-        expected_types = None
+        expected_types = []
         for ind, name in enumerate(key_components):
             index = ind + 1
             children_to_check = [
@@ -649,7 +649,7 @@ def validate_dict_against(
 
         return node
 
-    def is_documented(key: str, tree: NexusNode) -> Tuple[bool, bool]:
+    def is_documented(key: str, tree: NexusNode) -> bool:
         if mapping.get(key) is None:
             # This value is not really set. Skip checking its documentation.
             return True
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -145,7 +145,12 @@ def split_class_and_name_of(name: str) -> Tuple[Optional[str], str]:
     ), f"{name_match.group(2)}{'' if prefix is None else prefix}"
 
 
-def best_namefit_of(name: str, nodes: Iterable[NexusNode]) -> Optional[NexusNode]:
+def best_namefit_of(
+    name: str,
+    nodes: Iterable[NexusNode],
+    expected_types: List[str],
+    check_types: bool = False,
+) -> Optional[NexusNode]:
     """
     Get the best namefit of `name` in `keys`.
 
@@ -166,6 +171,19 @@ def best_namefit_of(name: str, nodes: Iterable[NexusNode]) -> Optional[NexusNode
     for node in nodes:
         if not node.variadic:
             if instance_name == node.name:
+                if node.type not in expected_types and check_types:
+                    expected_types_str = " or ".join(expected_types)
+                    collector.collect_and_log(
+                        name,
+                        ValidationProblem.InvalidNexusTypeForNamedConcept,
+                        node,
+                        expected_types_str,
+                    )
+                    raise TypeError(
+                        f"The type ('{expected_types_str if expected_types else '<unknown>'}') "
+                        f"of the given concept {name} conflicts with another existing concept {node.name} (which is of "
+                        f"type '{node.type}')."
+                    )
                 if concept_name and concept_name != node.name:
                     inherited_names = [
                         name
@@ -176,6 +194,7 @@ def best_namefit_of(name: str, nodes: Iterable[NexusNode]) -> Optional[NexusNode
                         or (type_attr := elem.attrib.get("type"))
                         and len(type_attr) > 2
                     ]
+
                     if concept_name not in inherited_names:
                         if node.type == "group":
                             if concept_name != node.nx_class[2:].upper():
@@ -600,25 +619,52 @@ def validate_dict_against(
         # TODO: Raise error or log the issue?
         pass
 
-    def add_best_matches_for(key: str, node: NexusNode) -> Optional[NexusNode]:
-        for name in key[1:].replace("@", "").split("/"):
+    def add_best_matches_for(
+        key: str, node: NexusNode, check_types: bool = False
+    ) -> Optional[NexusNode]:
+        key_components = key[1:].split("/")
+        is_last_attr = key_components[-1].startswith("@")
+        if is_last_attr:
+            key_components[-1] = key_components[-1].replace("@", "")
+
+        key_len = len(key_components)
+
+        expected_types = None
+        for ind, name in enumerate(key_components):
+            index = ind + 1
             children_to_check = [
                 node.search_add_child_for(child)
                 for child in node.get_all_direct_children_names()
             ]
-            node = best_namefit_of(name, children_to_check)
+            if index < key_len - 1:
+                expected_types = ["group"]
+            elif index == key_len - 1:
+                expected_types = ["group"] if not is_last_attr else ["group", "field"]
+            elif index == key_len:
+                expected_types = ["attribute"] if is_last_attr else ["field"]
+            node = best_namefit_of(name, children_to_check, expected_types, check_types)
 
             if node is None:
                 return None
 
         return node
 
-    def is_documented(key: str, tree: NexusNode) -> bool:
+    def is_documented(key: str, tree: NexusNode) -> Tuple[bool, bool]:
         if mapping.get(key) is None:
             # This value is not really set. Skip checking its documentation.
             return True
 
-        node = add_best_matches_for(key, tree)
+        try:
+            node = add_best_matches_for(key, tree, check_types=True)
+        except TypeError:
+            node = None
+            keys_to_remove.append(key)
+
+            collector.collect_and_log(
+                key,
+                ValidationProblem.KeyToBeRemoved,
+                None,
+            )
 
         if node is None:
             key_path = key.replace("@", "")
@@ -1102,7 +1148,7 @@ def validate_dict_against(
         if is_documented(not_visited_key, tree):
             continue
 
-        if not ignore_undocumented:
+        if not ignore_undocumented and not_visited_key not in keys_to_remove:
             collector.collect_and_log(
                 not_visited_key, ValidationProblem.MissingDocumentation, None
             )
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -194,7 +194,6 @@ def best_namefit_of(
                         or (type_attr := elem.attrib.get("type"))
                         and len(type_attr) > 2
                     ]
-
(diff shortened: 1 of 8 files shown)


src/pynxtools/dataconverter/validation.py
```diff
@@ -588,13 +607,30 @@ def handle_unknown_type(node: NexusNode, keys: Mapping[str, Any], prev_path: str
...
                 node.search_add_child_for(child)
                 for child in node.get_all_direct_children_names()
             ]
-            node = best_namefit_of(name, children_to_check)
+            if index < key_len - 1:
+                expected_types = ["group"]
+            elif index == key_len - 1:
+                expected_types = ["group"] if not is_last_attr else ["group", "field"]
+            elif index == key_len:
+                expected_types = ["attribute"] if is_last_attr else ["field"]
```
lukaspie, line 632: This is the problematic line for links. We could change it to `else ["field, "group"]`, but then we cannot find the case where a field is given, but a group is defined in the NeXus definitions.

lukaspie: I suspect this won't be possible just based on the structure of the key itself, we need additional context (like that there is the str "link" in the value that belongs to this key).

lukaspie: Added a workaround that checks if  the str "link" in the value that belongs to this key.

rettigl:
If I add 
```
  "/ENTRY/entry_identifier": {
    "identifier":"@attrs:metadata/loader/scan_path"
  },
```
I get:
```
ERROR: The type ('group') of the given concept 'entry_identifier' conflicts with another existing concept of the same name, which is of type 'field').
WARNING: The attribute /ENTRY[entry]/entry_identifier/identifier will not be written.
```
The validation does not pick up that I try to write a group with a name that is defined as field.
If I add a concept, I get:
```
ERROR: The type ('group') of the given concept 'COLLECTION[entry_identifier]' conflicts with another existing concept of the same name, which is of type 'field').
WARNING: The attribute /ENTRY[entry]/COLLECTION[entry_identifier]/identifier will not be written.
WARNING: Given field name 'COLLECTION' conflicts with the non-variadic name 'entry_identifier (opt)'
```
If I have the conflicting field also in the dict, I get:
```
ERROR: The type ('group') of the given concept 'COLLECTION[entry_identifier]' conflicts with another existing concept of the same name, which is of type 'field').
WARNING: The attribute /ENTRY[entry]/COLLECTION[entry_identifier]/identifier will not be written.
WARNING: Given field name 'COLLECTION' conflicts with the non-variadic name 'entry_identifier (opt)'
```
But the conversion works.

rettigl:
Link checking does not seem to work properly yet. If I add a link
```
    "RESOLUTION[angular_resolution]": {
      "resolution": "@link:/entry/instrument/electronanalyzer/angular_resolution",
      "resolution/@units": "deg",
      "physical_quantity": "angle",
      "type": "estimated"
    },
```
I don't get a warning, even though I linked a group to a defined field name.

lukaspie:
> Link checking does not seem to work properly yet. If I add a link
> 
> ```
>     "RESOLUTION[angular_resolution]": {
>       "resolution": "@link:/entry/instrument/electronanalyzer/angular_resolution",
>       "resolution/@units": "deg",
>       "physical_quantity": "angle",
>       "type": "estimated"
>     },
> ```
> 
> I don't get a warning, even though I linked a group to a defined field name. If I add the same structure manually
> 
> ```
>     "RESOLUTION[angular_resolution]": {
>       "resolution": {
>         "RESOLUTION[angular_resolution]": {
>           "resolution": 1.0
>         }
>       },
>       "resolution/@units": "deg",
>       "physical_quantity": "angle",
>       "type": "estimated"
>     },
> ```
> 
> , i get:
> 
> ```
> ERROR: The type ('group') of the given concept 'resolution' conflicts with another existing concept of the same name, which is of type 'field').
> WARNING: The attribute /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution]/resolution/RESOLUTION[angular_resolution]/resolution will not be written.
> WARNING: Unit /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution]/resolution/@units in dataset without its field /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution]/resolution.
> WARNING: The attribute /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution]/resolution/@units will not be written.
> ```

Yeah,  this is expected behavior as we don't really follow links currently.

The only thing that I implemented w.r.t. links in this PR is that it picks up that if there is a link, the type you expect to write can be either group or field. Like, if we have `ENTRY[entry]/SAMPLE[sample]/temperature_env/temperature_sensor`, from the key structure we expect a field for temperature_sensor. But if we have "link" in the value, we also allow defining groups like this (and don't throw the newly implemented error). 

To actually follow the link and apply all the same checks as for regular fields is something I would implement in a separate PR.

tests/dataconverter/test_validation.py
```diff
@@ -785,6 +785,32 @@ def listify_template(data_dict: Template):
...
         ),
+        pytest.param(
+            alter_dict(
+                TEMPLATE,
+                "/ENTRY[my_entry]/identified_calibration/identifier_1/some_field",
+                "123",
+            ),
+            [
+                "The type ('group') of the given concept 'identifier_1' conflicts with another "
+                "existing concept of the same name, which is of type 'field').",
```
rettigl, line 796:
```suggestion
                "existing concept of the same name, which is of type 'field'.",
```

src/pynxtools/dataconverter/validation.py
```diff
@@ -618,7 +658,17 @@ def is_documented(key: str, tree: NexusNode) -> bool:
             # This value is not really set. Skip checking its documentation.
             return True
 
-        node = add_best_matches_for(key, tree)
+        try:
+            node = add_best_matches_for(key, tree, check_types=True)
+        except TypeError:
+            node = None
```
rettigl, line 664:
```suggestion
             type=node.type
            node = None
```

src/pynxtools/dataconverter/validation.py
```diff
@@ -618,7 +658,17 @@ def is_documented(key: str, tree: NexusNode) -> bool:
...
+            keys_to_remove.append(key)
+
+            collector.collect_and_log(
+                key,
+                ValidationProblem.KeyToBeRemoved,
+                None,
```
rettigl, line 670:
```suggestion
                type,
```

rettigl: + Adopt warning message to include type

tests/dataconverter/test_validation.py
```diff
@@ -785,6 +785,32 @@ def listify_template(data_dict: Template):
...
+        ),
+        pytest.param(
+            alter_dict(
+                TEMPLATE,
+                "/ENTRY[my_entry]/identified_calibration",
+                "123",
+            ),
+            [
+                "The type ('field') of the given concept 'identified_calibration' conflicts with another "
+                "existing concept of the same name, which is of type 'group').",
```
rettigl, line 809:
```suggestion
                "existing concept of the same name, which is of type 'group'.",
```

tests/dataconverter/test_validation.py
```diff
@@ -785,6 +785,32 @@ def listify_template(data_dict: Template):
...
+                "The attribute /ENTRY[my_entry]/identified_calibration will not be written.",
```
rettigl, line 810:
```suggestion
                "The field /ENTRY[my_entry]/identified_calibration will not be written.",
```

tests/dataconverter/test_validation.py
```diff
@@ -785,6 +785,32 @@ def listify_template(data_dict: Template):
...
+        pytest.param(
+            alter_dict(
+                TEMPLATE,
+                "/ENTRY[my_entry]/identified_calibration/identifier_1/some_field",
+                "123",
+            ),
+            [
+                "The type ('group') of the given concept 'identifier_1' conflicts with another "
+                "existing concept of the same name, which is of type 'field').",
+                "The attribute /ENTRY[my_entry]/identified_calibration/identifier_1/some_field will not be written.",
```
rettigl, line 797:
```suggestion
                "The field /ENTRY[my_entry]/identified_calibration/identifier_1/some_field will not be written.",
```

rettigl: This still does not reflect that we are also removing the group in the end.

tests/dataconverter/test_validation.py
```diff
@@ -1299,8 +1325,6 @@ def format_error_message(msg: str) -> str:
             "baseclass-field-with-illegal-unit",
             "open-enum-with-new-item",
             "baseclass-open-enum-with-new-item",
-            "variadic-nxcollection",
-            "nonvariadic-nxcollection",
```
rettigl, line 1303: Why did you remove these here?

rettigl (approved):
Mostly lgtm and seems to do what it should, apart from a few comments

rettigl:
Not sure why the tests fail, though

src/pynxtools/dataconverter/validation.py
lukaspie: Yes, that's correct. Here we are adding "group" to the ["field"] list if there is a link present.

> the mathching of link target type to node type should probably be checked separately.

Turns out this is surprisingly complicated with the current logic. I will try to do it, but may have to change some other things.

src/pynxtools/dataconverter/validation.py
lukaspie: This doesn't quite work, node is None here. Implemented a slightly different logic.

src/pynxtools/dataconverter/validation.py
```diff
@@ -176,6 +194,7 @@ def best_namefit_of(name: str, nodes: Iterable[NexusNode]) -> Optional[NexusNode
                         or (type_attr := elem.attrib.get("type"))
                         and len(type_attr) > 2
                     ]
+
```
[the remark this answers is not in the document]
lukaspie: done

rettigl:
> > Not sure why the tests fail, though
> 
> I think at least for pynxtools-mpes/igor this is related to #641 and [FAIRmat-NFDI/pynxtools-mpes#52](https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/52).

Indeed this seems to work with merged #641: https://github.com/FAIRmat-NFDI/pynxtools-igor/pull/8
I still don't understand why this only occurs now in this PR.

src/pynxtools/dataconverter/validation.py
lukaspie: Not sure if we can even link attributes, considering one should probably write a `@target` attribute in HDF5. I suggest we don't touch this for now until NIAC has cleared up the discussion about links and targets.

rettigl: Yes, exactly my point. I still don't understand the link mechanism in Nexus, and how this differs from plain h5 links

RubelMozumder:
Test from `pynxtools-spm` is failling because of the field and group having the same  name conflict in [NXsample](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/base_classes/NXsample.nxdl.xml#L248) and NXspm is actively using `NXsample_component` group to collect data for `substrate` and `experiment_sample`.

lukaspie:
> Test from `pynxtools-spm` is failling because of the field and group having the same name conflict in [NXsample](https://github.com/FAIRmat-NFDI/nexus_definitions/blob/fairmat/base_classes/NXsample.nxdl.xml?rgh-link-date=2025-05-21T12%3A00%3A40Z#L248) and NXspm is actively using `NXsample_component` group to collect data for `substrate` and `experiment_sample`.

Yes, that's true. I think if you want to use a `NXsample_component` group, you cannot name it `sample_component` as this is blocked for use in fields. Can you rename it?

lukaspie:
@RubelMozumder we now see a bunch of errors and warnings for pynxtools-spm and pynxtools-xrd because we levelled up some of the previous warnings to errors, see [`helpers.py`](https://github.com/FAIRmat-NFDI/pynxtools/pull/638/files#diff-eb8ab2853571e96a900abf9a24b9b7fd9132d48d8974ea6a682dbf05e1669b07). IMO, these should be errors, e.g. if a required field is missing. So I suggest we merge here and then you fix spm and xrd separately.

PR #639, opened 2025-05-15
Validation check for reserved prefixes and suffixes

lukaspie:
NeXus defines [reserved prefixes and suffixes](https://manual.nexusformat.org/datarules.html#index-5) that may only be used in a certain context.

Here, we implement checks for both prefixes and suffixes:
- **prefixes**: these can only be used within a certain context (e.g. `DECTRIS_` prefix is only valid for attributes in `DECTRIS` detectors, i.e., in `NXmx`). Any other use is prohibited. We check the context and log an error if such attributes exist out of context. Note that for some reserved prefixes, we can not write them from pynxtools, so their use is disallowed. We still write these attributes as undocumented, though (could be removed).
- **suffixes**: These are already explicitly defined in NeXus (mostly in `NXobject`, but also in `NXdata` and `NXtransformations`). Note that you can use these suffixes freely outside of their scope (e.g. you can use `FIELDNAME_end`) anyway you like _if_ you are not within `NXtransformations`. Here, we basically check that if a reserved suffix is used, the initial field also exists (e.g. if we define a field `my_field_weights`, we check that `my_field` exists as well.

### DIFF
--- a/src/pynxtools/data/NXtest.nxdl.xml
+++ b/src/pynxtools/data/NXtest.nxdl.xml
@@ -23,6 +23,15 @@
             <field name="required_field" required="true" type="NX_INT">
                 <doc>A dummy entry to test optional parent check for a required child.</doc>
             </field>
+            <field name="required_field_set" optional="true" type="NX_INT">
+                <doc>A dummy entry to test reserved suffixes.</doc>
+            </field>
+            <field name="some_field_set" optional="true" type="NX_INT">
+                <doc>
+                    A dummy entry to test reserved suffixes where the actual field is not given.
+                    Note that this is not allowed by NeXus, but we do this here to test the validation.
+                </doc>
+            </field>
             <field name="optional_field" optional="true" type="NX_INT">
                 <doc>A dummy entry to test optional parent check for an optional child.</doc>
             </field>
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -69,7 +69,7 @@ class ValidationProblem(Enum):
     KeyToBeRemoved = 21
     InvalidConceptForNonVariadic = 22
     ReservedSuffixWithoutField = 23
-    ReservedPrefixInWrongApplication = 24
+    ReservedPrefixInWrongContext = 24
 
 
 class Collector:
@@ -163,7 +163,7 @@ class Collector:
             logger.warning(
                 f"Reserved suffix {path} was used, but there is no associated field {value}."
             )
-        elif log_type == ValidationProblem.ReservedPrefixInWrongApplication:
+        elif log_type == ValidationProblem.ReservedPrefixInWrongContext:
             log_text = f"Reserved prefix {path} was used in key {args[0] if args else '<unknown>'}, but is not valid here."
             if value != "<unknown>":
                 log_text += f" It is only valid in the context of {value}."
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -165,6 +165,7 @@ class Collector:
             )
         elif log_type == ValidationProblem.ReservedPrefixInWrongContext:
             log_text = f"Reserved prefix {path} was used in key {args[0] if args else '<unknown>'}, but is not valid here."
+            # Note that value=None" gets converted to "<unknown>"
             if value != "<unknown>":
                 log_text += f" It is only valid in the context of {value}."
             logger.error(log_text)
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -68,6 +68,8 @@ class ValidationProblem(Enum):
     NXdataAxisMismatch = 20
     KeyToBeRemoved = 21
     InvalidConceptForNonVariadic = 22
+    ReservedSuffixWithoutField = 23
+    ReservedPrefixInWrongApplication = 24
 
 
 class Collector:
@@ -157,6 +159,15 @@ class Collector:
             if value.type == "group":
                 log_text += f", which should be of type {value.nx_class}."
             logger.warning(log_text)
+        elif log_type == ValidationProblem.ReservedSuffixWithoutField:
+            logger.warning(
+                f"Reserved suffix {path} was used, but there is no associated field {value}."
+            )
+        elif log_type == ValidationProblem.ReservedPrefixInWrongApplication:
+            log_text = f"Reserved prefix {path} was used in key {args[0] if args else '<unknown>'}, but is not valid here."
+            if value != "<unknown>":
+                log_text += f" It is only valid in the context of {value}."
+            logger.error(log_text)
 
     def collect_and_log(
         self,
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -161,7 +161,7 @@ class Collector:
             logger.warning(log_text)
         elif log_type == ValidationProblem.ReservedSuffixWithoutField:
             logger.warning(
-                f"Reserved suffix {path} was used, but there is no associated field {value}."
+                f"Reserved suffix '{args[0]}' was used in {path}, but there is no associated field {value}."
             )
         elif log_type == ValidationProblem.ReservedPrefixInWrongContext:
             log_text = f"Reserved prefix {path} was used in key {args[0] if args else '<unknown>'}, but is not valid here."
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -860,7 +860,24 @@ def validate_dict_against(
         return (False, 0)
 
     def check_reserved_suffix(key: str, mapping: MutableMapping[str, Any]) -> bool:
-        """Check if an associated field exists for a key with a reserved suffix."""
+        """
+        Check if an associated field exists for a key with a reserved suffix.
+
+        Reserved suffixes imply the presence of an associated base field (e.g.,
+        "temperature_errors" implies "temperature" must exist in the mapping).
+
+        Args:
+            key (str):
+                The full key path (e.g., "/ENTRY[entry1]/sample/temperature_errors").
+            mapping (MutableMapping[str, Any]):
+                The mapping containing the data to validate.
+                This should be a dict of `/` separated paths.
+
+        Returns:
+            bool:
+                True if the suffix usage is valid or not applicable.
+                False if the suffix is used without the expected associated base field.
+        """
         reserved_suffixes = (
             "_end",
             "_increment_set",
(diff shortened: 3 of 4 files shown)


src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+                "@PDBX_": None,  # do not use anywhere
+                "@SAS_": "NXcanSAS",
+                "@SILX_": None,  # do not use anywhere
+            },
+            "field": {
+                "DECTRIS_": "NXmx",
+            },
+        }
+
+        prefixes = reserved_prefixes[nx_type]
```
rettigl, line 912: This will give a key error when nx_type == "group" (mabye never happens, but then restrict type of nx_type)

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+
+        if not key.rsplit("/", 1)[-1].startswith(tuple(prefixes.keys())):
+            # Ignore this test
+            return False
```
rettigl, line 916: True?

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+                collector.collect_and_log(
+                    prefix,
+                    ValidationProblem.ReservedPrefixInWrongApplication,
+                    None,
+                    key,
+                )
+                return False
+            elif context == "all":
+                # We can freely use this prefix everywhere.
+                continue
```
rettigl, line 932: Can't we return already here? We only end up here if the key starts with the prefix, so we probably don't have to check others.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -157,6 +159,15 @@ def _log(self, path: str, log_type: ValidationProblem, value: Optional[Any], *ar
             if value.type == "group":
                 log_text += f", which should be of type {value.nx_class}."
             logger.warning(log_text)
+        elif log_type == ValidationProblem.ReservedSuffixWithoutField:
+            logger.warning(
+                f"Reserved suffix {path} was used, but there is no associated field {value}."
+            )
+        elif log_type == ValidationProblem.ReservedPrefixInWrongApplication:
+            log_text = f"Reserved prefix {path} was used in key {args[0] if args else '<unknown>'}, but is not valid here."
+            if value != "<unknown>":
```
rettigl, line 168: What does value contain here? How would it be filled by `<unknown>`?

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+
+        for prefix, context in prefixes.items():
+            if not key.rsplit("/", 1)[-1].startswith(prefix):
+                continue
+            if not context:
+                # This prefix should not be used by pynxtools.
+                collector.collect_and_log(
+                    prefix,
+                    ValidationProblem.ReservedPrefixInWrongApplication,
+                    None,
```
rettigl, line 926: This sends "none" and not `<unknown>`

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+                if match:
+                    definition_key = f"{match.group(1)}/definition"
+                    if mapping.get(definition_key) != context:
+                        collector.collect_and_log(
+                            prefix,
+                            ValidationProblem.ReservedPrefixInWrongApplication,
+                            context,
+                            key,
+                        )
+                    return False
```
rettigl, line 945: Should this not only return false if we log the warning (i.e. one level down)?

src/pynxtools/dataconverter/validation.py
```diff
@@ -937,6 +1036,18 @@ def startswith_with_variations(
...
+                not_visited_key, mapping, "field"
+            ):
+                continue
+        else:
+            associated_field = not_visited_key.rsplit("/", 1)[-2]
+            if check_reserved_prefix(
+                not_visited_key, mapping, "attribute"
+            ) and check_reserved_prefix(associated_field, mapping, "field"):
+                continue
+
```
rettigl, line 1050: This means you skip the test for documentation, if these tests return true. Is that really what you want to do?

src/pynxtools/data/NXtest.nxdl.xml
```diff
@@ -23,6 +23,15 @@
...
             </field>
+            <field name="required_field_set" optional="true" type="NX_INT">
+                <doc>A dummy entry to test reserved suffixes.</doc>
+            </field>
+            <field name="some_field_set" optional="true" type="NX_INT">
+                <doc>
+                    A dummy entry to test reserved suffixes where the actual field is not given.
+                    Note that this is not allowed by NeXus, but we do this here to test the validation.
+                </doc>
+            </field>
```
rettigl, line 34: Why should they be part of the definition? I think they are (or not) defined from inheritance?

tests/dataconverter/test_validation.py
```diff
@@ -1216,13 +1216,75 @@ def listify_template(data_dict: Template):
...
+                "It is only valid in the context of NXmx.",
+                "Field /ENTRY[my_entry]/OPTIONAL_group[my_group]/DECTRIS_field written without documentation.",
+            ],
+            id="reserved-prefix",
+        ),
     ],
 )
 def test_validate_data_dict(caplog, data_dict, error_messages, request):
     """Unit test for the data validation routine."""
+    # validate_dict_against("NXtest", data_dict)[0]
```
rettigl, line 1281: ?

tests/dataconverter/test_validation.py
```diff
@@ -1216,13 +1216,75 @@ def listify_template(data_dict: Template):
...
+                "Attribute /ENTRY[my_entry]/OPTIONAL_group[my_group]/@BLUESKY_attr written without documentation.",
+                "Reserved prefix @DECTRIS_ was used in key /ENTRY[my_entry]/OPTIONAL_group[my_group]/@DECTRIS_attr, but is not valid here. "
+                "It is only valid in the context of NXmx.",
+                "Attribute /ENTRY[my_entry]/OPTIONAL_group[my_group]/@DECTRIS_attr written without documentation.",
+                "Reserved prefix DECTRIS_ was used in key /ENTRY[my_entry]/OPTIONAL_group[my_group]/DECTRIS_field, but is not valid here. "
+                "It is only valid in the context of NXmx.",
+                "Field /ENTRY[my_entry]/OPTIONAL_group[my_group]/DECTRIS_field written without documentation.",
+            ],
+            id="reserved-prefix",
+        ),
```
rettigl, line 1276: This does only test the invalid cases, not the valid cases.

rettigl (commented):
I don't completely oversee how this is intended to work, but generally looks reasonable to me apart from a few questions

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
         # default
         return (False, 0)
 
+    def check_reserved_suffix(key: str, mapping: MutableMapping[str, Any]) -> bool:
+        """Check if an associated field exists for a key with a reserved suffix."""
```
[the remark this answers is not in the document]
lukaspie: done

src/pynxtools/dataconverter/validation.py
```diff
@@ -847,6 +859,93 @@ def startswith_with_variations(
...
+            "_mask",
+            "_set",
+            "_weights",
+            "_scaling_factor",
+            "_offset",
+        )
+
+        if not key.endswith((reserved_suffixes)):
+            # Ignore this test
+            return False
```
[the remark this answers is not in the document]
lukaspie: Yeah, makes sense. See comment about disentangling below.

src/pynxtools/dataconverter/validation.py
lukaspie: I was initially thinking that for reserved suffixes, we can just skip here (as they are anyway defined), but it's probably best to just disentangle these two checks. Changed it to just go on if the prefix/suffix checks are passed here.

src/pynxtools/dataconverter/validation.py
lukaspie: Yes. I changed the logic a bit, but used your comment.

src/pynxtools/dataconverter/helpers.py
lukaspie: See comment below, `value=None` gets converted to <unknown>. Not sure if this is great for maintenance, but works for now.

lukaspie:
@rettigl fixed most of what you were suggesting.

Before, I was also missing that fields with reserved suffixes must be written as e.g. `FIELDNAME_weights[my_field_weights]`. This makes the checking of associated fields more complicated, due to the concept  name and brackets. Basically, I check that a key that starts with the same parent path and that either ends on `my_field` or `[my_field]` exists. A bit clunky, but seems to work for now.

src/pynxtools/data/NXtest.nxdl.xml
rettigl: makes sense

src/pynxtools/dataconverter/validation.py
rettigl: Maybe add a comment about that somewhere.

src/pynxtools/dataconverter/validation.py
```diff
@@ -182,6 +193,9 @@ def best_namefit_of(name: str, nodes: Iterable[NexusNode]) -> Optional[NexusNode
                         return None
                 return node
         else:
+            PRINT_HERE = True if "weights" in name else False
+            # if PRINT_HERE:
+            #     print(node)
```
rettigl, line 198: remove?

rettigl: You can use debug mode and conditional break points.

src/pynxtools/dataconverter/validation.py
```diff
@@ -589,12 +604,18 @@ def handle_unknown_type(node: NexusNode, keys: Mapping[str, Any], prev_path: str
...
             children_to_check = [
                 node.search_add_child_for(child)
                 for child in node.get_all_direct_children_names()
             ]
+            if PRINT:
+                print("<<<<<<<", name, node)
             node = best_namefit_of(name, children_to_check)
+            if PRINT:
+                print("\t", children_to_check)
+                print(">>>>>>", node)
```
rettigl, line 618: Remove debug code

rettigl (approved):
LGTM now, apart from remaining debug code

src/pynxtools/dataconverter/validation.py
lukaspie: done

src/pynxtools/dataconverter/validation.py
lukaspie: Yeah, I usually do, this was just for a quick fix, sorry that I kept it in. Have removed all occurences.

src/pynxtools/dataconverter/validation.py
lukaspie: Done

PR #640, opened 2025-05-20
pick up `AXISNAME_indices` attribute in `read_nexus` tool

lukaspie:


lukaspie:
Needs to be fixed on the writer level, closing here.

PR #641, opened 2025-05-20
handle variable-name attributes in writer

lukaspie:
For variable-name attributes, we are now using this notation in the template: `"@AXISNAME_indices[energy_indices]"` (see also https://github.com/FAIRmat-NFDI/pynxtools-mpes/pull/52/files). The existing logic was interpreting this to write an HDF5 attribute  `@energy_indices`, but we don't actuallly want to write the `@` into the file. With this PR, the name of the attribute is `energy_indices`, as expected.

### DIFF
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -590,8 +590,10 @@ def get_name_from_data_dict_entry(entry: str) -> str:
     results = get_regex().search(entry)
     if results is None:
         return entry
+
     if entry[0] == "@":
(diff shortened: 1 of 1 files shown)


lukaspie:
> > For variable-name attributes, we are not using this notation in the template
> 
> Are you meaning " we are now using this notation"?
> 
> I thought we agreed on `"@AXISNAME_indices[@energy_indices]"`

Yes, sorry. We are using  `"@AXISNAME_indices[@energy_indices]"`, correct.

rettigl (approved):
LGTM

PR #642, opened 2025-05-21
Follow links and resolve

lukaspie:
@rettigl with this PR, we properly follow links as discussed in https://github.com/FAIRmat-NFDI/pynxtools/pull/638#issuecomment-2891469373 and raise an error if the type of the key does not match the type of the linked entity. This works both for concepts from appdefs as well as from base classes (see the two new tests).

I had to make a lot of changes on the way unfortunately, hence this new PR (pointing to my previous one):
- Link following was completely changed. Before, we could only follow if a single field within a group was linked, but already a structure like 
  ```
  "/ENTRY[my_entry]/OPTIONAL_group[some_group]/required_field": {"link": ..."},
  "/ENTRY[my_entry]/OPTIONAL_group[some_group]/optional_field": 1.0,
  ```
  could not be resolved.
- Now also following links when we check in the `is_documented` part for concepts from the base classes.

Further improvements:
- ValidationProblems are auto-enumerated (meaning we can add new ones anywhere)
- Keys to removed are stored along the way and then removed directly in the `validate_data_dict` function (so we don't have to pass them to the converter) -> `validate_dict_against` just returns a boolean, as before.

Questions:
- Can you double check that your initial use case works now?
- Do we want to remove keys where the type of the linked entity doesn't match?
- More broadly, do we want to remove fields/groups if they lead to `ExpectedGroup/ExpectedField` errors?

### DIFF
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 #
 import re
+import copy
 from collections import defaultdict
 from functools import reduce
 from operator import getitem
@@ -461,6 +462,7 @@ def validate_dict_against(
                 continue
             nx_class, _ = split_class_and_name_of(variant)
             if not isinstance(keys[variant], Mapping):
+                # Groups should have subelements
                 if nx_class is not None:
                     collector.collect_and_log(
                         variant_path,
@@ -495,6 +497,10 @@ def validate_dict_against(
         Resolves internal dictionary "links" by replacing any keys containing a
         {"link": "/path/to/target"} structure with the actual referenced content.
 
+        Note that the keys are only replaced in copies of the incoming keys, NOT in
+        the gloval mapping. That is, links are resolved for checking, but we still write
+        links into the HDF5 file.
+
         This function traverses the mapping and recursively resolves any keys that
         contain a "link" to another path (relative to the global template).
         If the link cannot be resolved, the issue is logged.
@@ -514,8 +520,6 @@ def validate_dict_against(
 
         if not isinstance(keys, dict):
             return keys
-
-        import copy
 
         resolved_keys = copy.deepcopy(keys)
         for key, value in keys.copy().items():
@@ -563,6 +567,8 @@ def validate_dict_against(
             if isinstance(keys[variant], Mapping) and not all(
                 k.startswith("@") for k in keys[variant]
             ):
+                # A field should not have a dict of keys that are _not_ all attributes,
+                # i.e. no sub-fields or sub-groups.
                 collector.collect_and_log(
                     variant_path,
                     ValidationProblem.ExpectedField,
@@ -766,7 +772,7 @@ def validate_dict_against(
                     ValidationProblem.ExpectedGroup,
                     None,
                 )
-                # TODO: decide if we want to remove such keys (and below)
+                # TODO: decide if we want to remove such keys
                 # collector.collect_and_log(
                 #     key,
                 #     ValidationProblem.KeyToBeRemoved,
@@ -776,7 +782,8 @@ def validate_dict_against(
                 return False
 
             elif node.type == "field":
-                # Field should only have values.
+                # A field should not have a dict of keys that are _not_ all attributes,
+                # i.e. no sub-fields or sub-groups.
                 if is_mapping and not all(
                     k.startswith("@") for k in resolved_link[key]
                 ):
@@ -785,7 +792,7 @@ def validate_dict_against(
                         ValidationProblem.ExpectedField,
                         None,
                     )
-                    # TODO: decide if we want to remove such keys (and below)
+                    # TODO: decide if we want to remove such keys
                     # collector.collect_and_log(
                     #     key,
                     #     ValidationProblem.KeyToBeRemoved,
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -467,6 +467,13 @@ def validate_dict_against(
                         ValidationProblem.ExpectedGroup,
                         None,
                     )
+                    # TODO: decide if we want to remove such keys
+                    # collector.collect_and_log(
+                    #     variant_path,
+                    #     ValidationProblem.KeyToBeRemoved,
+                    #     node.type,
+                    # )
+                    # keys_to_remove.append(not_visited_key)
                 continue
             if node.nx_class == "NXdata":
                 handle_nxdata(node, keys[variant], prev_path=variant_path)
@@ -497,8 +504,6 @@ def validate_dict_against(
                 May be None or a non-dict value, in which case it's returned as-is.
             prev_path (str): The path leading up to the current `keys` context, used
                 for logging and error reporting.
-            p (bool, optional): Unused parameter (possibly for debugging); included
-                for interface compatibility. Defaults to False.
 
         Returns:
             Optional[Any]: A dictionary with resolved links, the original value if
@@ -510,7 +515,9 @@ def validate_dict_against(
         if not isinstance(keys, dict):
             return keys
 
-        resolved_keys = keys
+        import copy
+
+        resolved_keys = copy.deepcopy(keys)
         for key, value in keys.copy().items():
             if isinstance(value, dict) and len(value) == 1 and "link" in value:
                 key_path = f"{prev_path}/{key}" if prev_path else key
@@ -561,6 +568,13 @@ def validate_dict_against(
                     ValidationProblem.ExpectedField,
                     None,
                 )
+                # TODO: decide if we want to remove such keys
+                # collector.collect_and_log(
+                #     variant_path,
+                #     ValidationProblem.KeyToBeRemoved,
+                #     node.type,
+                # )
+                # keys_to_remove.append(variant_path)
                 continue
             if node.optionality == "required" and isinstance(keys[variant], Mapping):
                 # Check if all fields in the dict are actual attributes (startswith @)
@@ -713,13 +727,13 @@ def validate_dict_against(
         except TypeError:
             node = None
             nx_type = "attribute" if key.split("/")[-1].startswith("@") else "field"
-            keys_to_remove.append(key)
 
             collector.collect_and_log(
                 key,
                 ValidationProblem.KeyToBeRemoved,
                 nx_type,
             )
+            keys_to_remove.append(key)
 
         if node is None:
             key_path = key.replace("@", "")
@@ -752,12 +766,13 @@ def validate_dict_against(
                     ValidationProblem.ExpectedGroup,
                     None,
                 )
+                # TODO: decide if we want to remove such keys (and below)
                 # collector.collect_and_log(
                 #     key,
                 #     ValidationProblem.KeyToBeRemoved,
                 #     "group",
                 # )
-                keys_to_remove.append(key)
+                # keys_to_remove.append(key)
                 return False
 
             elif node.type == "field":
@@ -770,13 +785,14 @@ def validate_dict_against(
                         ValidationProblem.ExpectedField,
                         None,
                     )
+                    # TODO: decide if we want to remove such keys (and below)
                     # collector.collect_and_log(
                     #     key,
                     #     ValidationProblem.KeyToBeRemoved,
                     #     "field",
                     # )
-                    keys_to_remove.append(key)
-                    return False
+                    # keys_to_remove.append(key)
+                    # return False
                 resolved_link[key] = is_valid_data_field(
                     resolved_link[key], node.dtype, node.items, node.open_enum, key
                 )
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -760,11 +760,11 @@ def validate_dict_against(
                 keys_to_remove.append(key)
                 return False
 
-            elif node.type == "field" and not all(
-                k.startswith("@") for k in resolved_link[key]
-            ):
+            elif node.type == "field":
                 # Field should only have values.
-                if is_mapping:
+                if is_mapping and not all(
+                    k.startswith("@") for k in resolved_link[key]
+                ):
                     collector.collect_and_log(
                         key,
                         ValidationProblem.ExpectedField,
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -484,6 +484,26 @@ def validate_dict_against(
     def _follow_link(
         keys: Optional[Mapping[str, Any]], prev_path: str, p=False
     ) -> Optional[Any]:
+        """
+        Resolves internal dictionary "links" by replacing any keys containing a
+        {"link": "/path/to/target"} structure with the actual referenced content.
+
+        This function traverses the mapping and recursively resolves any keys that
+        contain a "link" to another path (relative to the global template).
+        If the link cannot be resolved, the issue is logged.
+
+        Args:
+            keys (Optional[Mapping[str, Any]]): The dictionary structure to process.
+                May be None or a non-dict value, in which case it's returned as-is.
+            prev_path (str): The path leading up to the current `keys` context, used
+                for logging and error reporting.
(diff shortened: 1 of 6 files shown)


src/pynxtools/dataconverter/validation.py
```diff
@@ -690,7 +741,46 @@ def is_documented(key: str, tree: NexusNode) -> bool:
...
+            is_mapping = isinstance(resolved_link[key], Mapping)
+
+            if node.type == "group" and not is_mapping:
+                # Groups must have subelements
+                collector.collect_and_log(
+                    key,
+                    ValidationProblem.ExpectedGroup,
+                    None,
+                )
+                # collector.collect_and_log(
```
lukaspie, line 755: remove if not matching?

rettigl:
I don't have time for a full review, but these are my observations:
- Valid "field"-level link:
```
    "RESOLUTION[angular_resolution]": {
      "resolution": "@link:/entry/instrument/electronanalyzer/angular_resolution/resolution"
    },
```
gives:
```
File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validation.py:763, in validate_dict_against.<locals>.is_documented(key, tree)
    760     keys_to_remove.append(key)
    761     return False
--> 763 elif node.type == "field" and not all(
    764     k.startswith("@") for k in resolved_link[key]
    765 ):
    766     # Field should only have values.
    767     if is_mapping:
    768         collector.collect_and_log(
    769             key,
    770             ValidationProblem.ExpectedField,
    771             None,
    772         )

TypeError: 'float' object is not iterable
```

- Other valid "field"-level link:
"calibrated_axis": "@link:/entry/data/angular0",
gives:
```
File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validation.py:764, in <genexpr>(.0)
    760     keys_to_remove.append(key)
    761     return False
    763 elif node.type == "field" and not all(
--> 764     k.startswith("@") for k in resolved_link[key]
    765 ):
    766     # Field should only have values.
    767     if is_mapping:
    768         collector.collect_and_log(
    769             key,
    770             ValidationProblem.ExpectedField,
    771             None,
    772         )

AttributeError: 'numpy.float64' object has no attribute 'startswith'
```
- Valid "group" level link works:
`    "RESOLUTION[angular_resolution]": "@link:/entry/instrument/electronanalyzer/angular_resolution",`
- Invalid "field" to "group" level 
```
    "RESOLUTION[angular_resolution]": {
      "resolution": "@link:/entry/instrument/electronanalyzer/angular_resolution"
    },
```
gives correct error:
`ERROR: Expected a field at /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution]/resolution but found a group.`
- Invalid "Field" to "Group" link 
`    "RESOLUTION[angular_resolution]": "@link:/entry/instrument/electronanalyzer/angular_resolution/resolution",`
gives 
`ERROR: Expected a group at /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution] but found a field or attribute.`

lukaspie:
> I don't have time for a full review, but these are my observations:
> 
> * Valid "field"-level link:
> 
> ```
>     "RESOLUTION[angular_resolution]": {
>       "resolution": "@link:/entry/instrument/electronanalyzer/angular_resolution/resolution"
>     },
> ```
> 
> gives:
> 
> ```
> File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validation.py:763, in validate_dict_against.<locals>.is_documented(key, tree)
>     760     keys_to_remove.append(key)
>     761     return False
> --> 763 elif node.type == "field" and not all(
>     764     k.startswith("@") for k in resolved_link[key]
>     765 ):
>     766     # Field should only have values.
>     767     if is_mapping:
>     768         collector.collect_and_log(
>     769             key,
>     770             ValidationProblem.ExpectedField,
>     771             None,
>     772         )
> 
> TypeError: 'float' object is not iterable
> ```
> 
> * Other valid "field"-level link:
>   "calibrated_axis": "@link:/entry/data/angular0",
>   gives:
> 
> ```
> File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validation.py:764, in <genexpr>(.0)
>     760     keys_to_remove.append(key)
>     761     return False
>     763 elif node.type == "field" and not all(
> --> 764     k.startswith("@") for k in resolved_link[key]
>     765 ):
>     766     # Field should only have values.
>     767     if is_mapping:
>     768         collector.collect_and_log(
>     769             key,
>     770             ValidationProblem.ExpectedField,
>     771             None,
>     772         )
> 
> AttributeError: 'numpy.float64' object has no attribute 'startswith'
> ```

These were coming from the same issue which should be fixed with [1976f30](https://github.com/FAIRmat-NFDI/pynxtools/pull/642/commits/1976f300711feee596b3390f30a65148a6d1b381). Namely, the logic for checking linked fields (in the validation against base classes) was ordered incorrectly. Should be fixed now, can you please try again? I also added a test for this in that commit.

rettigl:
> These were coming from the same issue which should be fixed with [1976f30](https://github.com/FAIRmat-NFDI/pynxtools/pull/642/commits/1976f300711feee596b3390f30a65148a6d1b381). Namely, the logic for checking linked fields (in the validation against base classes) was ordered incorrectly. Should be fixed now, can you please try again? I also added positive tests for links (both for appdef and base class concepts) in that and the next commit.

It seems to work now.
However, it silently removes keys that are conflicting, and does not inform the user that it does this:
`ERROR: Expected a group at /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution] but found a field or attribute.`
From this I don't find it obvious that the key will be removed...

lukaspie:
> > These were coming from the same issue which should be fixed with [1976f30](https://github.com/FAIRmat-NFDI/pynxtools/pull/642/commits/1976f300711feee596b3390f30a65148a6d1b381). Namely, the logic for checking linked fields (in the validation against base classes) was ordered incorrectly. Should be fixed now, can you please try again? I also added positive tests for links (both for appdef and base class concepts) in that and the next commit.
> 
> It seems to work now. However, it silently removes keys that are conflicting, and does not inform the user that it does this: `ERROR: Expected a group at /ENTRY[entry]/INSTRUMENT[instrument]/RESOLUTION[angular_resolution] but found a field or attribute.` From this I don't find it obvious that the key will be removed...

Great that it works. Currently, I have the warning messages disabled, but indeed we remove these keys. This was my question, shall we remove them in the first place? More broadly, do we want to remove fields/groups if they lead to ExpectedGroup/ExpectedField errors (not just from links)? If so, obviously we can enable the messages again.

rettigl:
> This was my question, shall we remove them in the first place? More broadly, do we want to remove fields/groups if they lead to ExpectedGroup/ExpectedField errors (not just from links)?

Maybe we only remove them if they lead to write errors because of conflicts in the dict? Or we just raise an error in that case.

lukaspie:
> > This was my question, shall we remove them in the first place? More broadly, do we want to remove fields/groups if they lead to ExpectedGroup/ExpectedField errors (not just from links)?
> 
> Maybe we only remove them if they lead to write errors because of conflicts in the dict? Or we just raise an error in that case.

I think it's not possible to predict if we'll get a write error while we are validating. The original error that we got here only became a problem because of the `deprecated` attribute.

I think for the original error in the writer we can just do
```
if nx_class := attrs.get("type"):
    grp.attrs["NX_class"] = nx_class
 else:
    logger.error(f"No attribute 'NX_class' could be written for {parent_path}.")
```
Then we don't even need to raise anymore and we can keep the keys in the template (at least for now, maybe there will be further problems later).

lukaspie:
EDIT: if we don't remove such keys, we get warnings like `Expected a group at /ENTRY[my_entry]/USER[my_user] but found a field or attribute.`

I think this is more confusing then removing the keys.

src/pynxtools/dataconverter/validation.py
```diff
@@ -515,6 +551,17 @@ def handle_field(node: NexusNode, keys: Mapping[str, Any], prev_path: str):
...
+
+            if isinstance(keys[variant], Mapping) and not all(
+                k.startswith("@") for k in keys[variant]
+            ):
+                collector.collect_and_log(
+                    variant_path,
+                    ValidationProblem.ExpectedField,
+                    None,
+                )
+                continue
```
rettigl, line 564: This is quite hard for me to follow, maybe add some notes

src/pynxtools/dataconverter/validation.py
```diff
@@ -524,44 +571,47 @@ def handle_field(node: NexusNode, keys: Mapping[str, Any], prev_path: str):
...
                     collector.collect_and_log(
-                        f"{prev_path}/{variant}",
+                        variant_path,
                         ValidationProblem.MissingUnit,
                         node.unit,
                     )
                 # TODO: Check unit with pint
 
+            field_attributes = get_field_attributes(variant, keys)
+            field_attributes = _follow_link(field_attributes, variant_path)
```
rettigl, line 609: Will this replace links in the dict by copied data, or will the links be maintained to be written as links into the file?

src/pynxtools/dataconverter/validation.py
```diff
@@ -455,53 +455,89 @@ def handle_group(node: NexusGroup, keys: Mapping[str, Any], prev_path: str):
...
                     )
                 continue
             if node.nx_class == "NXdata":
-                handle_nxdata(node, keys[variant], prev_path=f"{prev_path}/{variant}")
+                handle_nxdata(node, keys[variant], prev_path=variant_path)
             if node.nx_class == "NXcollection":
                 return
             else:
-                recurse_tree(node, keys[variant], prev_path=f"{prev_path}/{variant}")
+                variant_keys = _follow_link(keys[variant], variant_path)
```
rettigl, line 476: Will this replace links in the dict by copied data, or will the links be maintained to be written as links into the file?

src/pynxtools/dataconverter/validation.py
```diff
@@ -690,7 +741,46 @@ def is_documented(key: str, tree: NexusNode) -> bool:
             return True
 
         if isinstance(mapping[key], dict) and "link" in mapping[key]:
-            # TODO: Follow link and check consistency with current field
+            resolved_link = _follow_link({key: mapping[key]}, "")
+
+            is_mapping = isinstance(resolved_link[key], Mapping)
+
+            if node.type == "group" and not is_mapping:
+                # Groups must have subelements
```
rettigl, line 749: Technically, I think this is not necessary, neither by h5 nor by Nexus, no?

src/pynxtools/dataconverter/validation.py
rettigl: If we remove, we should log.

src/pynxtools/dataconverter/validation.py
```diff
@@ -795,14 +879,8 @@ def check_attributes_of_nonexisting_field(
                         collector.collect_and_log(
                             key[0:last_index],
                             ValidationProblem.AttributeForNonExistingField,
-                            None,
-                        )
-                        collector.collect_and_log(
-                            key,
-                            ValidationProblem.KeyToBeRemoved,
```
rettigl, line 802: Where is this logged now (that we remove the key)?

src/pynxtools/dataconverter/validation.py
```diff
@@ -1160,7 +1241,11 @@ def check_reserved_prefix(
     # clear lru_cache
     NexusNode.search_add_child_for.cache_clear()
 
-    return (not collector.has_validation_problems(), keys_to_remove)
+    # remove keys that are incorrect
+    for key in set(keys_to_remove):
+        del mapping[key]
```
rettigl, line 1246: Should we give the warning here?

rettigl (approved):
I can't say I understand all the changes to the logic here. So at least some more comments in the code would be good for later code maintenance.
Otherwise, it appears to do the correct thing (acccording to my few tests).

src/pynxtools/dataconverter/validation.py
```diff
@@ -455,53 +455,89 @@ def handle_group(node: NexusGroup, keys: Mapping[str, Any], prev_path: str):
...
+        contain a "link" to another path (relative to the global template).
+        If the link cannot be resolved, the issue is logged.
+
+        Args:
+            keys (Optional[Mapping[str, Any]]): The dictionary structure to process.
+                May be None or a non-dict value, in which case it's returned as-is.
+            prev_path (str): The path leading up to the current `keys` context, used
+                for logging and error reporting.
+            p (bool, optional): Unused parameter (possibly for debugging); included
+                for interface compatibility. Defaults to False.
```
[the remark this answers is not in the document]
lukaspie: Ah yes, had added this to test something. Will remove.

src/pynxtools/dataconverter/validation.py
```diff
@@ -455,53 +455,89 @@ def handle_group(node: NexusGroup, keys: Mapping[str, Any], prev_path: str):
...
+
+        Returns:
+            Optional[Any]: A dictionary with resolved links, the original value if
+            `keys` is not a dict, or None if `keys` is None.
+        """
```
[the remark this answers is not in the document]
lukaspie: We replace them temporarily for checking, but keep it in a separate dict. So we can test them as if they were actually plain groups/fields, but we still write links in the end.

src/pynxtools/dataconverter/validation.py
```diff
@@ -455,53 +455,89 @@ def handle_group(node: NexusGroup, keys: Mapping[str, Any], prev_path: str):
...
+                    collector.collect_and_log(
+                        key_path,
+                        ValidationProblem.KeyToBeRemoved,
+                        "key",
                     )
-                    return None
-                current_keys = current_keys[dict_path_elem]
-            return current_keys
-        return keys
+                    del resolved_keys[key]
```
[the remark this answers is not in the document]
lukaspie: just for checking, see above

src/pynxtools/dataconverter/validation.py
lukaspie: It's basically saying "A field should not have a dict of keys that are _not_ all attributes, i.e. no sub-fields or sub-groups".

Added this as a comment.

src/pynxtools/dataconverter/validation.py
lukaspie: See above, this is just for checking. We are not making changes to the `mapping` dict.

src/pynxtools/dataconverter/validation.py
lukaspie: see above

src/pynxtools/dataconverter/validation.py
lukaspie: So then we should also allow keys to be None?

src/pynxtools/dataconverter/validation.py
lukaspie: For now, I have decided not to remove. The logic happens exactly here, through `ValidationProblem.KeyToBeRemoved`.

src/pynxtools/dataconverter/validation.py
lukaspie: I thought above this, but basically we are losing the context if we are removing a group/field/attribute. With the current version, we also get the messages why it is an error and then directly afterwards that the key is removed, I quite like that.

src/pynxtools/dataconverter/validation.py
lukaspie: Not that we always log `KeyToBeRemoved` exactly when we add to `keys_to_remove`, not when we actually remove (see comment below). This allows us to pass valuable context to the logger.

src/pynxtools/dataconverter/validation.py
rettigl: okay, makes sense, was just a thought I had here. Maybe add a comment

src/pynxtools/dataconverter/validation.py
rettigl: I don't understand. there is still line 901, where the key is removed, but you removed the corresponding warning (?)

PR #646, opened 2025-05-26
Error for multiple different variadic concepts with the same name

lukaspie:
As discussed in #636 and #638, we should not allow the same instance name for two unnamed groups, e.g. `NXuser` and `NXsample`. We implement this here by running a check on the whole mapping table before any of the other validation starts. Note that we _have_ to remove such keys as they will necessarily lead to HDF5 conflicts (HDF5 depends on unique names for groups and datasets).

Error message may need some improvement, wasn't too sure how to report this issue.

### DIFF
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -911,13 +911,32 @@ def validate_dict_against(
                     ValidationProblem.DifferentVariadicNodesWithTheSameName,
                     entries,
                 )
+                # Now that we have name conflicts, we still need to check that there are
+                # at least two valid keys in that conclit. Only then we remove these.
+                # This takes care of the example with keys like
+                # /ENTRY[my_entry]/USER[some_name]/name and /ENTRY[my_entry]/USERS[some_name]/name,
+                #  where we only want to keep the first one.
+                valid_keys_with_name_conflicts = []
+
                 for key in keys:
+                    try:
+                        node = add_best_matches_for(key, tree)
+                        if node is not None:
+                            valid_keys_with_name_conflicts.append(key)
+                            continue
+                    except TypeError:
+                        pass
                     collector.collect_and_log(
-                        key,
-                        ValidationProblem.KeyToBeRemoved,
-                        "key",
+                        key, ValidationProblem.KeyToBeRemoved, "key"
                     )
-                keys_to_remove += keys
+                    keys_to_remove.append(key)
+
+                if len(valid_keys_with_name_conflicts) > 1:
+                    for valid_key in valid_keys_with_name_conflicts:
+                        collector.collect_and_log(
+                            valid_key, ValidationProblem.KeyToBeRemoved, "key"
+                        )
+                        keys_to_remove.append(valid_key)
 
     def check_attributes_of_nonexisting_field(
         node: NexusNode,
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -898,6 +898,30 @@ def validate_dict_against(
 
                 # Determine the parent path up to just before this match
                 parent_path = key[: match.start()]
+                child_path = key[match.start() :].split("/", 1)[-1]
+
+                # Here we check if for this key with a concept name, another valid key
+                # with a non-concept name exists.
+                non_concept_key = f"{parent_path}{instance_name}/{child_path}"
+
+                if non_concept_key in mapping:
+                    try:
+                        node = add_best_matches_for(non_concept_key, tree)
+                        if node is not None:
+                            collector.collect_and_log(
+                                key,
+                                ValidationProblem.KeysWithAndWithoutConcept,
+                                non_concept_key,
+                                concept_name,
+                            )
+                            collector.collect_and_log(
+                                key, ValidationProblem.KeyToBeRemoved, "key"
+                            )
+                            keys_to_remove.append(key)
+                            continue
+                    except TypeError:
+                        pass
+
                 instance_usage[(instance_name, parent_path)].append((concept_name, key))
 
         for (instance_name, parent_path), entries in sorted(instance_usage.items()):
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -929,12 +929,23 @@ def validate_dict_against(
                     )
                     keys_to_remove.append(key)
 
-                if len(valid_keys_with_name_conflicts) > 1:
-                    for valid_key in valid_keys_with_name_conflicts:
-                        collector.collect_and_log(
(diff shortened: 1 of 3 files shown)


src/pynxtools/dataconverter/validation.py
```diff
@@ -850,6 +850,72 @@ def recurse_tree(
...
+                in the case of conflicts.
+
+        """
+        pattern = re.compile(r"(?P<concept_name>[^\[\]/]+)\[(?P<instance>[^\]]+)\]")
+
+        # Tracks instance usage with respect to their parent group
+        instance_usage: Dict[Tuple[str, str], List[Tuple[str, str]]] = defaultdict(list)
+
+        for key in mapping:
+            matches = list(pattern.finditer(key))
```
RubelMozumder, line 893: value can be checked if it is none; if so, the key can be removed.

src/pynxtools/dataconverter/validation.py
```diff
@@ -850,6 +850,72 @@ def recurse_tree(
...
+            for i, match in enumerate(matches):
```
RubelMozumder, line 894: Should the matches be enumerated? I do not see `i` has been used anywhere.

RubelMozumder (changes requested):
LGTM!

src/pynxtools/dataconverter/validation.py
lukaspie: Yeah, enumeration is not needed, will remove

src/pynxtools/dataconverter/validation.py
lukaspie: Implemented my previous comment

src/pynxtools/dataconverter/validation.py
```diff
@@ -850,6 +850,75 @@ def recurse_tree(
...
+        But this is NOT a conflict:
+            /ENTRY[entry1]/INSTRUMENT[instrument]/FIELD[my_name]/...
+            /ENTRY[entry1]/INSTRUMENT[instrument]/DETECTOR[detector]/FIELD[my_name]/...
+
+        When such conflicts are found, an error is logged indicating the instance name
+        and the conflicting concept names. Additionally, all keys involved in the conflict
+        are logged and added to the `keys_to_remove` list.
+
+        Parameters:
+            mapping (MutableMapping[str, str]):
```
RubelMozumder, line 878: This code base uses google style docs, e.g. `Args` instead of `Paramters`.

src/pynxtools/dataconverter/validation.py
RubelMozumder: I meant to check the `key-value` pair from mapping, and if `value` is none.

src/pynxtools/dataconverter/validation.py
lukaspie: Yup, changed

RubelMozumder (approved):

src/pynxtools/dataconverter/validation.py
lukaspie: Found a solution, we check if each of these conflicting keys is actually valid. If any is not valid, we remove it instantly. If after this check there are still more than two conflicting keys remaining, we remove these, too.

rettigl (approved):
This looks good to me, however there is at least one case this does not cover:
If I define the same concept twice, once with a variadic name, once without (for a named concept), the writer will fail, because the entry is already written:
Example: 
`/ENTRY/INSTRUMENT/energy_resolution/type`
and 
`/ENTRY/INSTRUMENT/RESOLUTION[energy_resolution]/type`
both defined.
I think, however, that we don't have to catch such cases, as this is clearly a broken input. I think we probably can never and don't have to make this completly failsafe.

rettigl:
The identification of valid or invalid concepts don't seem to fully work yet. If I have two similar groups with sub-fields:
```
  "/ENTRY/USER[user]": {
    "name": ,
    "role": ",
    "affiliation": ,
    "address": ,
    "email": 
  },
  "/ENTRY/ILLEGAL[user]": {
    "name": ,
    "role": ",
    "affiliation": ,
    "address": ,
    "email": 
  },
```
all keys are removed:
```
WARNING: Instance name 'user' used for multiple different concepts: ILLEGAL, USER. The following keys are affected: /ENTRY[entry]/ILLEGAL[user]/address, /ENTRY[entry]/ILLEGAL[user]/affiliation, /ENTRY[entry]/ILLEGAL[user]/email, /ENTRY[entry]/ILLEGAL[user]/name, /ENTRY[entry]/ILLEGAL[user]/role, /ENTRY[entry]/USER[user]/address, /ENTRY[entry]/USER[user]/affiliation, /ENTRY[entry]/USER[user]/email, /ENTRY[entry]/USER[user]/name, /ENTRY[entry]/USER[user]/role.
WARNING: The key /ENTRY[entry]/ILLEGAL[user]/address will not be written.
WARNING: The key /ENTRY[entry]/ILLEGAL[user]/affiliation will not be written.
WARNING: The key /ENTRY[entry]/ILLEGAL[user]/email will not be written.
WARNING: The key /ENTRY[entry]/ILLEGAL[user]/name will not be written.
WARNING: The key /ENTRY[entry]/ILLEGAL[user]/role will not be written.
WARNING: The key /ENTRY[entry]/USER[user]/address will not be written.
WARNING: The key /ENTRY[entry]/USER[user]/affiliation will not be written.
WARNING: The key /ENTRY[entry]/USER[user]/email will not be written.
WARNING: The key /ENTRY[entry]/USER[user]/name will not be written.
WARNING: The key /ENTRY[entry]/USER[user]/role will not be written.
```
even though the second one is invalid.

lukaspie:
> This looks good to me, however there is at least one case this does not cover: If I define the same concept twice, once with a variadic name, once without (for a named concept), the writer will fail, because the entry is already written: Example: `/ENTRY/INSTRUMENT/energy_resolution/type` and `/ENTRY/INSTRUMENT/RESOLUTION[energy_resolution]/type` both defined. I think, however, that we don't have to catch such cases, as this is clearly a broken input. I think we probably can never and don't have to make this completly failsafe.

I implemented a check (+test) for this now and if the non-concept key is valid, we remove the one with the concept. That is, in your example, we remove `/ENTRY/INSTRUMENT/RESOLUTION[energy_resolution]/type` and keep `/ENTRY/INSTRUMENT/energy_resolution/type`.

> The identification of valid or invalid concepts don't seem to fully work yet. If I have two similar groups with several sub-fields [...] If there is just one sub-key in the valid entry, it works as expected, which is also what the test contains. I suggest to fix this and extend the test by a second sub-key. Maybe also break the test apart into several ones that test different aspects of this functionality.

This should be fixed now. I also split up the tests to cover the different cases individually.

rettigl (approved):
LGTM

PR #656, opened 2025-06-16
drop support for python 3.8

lukaspie:


mkuehbach (approved):
Good, yeah py3.8 passed end-of-life
https://www.python.org/downloads/

We should add py3.13 to the test matrix though

**REBASE against latest master REQUIRED before merge!**

PR #657, opened 2025-06-18
update ruff

lukaspie:
- Use the same ruff rules as NOMAD
- Exception: ```"PLC0415", # `import` should be at the top-level of a file``` -> we ignore this for now since this would require too many changes and prevents lazy loading, especially in the NOMAD entrypoints (this feature is still disucssed anyway, see https://github.com/astral-sh/ruff/issues/18728)
- Most important change: implement `"I001"`, isort-like ordering of imports


mkuehbach (approved):
Reasonable changes

PR #660, opened 2025-06-30
add units files to MANIFEST.in

lukaspie:
This is needed to ship the two units .txt files when building the package.


mkuehbach (approved):
lgtm but check that CI runs through

PR #667, opened 2025-07-18
EM, APM, CG, MS database reprocessing

mkuehbach:
- [ ] Currently, test_nexus.py for the WSe example fails > as definitions have changed a lot that reference test was generated years ago, this test should be updated to work with a file that is not necessarily domain-specific and that can be regenerated with clear provenance of the test data, also local file paths shared in the test which should not be done


mkuehbach:
This was a branch and PR generated for implementing a feature branch in pynxtools-em when the definitions where not yet voted upon by the NIAC. In the meantime points that changed in pynxtools addressed --- specifically https://github.com/FAIRmat-NFDI/pynxtools/pull/688 updates required for definitions so this PR can be closed, the `em_apm_database feature branch though needs to be kept until v0.4.0 of pynxtools-em has been release on pypi.

mkuehbach:
Closed, draft, used for a demo, local backup by @mkuehbach exists

PR #669, opened 2025-07-21
Resolve "Validation of open enumerations does not search for @custom attribute"

### LINKED ISSUE #666: [Bug]: Validation of open enumerations does not search for `@custom` attribute
### Contact Details

_No response_

### What happened?

When open enumerations were introduced in https://github.com/nexusformat/definitions/pull/1521, it was decided that 
- for fields, an attribute ``@custom=True``.
- * for attributes, set an additional attribute ``@my_attribute_custom=True`` (where ``my_attribute`` is the name of the attribute with the open enumeration)
shall be added to the file. In that case, only an info shall be logged, whereas if these attributes are missing, a warning is expected. We never actually implemented that in the validation.

### Relevant log output

```shell

```

lukaspie:
Closes #666

Implementation is the following:
- Logic for closed enums remains untouched.
- For open enumerations where a value is used that does not match any of the enumeration items, the validator searches for `@custom` attribute (or `@my_attribute_custom` attribute for a a given attribute `my_attribute`).
-  If that attribute is correctly set to True, we just give an info message.
- An warning level message is given if `@custom` is set to False.
- For HDF5 validation: If the `@custom` attribute is missing, a warning is logged.
- For template validation: If the `@custom` attribute is missing, we add it automatically to the template. As the produced file is then correct, we only log an info message (incl. the information that the attribute was added automatically).

Further changes:
- In the testing framework, I extended the logic for ignoring lines within specific sections in the log comparison. Now the person writing the test can just set `ignore_sections = {"FIELD (//Ag__002__VB/start_time)": ["DEBUG - value:"]}` and this will also ignore the `"DEBUG - value:"` line in all sections that start with `"FIELD (//Ag__002__VB/start_time)`. We need this as for fields the line is actually `DEBUG - ===== FIELD (//Ag__002__VB/start_time): <HDF5 dataset "start_time": shape (), type "|O">` and I don't want people having to give the whole line including the dataset information.

### DIFF
--- a/.github/workflows/plugin_test.yaml
+++ b/.github/workflows/plugin_test.yaml
@@ -42,7 +42,7 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-xps
-            branch: custom-attr-for-open-enum
+            branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-xrd
             branch: main
--- a/.github/workflows/plugin_test.yaml
+++ b/.github/workflows/plugin_test.yaml
@@ -42,7 +42,7 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-xps
-            branch: main
+            branch: custom-attr-for-open-enum
             tests_to_run: tests/.
           - plugin: pynxtools-xrd
             branch: main
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -154,7 +154,7 @@ class Collector:
 
         elif log_type == ValidationProblem.InvalidEnum:
             logger.warning(
-                f"The value '{args[0]}' at {path} should be one of the following: {value}."
+                f"The value {args[0]} at {path} should be one of the following: {value}."
             )
         elif log_type == ValidationProblem.OpenEnumWithCustom:
             logger.info(
@@ -162,7 +162,7 @@ class Collector:
             )
         elif log_type == ValidationProblem.OpenEnumWithCustomFalse:
             logger.warning(
-                f"The value '{args[0]}' at {path} does not match with the enumerated items from the open enumeration: {value}."
+                f"The value '{args[0]}' at {path} does not match with the enumerated items from the open enumeration: {value}. "
                 "When a different value is used, the boolean 'custom' attribute cannot be False."
             )
         elif log_type == ValidationProblem.OpenEnumWithMissingCustom:
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -172,7 +172,9 @@ class Collector:
             )
             if args[1] is True:
                 log_text += " It was added here automatically."
-            logger.info(log_text)
+                logger.info(log_text)
+            else:
+                logger.warning(log_text)
         elif log_type == ValidationProblem.MissingRequiredGroup:
             logger.warning(f"The required group {path} hasn't been supplied.")
         elif log_type == ValidationProblem.MissingRequiredField:
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -87,8 +87,9 @@ class ValidationProblem(Enum):
     UnitWithoutDocumentation = auto()
     InvalidUnit = auto()
     InvalidEnum = auto()
-    OpenEnumWithCorrectNewItem = auto()
-    OpenEnumWithIncorrectNewItem = auto()
+    OpenEnumWithCustom = auto()
+    OpenEnumWithCustomFalse = auto()
+    OpenEnumWithMissingCustom = auto()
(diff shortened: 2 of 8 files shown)


.github/workflows/plugin_test.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
             branch: main
             tests_to_run: tests/.
           - plugin: pynxtools-xps
-            branch: main
+            branch: custom-attr-for-open-enum
```
lukaspie, line 45: reminder to revert

src/pynxtools/dataconverter/helpers.py
```diff
@@ -878,18 +872,120 @@ def validate_data_value(
...
+                    custom_attr = mapping.get(custom_path)
+                    custom_added_auto = True
+
+                if custom_attr == True:  # noqa: E712
+                    collector.collect_and_log(
+                        path,
+                        ValidationProblem.OpenEnumWithCustom,
+                        nxdl_enum,
+                        value,
+                    )
```
RubelMozumder, line 963: Not sure what it does. `enum_value` is in `nxdl_open_enum` and `custom_attr` is true but still writing a warning message.

src/pynxtools/dataconverter/validation.py
```diff
@@ -1410,10 +1434,16 @@ def handle_attribute(node: NexusNode, keys: Mapping[str, Any], prev_path: str):
...
+                variant_path,
+            )
+            is_valid_enum(
+                mapping[variant_path],
                 node.items,
                 node.open_enum,
                 variant_path,
+                mapping,
             )
+            remove_from_not_visited(get_custom_attr_path(variant_path))
```
RubelMozumder, line 1446: It needs to be called outside of the for loop. Otherwise, Logger will be repeating the waning messages for all variants except the variant that contains the enum value.

src/pynxtools/testing/nexus_conversion.py
```diff
@@ -241,17 +241,25 @@ def load_logs(
...
             """Compare log lines, ignoring specific differences."""
 
+            def get_section_ignore_lines(line: str) -> list[str]:
+                """Return ignore lines for a section if the line starts with the section."""
+                section = line.rsplit(SECTION_SEPARATOR, 1)[-1].strip()
+                for key, ignore_lines in IGNORE_SECTIONS.items():
+                    if section.startswith(key):
+                        return ignore_lines
+
+                return []
```
RubelMozumder, line 251: Can you add a simple example about `ignore_line` and `ignore_section` in the testing framework docs? I had a hard time using them.

tests/dataconverter/test_validation.py
```diff
@@ -867,24 +867,114 @@ def format_error_message(msg: str) -> str:
...
+                            "a very different type",
+                        ),
+                        "/ENTRY[my_entry]/NXODD_name[nxodd_name]/type2/@custom",
+                        True,
+                    ),
+                    "/ENTRY[my_entry]/NXODD_name[nxodd_name]/type2/@attribute_with_open_enum",
+                    "3rd option",
+                ),
+                "/ENTRY[my_entry]/NXODD_name[nxodd_name]/type2/@attribute_with_open_enum_custom",
+                True,
```
RubelMozumder, line 892: `attribute_with_open_enum_` it should be an attribute name as described in PR description, in this case `3rd option`?

tests/dataconverter/test_validation.py
```diff
@@ -867,24 +867,114 @@ def format_error_message(msg: str) -> str:
...
             id="open-enum-with-new-item",
         ),
+        pytest.param(
+            alter_dict(
+                alter_dict(
+                    alter_dict(
+                        alter_dict(
+                            TEMPLATE,
+                            "/ENTRY[my_entry]/NXODD_name[nxodd_name]/type2",
+                            {"compress": "a very different type", "strength": 2},
```
RubelMozumder, line 909: It is not related to the change in this PR. What are the benefits if `compress` only contains a single string?

src/pynxtools/dataconverter/validation.py
lukaspie: I don't understand. We do want to emit warnings for any variant. If I have a field call `my_field_ID` that has an enumeration, I want to check the validity of the enumeration for all variants, e.g. `my_field_1`, and `my_field_2`. If we call outside the for loop, only `my_field_2` would be checked.

tests/dataconverter/test_validation.py
lukaspie: It is related to this PR as type2 is an open enum and we want to check for the custom attribute. Specfically, we want to check that compressed items get unpacked and compared to the enumeration properly.

Indeed, having a string as the payload does not make too much sense for compression. But I didn't want to add another enumerated concept of another type to NXtest. But the logic is the same for strings as for e.g. large arrays.

RubelMozumder (approved):
LGTM! I explained a resolved comment further. You can take care of that if it makes sense to you.

PR #673, opened 2025-07-28
update python in CI/CD and scripts

lukaspie:
This also updates `urllib3` to v2.5 (you can do this by addding `--upgrade-package urllib3` to the script to generate the dependencies).


mkuehbach (approved):
Lgtm

PR #682, opened 2025-07-31
use newer version of hdfdict, implement tests

lukaspie:
@sherjeelshabih there is a newer version of hdfdict available at https://github.com/SiggiGue/hdfdict/blob/master/hdfdict/hdfdict.py. I brought it here and implemented some tests for it. Can you check that this still works for your reader?


sherjeelshabih (approved):
I tested this with the json_map reader and it works.

Issue #684, opened 2025-08-01
[Bug]: #675 eventually drops annotations to enforce compression of datasets


mkuehbach:
### What happened?

A bug happened!

### Relevant log output

```shell

```

PR #686, opened 2025-08-04
implement more rigorous checks of compressed payloads

lukaspie:
As discussed in #684, we currently have a bug that compressed values for the sake of type checking get unpacked, but they are also passed to the writer unpacked (entirely removing the compression). We remove that check in #685, but then again, we do not check that the compressed value is actually valid.

This PR applies the same checking logic that we had for native types also do dicts like `{"compress": value, strength: 1}`, using the same function to check in both cases.

### DIFF
--- a/tests/dataconverter/test_validation.py
+++ b/tests/dataconverter/test_validation.py
@@ -53,6 +53,7 @@ def set_whole_group_to_none(
 
 
 def compress_paths_in_dict(data_dict: Template, paths=list[str]):
+    """For each path, compress the value in data_dict using a strength of 3."""
     types = {
         "int": np.int64,
         "float": np.float32,


src/pynxtools/dataconverter/helpers.py
```diff
@@ -200,6 +209,25 @@ def _log(self, path: str, log_type: ValidationProblem, value: Optional[Any], *ar
             logger.warning(
                 f"The key '{path}' uses the valid concept name '{args[0]}', but there is another valid key {value} that uses the non-variadic name of the node.'"
             )
+        elif log_type == ValidationProblem.InvalidCompressionStrength:
+            logger.warning(
+                f"Compression strength for {path} = {value} should be between 0 and 9."
+            )
+        elif log_type == ValidationProblem.DoNotCompressEnum:
+            logger.warning(
+                f"Compression for {path} = {value} should not be used for enumerated concepts."
```
lukaspie, line 218: @mkuehbach can you explain your reasoning why we should not compress these values if the node's values are enumerated?

src/pynxtools/dataconverter/helpers.py
```diff
@@ -723,78 +752,91 @@ def convert_int_to_float(value):
...
-        results = iso8601.search(value)
-        if results is None:
-            collector.collect_and_log(path, ValidationProblem.InvalidDatetime, value)
+    if isinstance(value, dict) and set(value.keys()) == {"compress", "strength"}:
+        compressed_value = value["compress"]
 
-    # Check enumeration
-    if nxdl_enum is not None and value not in nxdl_enum:
-        if nxdl_enum_open:
+        if not (0 <= value["strength"] <= value["strength"] <= 9):
```
mkuehbach, line 815:
```suggestion
        if not (0 <= value["strength"] <= 9):
```

src/pynxtools/dataconverter/helpers.py
```diff
@@ -723,78 +752,91 @@ def convert_int_to_float(value):
...
+        # Do not count other dicts as they represent a link value
+        if not isinstance(value, dict):
+            # Attempt type conversion
+            if accepted_types[0] is bool and isinstance(value, str):
+                try:
+                    value = convert_str_to_bool_safe(value)
+                except (ValueError, TypeError):
+                    value = original_value
+            elif accepted_types[0] is float:
+                value = convert_int_to_float(value)
```
mkuehbach, line 782: Please add a warning between 781 and 782 that this conversion can be result in precision loss but keep line 782 for now, as mentioned in another comment, yeah keep it here and we should at some point make a follow-up PR which removes this, main reason, avoiding side effects now for other users that rely on this conversion.

tests/dataconverter/test_validation.py
```diff
@@ -1694,27 +1711,64 @@ def listify_template(data_dict: Template):
...
+            [
+                "The value at /ENTRY[my_entry]/NXODD_name[nxodd_name]/int_value "
+                "should be one of the following Python types: "
+                "(<class 'int'>, <class 'numpy.integer'>), as defined in the "
+                "NXDL as NX_INT."
+            ],
+            id="appdef-compressed-wrong-type",
         ),
         pytest.param(
             alter_dict(
```
mkuehbach, line 1758: Adding one test probing strength = 0 is useful to test the aforementioned warning branch in cases of expectable low compression ratio

src/pynxtools/dataconverter/helpers.py
mkuehbach: > They also work within H5Web, I just checked.

So are these displayed decompressed in H5Web for end users?
If this is the case, I agree with that there should no warnings should be omitted
when strings or enums are passed with compression desired.

Wrt to the return value, we are also on the same page, another issue, another pr.

src/pynxtools/dataconverter/helpers.py
mkuehbach: Consistent with my wish to limit the scope of this PR and move on this is a reasonable compromise.
We keep the conversion int to float as it is for now and should address this is in a follow-up.

My view on the is-info-or-warning is: When not every int has exactly only one equivalent float it needs a warning cuz we cannot foresee if there are really no applications that can ignore the precision loss.

lukaspie:
@mkuehbach should be ready to be merged. Let's make a new release afterwards, see #688

mkuehbach (approved):
All changes were addressed professionally.

PR #699, opened 2025-09-13
Hook the CleanSpm Branch.

RubelMozumder:
Update the nexus definition.


src/pynxtools/nomad/parser.py
```diff
@@ -142,7 +143,7 @@ def _get_value(hdf_node):
     if hdf_value.dtype.kind in "iufc":
         return hdf_value
     if len(hdf_value.shape) > 0:
-        return str([i for i in hdf_value.astype(str)])
+        return str([str(decode_or_not(i)) for i in hdf_value])
```
mkuehbach, line 146: Please substantiate the motivation for this change.
At that point in the data parsing pipeline one should already have proper HDF5 strings
and that means e.g. variable length null-terminated utf8, if a change is necessary here for what worked for before for more than four application definitions it should rather be pynxtools-{stm,sts,spm} parsers that normalize the strings correctly.

mkuehbach (changes requested):
See conversations

src/pynxtools/nomad/parser.py
RubelMozumder: It was not intended to be reviewed. 
Actually this is not a problem from spm. The data coming through the `yaml` file and read with `utf-8` encoding. But while writing into the HDF5 file, the strings are stored as raw string of variable length of bytes. If the raw string consists only english alphabet, Python str type can handle the string, but letters from other languages (e.g., Dr. Cojal González (HU) comes in byte) are problematic which parsing the nexus file in Nomad. It needs to be decoded. 
This situation is likely to arise in the use of the installation from different source speakers.  
Basically, the code checks if any `b` strings are in the dataset and converts them according to `utf-8` encoding; other than that, it does do the same thing as the previous code did.

lukaspie: Can we add a test file for this problem and check that is being parsed properly in `tests/nomad/test_parser.py`?

src/pynxtools/nomad/parser.py
```diff
@@ -142,7 +143,15 @@ def _get_value(hdf_node):
     if hdf_value.dtype.kind in "iufc":
         return hdf_value
     if len(hdf_value.shape) > 0:
-        return str([i for i in hdf_value.astype(str)])
+
+        def recurse(x):
```
mkuehbach, line 147: Why to replace a forward loop with recursion when the shape is known apriori?

src/pynxtools/nomad/parser.py
RubelMozumder: We mainly collect all user provided data through the `yaml` file. In this PR description, `Dr. Cojal González (HU)` is the name of the user. There might come such issues for name of institutions, especially when publishing the data. The encoding problem is mainly coming for name string things. For other numerical data, there is no such issue. Because, issues related with numerical data come while wrtting data to a nexus file, and such issues should be handled in plugin level.

Mainly, I sse one thing is missing to include the test here.

src/pynxtools/nomad/parser.py
RubelMozumder: > Why to replace a forward loop with recursion when the shape is known apriori?

No, we do not know the exact shape.

lukaspie:
> > > * [x]    `structlog` should be part of the core dependencies as it is used in pynxtools test
> > 
> > 
> > As you said, it is only used in the tests. Given that the tests require a `dev` installation to run (`pytest` is also in `optional_dependencies.dev`), we don't need to move `structlog` to the core deps. It can stay in `dev` just fine.
> 
> Yeah, like you, it also surprised me seeing that `structlog` is not in the main dependency. But it is used [here](https://github.com/FAIRmat-NFDI/pynxtools/blob/master/src/pynxtools/testing/nexus_conversion.py?rgh-link-date=2025-10-01T09%3A15%3A43Z#L25)

That is different then. It is not used in the `tests`, it is used in the testing module which is part of the shipped package. Then it is fine to add it to the core deps.

src/pynxtools/nomad/parser.py
```diff
@@ -142,7 +143,15 @@ def _get_value(hdf_node):
...
+            if hasattr(x, "shape") and len(x.shape) > 0:
+                return [recurse(i) for i in x]
+            else:
+                return str(decode_or_not(x))
+
+        return str(recurse(hdf_value))
```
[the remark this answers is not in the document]
RubelMozumder: In the suggested code 
```py
 if hdf_value.dtype.kind == "S":
      return [x.decode("utf-8") for x in hdf_value]

```
It will not handle a multidimensional dataset and therefore may be problematic. We can remove this part as the last return statement
```python
hdf_node[()].decode()
```
will take care of this situation. 
Is it ok for you?

lukaspie (approved):
LGTM, provided the following is addressed:
- remaining comment on `parser.py`, as agreed
- change in `plugin_tests` get reverted
- rebase? for `nomad_compatibility` test

Eventually, when the definitions branch is merged, we should go back to the latest commit on the `fairmat` branch.

PR #704, opened 2025-09-29
Modifications for nomad meta-info underlining function change

RubelMozumder:


lukaspie:
Can you please elaborate why this change is needed and what it is doing? Hard to understand otherwise.

RubelMozumder:
> > With the new implementation nomad meta-info
> 
> Please provide context e.g. nomad-FAIR or gh commit where this was introduced by
> 
> > > CC @TLCFEM 

> > the annotation is expected
> 
> Which specific annotation?

RubelMozumder:
> > Can you please elaborate on why this change is needed and what it is doing? Hard to understand otherwise.
> 
> Currently, `m_annotation.get` or `m_annotation_get` returns a list, not an object. But, in both cases, we define a default unit which is `None`. These changes are needed due to some modifications injected by @TLCFEM. He can better give context on modification, what type of modifications he and the corresponding GH commit.

Alternatively, we can use also 
```
default_display_unit = quantity_def.m_annotations.get(
  eln", [None] 
  0].defaultDisplayUnit
```

TLCFEM (approved):

PR #706, opened 2025-10-01
allow prereleases in nomad compatibility check

lukaspie:


lukaspie:
> LGTM! But XRD is failing. I have not seen such an attribute `unit_custom` in XRD. Likely, this is coming from pynxtools modification.

Oh yeah, this is because of https://github.com/FAIRmat-NFDI/pynxtools/pull/669. I changed the XRD test files already (see https://github.com/FAIRmat-NFDI/pynxtools-xrd/commit/2c92768ce8ae5a9840aeb139e1c1c226ced8326b) so the tests would pass, but did want to bother you about it. We can just merge here and then the tests will work again in #669.

PR #714, opened 2025-10-16
Fixes Nomad example names to pythonic names

sherjeelshabih:


mkuehbach (approved):
lgtm, this should be taken as the blueprint on how to refactor the eln archive for each pynxtools plugin

mkuehbach (dismissed):
Tests dont run through

mkuehbach (approved):
Sherjeel Shabih pynx 714 only has failing tests for those plugins that need the fix provided by this 714 to work, so I think merging 714 is fine

PR #715, opened 2025-10-16
bump ruff, fix cython error

lukaspie:


mkuehbach:
> Bumping ruff is _helpful_ here because it actually catches these issue in v0.14 🤷🏻‍♂️

Good, next time noteworthy to mention in the PR's executive summary at the top

PR #718, opened 2025-10-17
Prepare the v0.12.0 release

mkuehbach:
Noteworthy, spellchecking is now performed before the mkdocs deployment.


mkuehbach (commented):
Next releases should critically inspect whether low-bound version constraints on the plugins are required.

Assume hypothetical tool `pynxtools-dummy` has releases 0.1, 0.2, 0.3 on pypi.
Assume further that pynxtools' pyproject.toml in the convert targets defines `dummy` with a lower-bound constrained `pynxtools-dummy>=0.2`

This will install the latest, i.e., 0.3, unless 0.3 and 0.2 get yanked later and hence there is no longer a compatible version.

Here is a reversion of responsibilities:
It should be the plugin developers responsibility to assure that their plugins work with a specific pynxtools version.
If they want to have their plugin included in one of the official targets offered by pynxtools they should communicate
if there are possible downwards compatibility issues. If that does not happen, we should always keep the lower bound constraint the latest version.

Wont break tests as these use main/master branches and not releases.

Maybe at some point these low-bound constraints were not updated when releases of pynxtools were made in the past.

Issue #583, opened 2025-03-12
Update `chemical_formula` normalization


lukaspie:
Following https://github.com/nexusformat/definitions/pull/1413 and the non-acceptance of `NXsample_component_set`, we'll have to adjust the normalizer acccordingly.

mkuehbach:
There is still a bug with parsing the chemical formula despite using a correct NeXus/HDF5 file:
https://nomad-lab.eu/prod/v1/gui/search/nexusapp/entry/id/7SjDXd9i7ts8vN84qVZpbfnT6wnM

Should parse from
```
/entry1/specimen/@NX_class = "NXsample"
/entry1/specimen/atom_types = "C, Cr, Cu, O, Si"
```
Exists in the HDF5 file at said location and is formatted correctly, still though no atom_types thus no material

https://nomad-lab.eu/prod/v1/gui/search/nexusapp/entry/id/7SjDXd9i7ts8vN84qVZpbfnT6wnM

PR #723, opened 2025-11-12
fix `atom_types` parsing

lukaspie:
This fixes the additional issue described in https://github.com/FAIRmat-NFDI/pynxtools/issues/583#issuecomment-3492130517. The problem was that 
```
(atom_types := sample.get("atom_types__field"))
```
does not actually return the value of `sample.atom_types__field`, but just the string `atom_types__field`. So we need to use the workaround presented here.

Local script for checking (using `nomad==1.4` and `pynxtools` from this branch):
```python
from nomad.datamodel import EntryArchive
from nomad.utils import get_logger
from pynxtools.nomad.parser import NexusParser

archive = EntryArchive()
NexusParser().parse(str("output.nxs"), archive, get_logger(__name__))

assert archive.data.ENTRY[0].specimen.atom_types__field == "C, Cr, Cu, O, Si"
assert archive.results.material.elements == ["C", "Cr", "Cu", "O", "Si"]
```

Here's the result section when running the GUI and appworker locally using the versions above:

image

@mkuehbach please test locally.

### DIFF
--- a/src/pynxtools/nomad/parser.py
+++ b/src/pynxtools/nomad/parser.py
@@ -478,6 +478,10 @@ class NexusParser(MatchingParser):
                     for symbol in atom_types.replace(" ", "").split(","):
                         if symbol in chemical_symbols[1:]:
                             element_set.add(symbol)
+                        else:
+                            self._logger.warn(
+                                f"Ignoring {symbol} as it is not for an element from the periodic table"
+                            )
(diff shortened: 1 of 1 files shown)


mkuehbach (approved):
Fix makes sense.

(in the original version `atom_types` was always instantiated irrespective whether the if branch is taken --- eventually to such that atom_types ends up as None)

PR #724, opened 2025-11-21
store json file in cache dir for local build

lukaspie:
Fixes NeXus metainfo JSON generation and packaging

Issue: The `nxs_metainfo_package_<version>.json` file was always written inside the package directory, causing `PermissionError` in read-only environments and preventing git installs from working properly. This led to an issue in the `test-oasis` deployment where we still install `pynxtools` from the GitHub `main` branch:

```
PermissionError: [Errno 13] Permission denied: '/opt/venv/lib/python3.12/site-packages/pynxtools/nomad/nxs_metainfo_package_v2024.02.post1.dev1999+gbce25499.json'

```

Fix:

- `get_package_filepath()` now uses a dual-path approach: packaged file if it exists, otherwise a user cache.
- Build mode (env variable `PYNXTOOLS_BUILD_PACKAGE=1`) writes the JSON into the package so it’s included in the wheel.
- Runtime installs (git/editable) generate the JSON safely in cache.

### DIFF
--- a/src/pynxtools/nomad/utils.py
+++ b/src/pynxtools/nomad/utils.py
@@ -141,26 +141,18 @@ CACHE_DIR.mkdir(parents=True, exist_ok=True)
 
 
 def get_package_filepath() -> Path:
-    """Return the filesystem path for the NeXus metainfo package JSON file.
+    """Return the path to the NeXus metainfo package JSON file.
 
-    The location is determined in the following order:
+    Resolution order:
 
-    1. **Build-mode override:**
-       If the environment variable ``PYNXTOOLS_BUILD_PACKAGE`` is set to ``"1"``,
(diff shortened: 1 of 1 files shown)


src/pynxtools/nomad/utils.py
```diff
@@ -133,8 +135,43 @@ def get_quantity_base_name(quantity_name):
...
+       time the file is generated.
+
+    Returns:
+        Path: The resolved path where the NeXus metainfo package JSON file
+        should be read from or written to.
+    """
     filename = f"nxs_metainfo_package_{get_nexus_version()}.json"
 
-    return os.path.join(local_dir, filename)
+    # 1. Build-mode override (forces packaging the file)
```
mkuehbach: The code following is so clean, would remove condense the three cases into the three respective in-function comments per case.

PR #747, opened 2026-03-17
Optimization related to compression, allowing multithreaded blosc2 to be used as a compression library

mkuehbach:
- [x] Adding whats required to use blosc2, essentially making explicit what the inclusion of hdf5plugin and pandas already delivered
- [x] **Write all non-scalar non-string datasets using chunked storage layout with h5py's autochunker active by default** the main motivation behind this move is that it allows everybody to take advantage of #750  and #752 chunked-by-chunk based processing
- [x] deflate, i.e. "gzip" is kept as the standard algo by default.


src/pynxtools/dataconverter/chunk.py
```diff
@@ -26,11 +27,29 @@
...
 
-COMPRESSION_FILTERS: list[str] = ["gzip"]  # deflate
+PYNX_ENABLE_BLOSC: bool = False  # deactivated by default
+# use only when it is acceptable to work with blosc2-compressed content downstreams
+# mind that doing so in C/C++, Matlab, and Fortran application requires specific
+# linking of these apps with a customized HDF5 library that links to the blosc library
+# consider that using blosc sets explicit a certain number of cores eligible for
+# doing compression and decompression work that may drain resources when pynxtools
+# is used in conjunction with other apps and services like NOMAD
+# check the set_nthreads in writer.py to modify accordingly for your best practice
```
sherjeelshabih: Too much of an in code comment. Let's simplify and use some defaults or add this to a more accessible README.md section/link to another .md and/or to the docs.

sherjeelshabih (dismissed):

src/pynxtools/dataconverter/writer.py
```diff
@@ -36,7 +38,28 @@
...
+    COMPRESSION_FILTERS,
+    COMPRESSION_STRENGTH,
+    PYNX_ENABLE_BLOSC,
+)
+
+if (
+    PYNX_ENABLE_BLOSC
+    and importlib.util.find_spec("hdf5plugin") is not None
+    and importlib.util.find_spec("blosc2") is not None
+):
```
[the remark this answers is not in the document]
mkuehbach: I removed the importlib calls as we also do not make such for other mandatory dependencies both now mandatory dependencies, now explicit in pyproject toml while previously this was only implicit as a transitive of pytables and pandas

src/pynxtools/dataconverter/writer.py
```diff
@@ -164,16 +193,22 @@ def handle_dicts_entries(data, grp, entry_name, output_path, path, append):
...
 
             if entry_name not in grp:
                 try:
+                    if compression_filter == "gzip":
+                        compression_config = dict(
+                            compression=compression_filter,
+                            compression_opts=compression_strength,
+                        )
+                    else:  # by virtue of construction blosc
+                        compression_config = hdf5plugin.Blosc2(cname="zstd", clevel=9)
```
RubelMozumder: `gzip` is the default; shouldn't it be under the else clause? Other requested compression would come in condition with a check it library is available in the environment, along with other conditions.

Though compression strength comes from `data` dict, for `blosc2` it is hardcoded to 9. By default, for blosc compression falls to strength 9, in case the user does not supply it.

docs/learn/pynxtools/compression.md
```diff
@@ -102,3 +102,28 @@ performance compromise that is tailored towards single storage operations like o
...
+
+Consequently, the feature of `pynxtools` to support also filters from the [`blosc2`](https://blosc.org/)
+project and compression library via the [`hdf5plugin`](https://pypi.org/project/hdf5plugin/) is an offering to our users with
+performance critical processing demands. We have configured `blosc` such that it uses the `zstd` compression algorithm.
+This custom compression filter has a multithreaded and vectorized implementation, two key optimizations over `deflate` that make it an fast alternative
+for users with demands for processing large data volume. By default, the usage of `blosc` is switched off as `deflate` is the default in `pynxtools`.
+
+Users should be aware though that custom compression filters, like `blosc2` are not typically included in the default compilation pipelines of HDF5.
```
lukaspie:
```suggestion
Users should be aware though that custom compression filters, like `blosc2`, are not typically included in the default compilation pipelines of HDF5.
```

lukaspie (approved):
Looks good, but I am not an expert

src/pynxtools/dataconverter/writer.py
```diff
@@ -152,10 +177,14 @@ def handle_dicts_entries(data, grp, entry_name, output_path, path, append):
...
     elif "compress" in data.keys():
         if not (isinstance(data["compress"], str) or np.isscalar(data["compress"])):
-            if ("filter" in data.keys()) and (data["filter"] in COMPRESSION_FILTERS):
-                compression_filter = data["filter"]
+            if (
+                PYNX_ENABLE_BLOSC_NTHREADS > 0
+                and "filter" in data.keys()
+                and data["filter"] == "blosc"
+            ):
+                compression_filter = "blosc"
```
RubelMozumder, line 167: At the very beginning, an attempt to import blosc2 fails the filter falls back to `zgip`. But, later in this file, it is always assumed that `blosc2` is imported and invoked where the user requested via `data` dict. So, in all cases it should be checked if blosc2 is available; otherwise, it falls back to the default `zgip`.

RubelMozumder (approved):
LGTM!
Though I have a general question, how effectively will this multi-threading work in NOMAD? Where NOMAD launches `NeXus` parser using an asynchronous thread. I do not have any idea about it.

docs/learn/pynxtools/compression.md
mkuehbach: Fixed

src/pynxtools/dataconverter/writer.py
mkuehbach: "gzip" is expected to be the most frequently used execution path, if else clauses should, unless it makes the readability super hard, put the more likely path in the if clause to profit from speculative execution.
Compression levels for gzip and blosc2 are not directly comparable, comment added in `chunk.py` on this

src/pynxtools/dataconverter/writer.py
mkuehbach: Modified

PR #749, opened 2026-03-17
Get correct append mode

RubelMozumder:
This PR fixes a bug: writes always append=True, if append comes in `kwargs`.

And uses the same syntax to get `append` from `kwargs.`


mkuehbach (approved):
Good catch, please merge

Issue #737, opened 2026-02-11
Tame the eventual memory hungryness of the nomad parser for large datasets


mkuehbach:
- [ ] Observed for 15.5 GiB NeXus/HDF5 NXem dataset, essentially all content in one 3d array which uncompressed is 32GiB, the array if stored chunked, upon parsing multiple times the uncompressed payload is allocated causing the parsing to break on small systems.

<img width="1806" height="915" alt="Image" src="https://github.com/user-attachments/assets/c326cd5d-f343-4401-a2c8-905d88b097f1" />

Suggestions, we should check for inefficiencies, assuming what happens when a dataset is large:
- [ ] Compute summary statistics and finiteness via taking advantage of chunked layout, chunk-by-chunk
- [ ] Could be useful to drop processing statistics for contiguously stored datasets if larger than fraction of maximum allocated memory accessible to the user in the deployment, like "drop of load"
- [ ] Access h5 objects metadata directly rather than via np.shape() calls

mkuehbach:
- [x] Should add guidance to users via docs, like "Tips for working with large datasets"

mkuehbach:
Implemented in #750 and #752 both ready for review

lukaspie:
completed, thanks for the fix

PR #750, opened 2026-03-17
Memory optimization NOMAD nexus parser

mkuehbach:
Planned for `v0.14.0`

Addressing one aspect to issue #737:

**Key changes for non-scalar iuf HDF5 datasets (integer, unsigned, float):**
- [x] Using float64 precision throughout for computation of the naive arithmetic mean
- [x] Dropping the standard deviation field cuz i) it will for chunked storage layout require a dual pass through the data, ii) several area B folks reported that this was never really used
- [x] Taking advantage of benefits available when datasets are stored using the chunked layout:
       Loading chunk by chunk then datasets completely, thus completely flattening memory spikes
- [x] Added example how to compute numerical more precise (non-naive) arithmetic mean and std (Welford algorithm) for chunked storage but deactivated it by default as it would have a substantial performance hit.
- [x] Tested with 10GB mock NXem dataset, previously contiguous storage layout, memory allocation for loading the dataset during parsing to get mean spiked at approx 40 - 45 GB,
       using chunked storage layout (irrespective with compression or not (both supported now)), memory flat, only a few MB kept while iterating over chunks. 
- [x] Additional benefit, with this running 2.5x faster than without if using no compression, with compression there is overhead.
- [x] Aim achieved, no memory spikes, out-of-memory NeXus files supported if these used chunked storage.
- [x] Choices for users using different storage layouts are considered at the dataset level, multiple options: to chunked or not to chunk?, to compress or not?, gzip (sequential, current and future default) or blosc (multithreaded compression #747) 
- [x] Unit tests added
- [x] Extension for complex data would be possible, would rather delegate this though to another PR on "How to support data with complex numbers in pynxtools" #770 
- [x] Fixed get_field handling of string types and distinguish properly from plain objects/structs that Metainfo does not supported, deactivated also parsing of such content. 

Currently most datasets by pynxtools-plugins are written using the simple contiguous storage layout. Upon parsing these will be loaded fully into main memory, unpacked via hdf_node[...] at once if of dtype kind iufc.
If chunked storage layout is used, iterating over chunks respectively hyperslabs is not taken advantage of.
Consequently, the old implementation unnecessarily loads entire datasets into main memory instead of processing these off chunk-by-chunk. For large datasets, e.g. image and spectra stacks the impact is significant, for laptops eventually deal breaking: keep a rather flat few MiB per chunk flat RAM usage profile rather than provoke GiB spikes that may exceed even system max RAM. These spikes are particularly nasty if hosts are shared by multiple users.

Pitfalls: np.mean has optimized numerics (not only for speed but also compensating precision), chunking may need https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

- [x] Implementation
- [x] Testing individual parts, specifically Welford part
- [x] Testing in production


src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+            if hdf_value.dtype.kind in "iufc":
+                return hdf_value
+    else:  # bool
+        if hdf_node.shape != ():
+            return bool(hdf_node[0])
+        else:
+            return bool(hdf_node[()])
 
-            return decode_array(hdf_value)
+    return hdf_node[()].decode()
```
RubelMozumder: It has two issues:
1. The last statement `return hdf_node[()].decode()` will never be executed.
2. If `hdf_node` contains string, it would raise an error while attempting to decode.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
     return new_section
 
 
-def _get_value(hdf_node):
+def _get_field(hdf_node):
     """
-    Get value from hdf5 node
+    Fetching routine to use for non-iuf and contiguous storage layout
+    Get value from hdf5 node if scalar.
+    Get one value sampled from hdf5 node if non-scalar.
```
RubelMozumder: The docstring says only one sampled value will be returned, but the implementation returns the entire array.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+                    delta = value - mean
+                    mean += delta / np.float64(n)
+                    # delta2 = value - mean
+                    # M2 += delta * delta2
+        mean_result = mean if n > 0 else np.float64(np.nan)
+        # std_result = np.float64(np.sqrt(M2 / np.float64(n))) if n > 0 else np.float64(np.nan)
+        # to match np.std uses ddof=0 by default, use M2 / (n - 1) for sample std
+
+    # need to cast to correct return type
+    if hdf_node.dtype.kind == "iu":
```
RubelMozumder:
```suggestion
    if hdf_node.dtype.kind in "iu":
```

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+    Get field stats when hdf_node uses contiguous storage layout
+    """
+    stats: dict = {}
+
+    stats["__min"] = np.float64(+np.inf)
+    stats["__max"] = np.float64(-np.inf)
+    n = np.int64(0)
+    sum = np.float64(0.0)
+
+    field = _get_field(hdf_node)
```
RubelMozumder: `_get_field` is consistent to reutrn array of strings.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+        stats["__max"] = np.maximum(stats["__max"], np.max(field[mask]))
+
+        sum = np.sum(field[mask], dtype=np.float64)
+        n += np.int64(n_values)
+        mean_result = sum / np.float64(n) if n > 0 else np.float64(np.nan)
+    else:
+        mean_result = np.float64(np.nan)
+
+    # need to cast to correct return type
+    if hdf_node.dtype.kind == "iu":
```
RubelMozumder:
```suggestion
    if hdf_node.dtype.kind in "iu":
```

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+    field = _get_field(hdf_node)
+    mask = np.isfinite(field)
+    n_values = np.count_nonzero(mask)
+    if n_values > 0:
+        stats["__min"] = np.minimum(stats["__min"], np.min(field[mask]))
+        stats["__max"] = np.maximum(stats["__max"], np.max(field[mask]))
+
+        sum = np.sum(field[mask], dtype=np.float64)
+        n += np.int64(n_values)
+        mean_result = sum / np.float64(n) if n > 0 else np.float64(np.nan)
```
RubelMozumder, line 250:
```suggestion
        mean_result = sum / np.float64(n)
```

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -363,20 +522,23 @@ def _populate_data(
...
+                        # I do not think that taking the mean as a proxy is useful
+                        if suffix != "__mean":  # , "__std"):
+                            stat_metainfo_def = resolve_variadic_name(
+                                current.m_def.all_quantities, concept_basename + suffix
+                            )
+                            stat = MQuantity.wrap(
+                                field_stats[suffix], instance_name + suffix
+                            )
+                            current.m_set(stat_metainfo_def, stat)
+                            # stat.m_set_attribute("m_nx_data_path", hdf_node.name)
```
RubelMozumder, line 526: Why are the statistical quantities commented out?

src/pynxtools/nomad/schema_packages/schema.py
```diff
@@ -683,20 +683,22 @@ def _add_quantity_stats(container: Section, quantity: Quantity):
...
+            dtype = FIELD_STATISTICS[suffix]["type"]
+            container.quantities.append(
+                Quantity(
+                    name=basename + suffix,
+                    variable=quantity.variable,
+                    shape=[],
+                    type=dtype if dtype else quantity.type,
+                    description="This is a NeXus template property. "
+                    "This quantity holds specific statistics of the NeXus data array.",
+                )
```
RubelMozumder, line 700: The `mean` stat is already calculated, but not used in any statistical data domain.

RubelMozumder (dismissed):
This PR needs atom tests to verify that all changes will reflect the previous stats. I found some bugs while reading the code. Probably there are a few left.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
-                        result.append(decode_array(x))
-                    else:
-                        result.append(str(decode_or_not(x)))
-                return result
+                return decode_array(hdf_value)
+            return str([i for i in hdf_value.astype(str)])  # fallback for other arrays
+        else:
+            if hdf_value.dtype.kind in "iufc":
+                return hdf_value
+    else:  # bool
```
rettigl: Why this separation of bool values?

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+    # that is also not without debate it is questionable though whether this warrants
+    # to use then one of the most costly algorithms despite it being theoretically the
+    # most precise one
+
+    # the implementation here shows two approaches:
+    # i) classical Welford
+    # ii) vectorized implementation of the naive formula summarizing over chunks
+    # for computing mean and std using np.float64 in the accumulator though
+    # iii) using np.float128 precision would result in software emulation on the hardware
+    # making the computation substantially more costly than for np.float64
```
rettigl: The precision considerations are not entirely clear to me. Do you have an example when a naive vectorized implementation fails, and is this a likely scenario?

rettigl: To benefit from the memory saving, one anyways needs to sequentially loop through the chunks, no?

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +125,176 @@ def _to_section(
...
+            slab = hdf_node[chunk]
+            values = slab[np.isfinite(slab)]
+            number_of_values = values.size
+            if number_of_values > 0:
+                stats["__min"] = np.minimum(stats["__min"], values.min())
+                stats["__max"] = np.maximum(stats["__max"], values.max())
+
+                mean_sum += np.sum(values, dtype=np.float64)
+                n += np.int64(number_of_values)
+        mean_result = mean_sum / np.float64(n) if n > 0 else np.float64(np.nan)
```
rettigl, line 192: If we assume chunks to be uncorrelated, we can also pretty easily estimate the variance or standard deviation, I think. This off course does not have to be true...

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -292,35 +427,59 @@ def _populate_data(
...
-                        ]
-                        field = field_stats[0]
-                        if not np.isfinite(field):
+            # Metainfo does not support precision higher than i8, u8, f8, c16
+            # is f2 supported ? maybe silently promoted to float or f8
+            if hdf_node.dtype.kind in "iufc" and hdf_node.dtype.itemsize > 8:
+                self._logger.warning(
+                    f"error while setting field {data_instance_name} in {current.m_def} precision {hdf_node.dtype.itemsize} too high for {field_name}"
+                )
+                return
```
rettigl, line 407: I would rather report the warning, but still cast down to f8 instead of not parsing.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -292,35 +427,59 @@ def _populate_data(
...
+                    if hdf_node.chunks is not None:  # iterate over hyperslabs (chunks)
+                        field_stats = _get_field_stats_iuf_chunked(hdf_node)
+                    else:  # load entire contiguous storage layout dataset at once
+                        field_stats = _get_field_stats_iuf_contiguous(hdf_node)
+
+                    for suffix in FIELD_STATISTICS:
+                        if not np.isfinite(field_stats.get(suffix, np.nan)):
                             self._logger.info(
-                                "set NaN for field of an array",
+                                f"found {suffix} non existent or not finite for iuf value",
```
rettigl: iuf might be difficult to understand here

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -292,35 +427,59 @@ def _populate_data(
...
                                 target_name=field_name + "[" + data_instance_name + "]",
                             )
                             return
```
rettigl, line 428: Does it make sense to stop parsing here entirely?

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -292,35 +427,59 @@ def _populate_data(
...
+                else:
+                    field = hdf_node[()]
+                if not np.isfinite(field):
+                    self._logger.info(
+                        "found non finite c value",
+                        target_name=field_name + "[" + data_instance_name + "]",
+                    )
+                    return
+            else:
+                field = _get_field(hdf_node)
```
rettigl: This can put potentially very large datasets in metainfo, no?

tests/nomad/welford_test.ipynb
rettigl, line 1: I think this should be eventually removed, tests should only contain runnable unit tests. What is its purpose?

rettigl (commented):
There are a lot of changes, where I don't completely understand many of them. Certainly needs thorough testing of the different branches.
The removing of std as field statistics is okay for me, as I never used this in an app or so. But I think under the assumption of statistical independence of the different chunks, we can also cheaply provide an estimation.

mkuehbach:
Hi, @rettigl, as you are currently reviewing, (thanks for this) would you mind if we replace the `mean` value that is used as a representative for value of array datasets with `0.5 * (min + max)`. Using the `mean` for arrays was anyway always a crude workaround, showing just also the first value of a multidimensional array is equally questionable, dropping adding of array datasets altogether I think is a too hard step I had the impression we do not want to pursue at this point, right ?

I will work through your feedback later, currently still on #752

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Yes, that was one of the root causes for excessive memory consumption, when updated, I will in this PR also include that pynxtools by default uses the chunked based storage layout, compression will of course stay optional.

If there are very specific demands to have some datasets use explicitly the so far used contiguous storage layout we should discuss this as well, probably an Area B Tuesday meeting is a good venue where this PR here, as well as the ones on compression, and #752 could be discussed all together.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: True, given that we allow people since #741 to chunk explicitly with a focus on displaying performance, I did not wanted to shortcut though the implementations here and rather have a robust solution. On the other side of extremes a full Welford though is not worth the substantial extra costs it will demand, when all we use it for is just to fill the value of what should be some useful scalar representative of an array's values.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Sequentially looping through the chunks is the price to pay for getting a flat memory base line.

One could, in a further optimization loop also parallel across batches of chunks but that would demand a parallel file system which is not what every e.g. NOMAD deployment runs or what is available to each user.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Taking the distribution of how frequently h5py.Datasets across our typical examples are of type bool, we can pull up in front the more frequently used types.

rettigl:
> Hi, @rettigl, as you are currently reviewing, (thanks for this) would you mind if we replace the `mean` value that is used as a representative for value of array datasets with `0.5 * (min + max)`. Using the `mean` for arrays was anyway always a crude workaround, showing just also the first value of a multidimensional array is equally questionable, dropping adding of array datasets altogether I think is a too hard step I had the impression we do not want to pursue at this point, right ?
> 
> I will work through your feedback later, currently still on #752

I think we should discuss what this might be useful for in the first place. If anything, I think a mean value is still the most universal 1D metric, even though I admittedly also never used it.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Supportive of this idea but would say that is another feature, personally would argue this to be a typical follow-up PR that addresses also #770

src/pynxtools/nomad/parsers/parser.py
mkuehbach: indeed I should better make it explicit

src/pynxtools/nomad/parsers/parser.py
mkuehbach: One could although so far we rather had this thought that the NOMAD pipeline should not be the strictest gatekeeper. As all statistics are at least from NeXus anyway not part of the standard, expect for `_errors` would also be questionable as to why we should stop parsing further instance data from the file just because something beyond the standard is breaking. For sure required is that we do not push non-finite values to elasticsearch.

tests/nomad/welford_test.ipynb
mkuehbach: It was meant for development purposes, supporting also here what also Rubel mentioned that ipynb content should be refactored into proper unit tests with eps-proximity checks if the mean is computed close enough to the theoretical value (for synthetic test data) or for a result computed with np.float128.

mkuehbach:
> I think we should discuss what this might be useful for in the first place. If anything, I think a mean value is still the most universal 1D metric, even though I admittedly also never used it.

Thanks for clarifying that, I thought that you had been the most frequent user of these attributes but maybe using the axes string vector for filtering in the mpes app was already sufficient for distinguishing different types of datasets for you. Dunno why that thought of you being a strong advocate for `mean` and `std` remained in my mind, anyway that makes the implementation easier.

rettigl:
> Dunno why that thought of you being a strong advocate for `mean` and `std` remained in my mind, anyway that makes the implementation easier.

The original motivation came from me, but mostly about the min/max/ndim fields, which I use in the mpes app.

mkuehbach:
> > Dunno why that thought of you being a strong advocate for `mean` and `std` remained in my mind, anyway that makes the implementation easier.
> 
> The original motivation came from me, but mostly about the min/max/ndim fields, which I use in the mpes app.

Exactly that I recall, was unsure about mean and std though technically though setting a value is required for a quantity as min, max, ... are attributes to the quantity, so probably that's why Sandor then though well why not use the `mean`.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Fixed

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Modified

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Fixed

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Fixed

src/pynxtools/nomad/parsers/parser.py
mkuehbach: comment rejected, we must not accept a devision by zero

src/pynxtools/nomad/parsers/parser.py
mkuehbach: after discussion with @rettigl and @RubelMozumder we decided to drop std, min, max, mean, size, and ndim remain, I modified the code accordingly

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Please clarify

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Changed

src/pynxtools/nomad/schema_packages/schema.py
mkuehbach: Please clarify

tests/nomad/welford_test.ipynb
mkuehbach: Removed it

src/pynxtools/nomad/parsers/parser.py
rettigl: Ah, okay, this only stops the parsing for this one field. Thought initially this was for the whole dataset. Then is fine.

tests/nomad/test_parsing_opt.py
```diff
@@ -0,0 +1,200 @@
...
+        ## pytest.param("contiguous", "float128", id="contiguous-float128"),
+        ## pytest.param("contiguous", "complex32", id="contiguous-complex32"),
+        ## pytest.param("contiguous", "complex64", id="contiguous-complex64"),
+        ## pytest.param("contiguous", "complex128", id="contiguous-complex128"),
+    ],
+)
+def test_parse_file_array_statistics(storage_layout, data_type, tmp_path, caplog):
+    """Test validation of a NeXus/HDF5 with the same content but different storage layout."""
+    file_path = tmp_path / f"{storage_layout}.{data_type}.nxs"
+    prng = np.random.default_rng(seed=42)  # deterministic seeding
```
rettigl, line 92: I would not use deterministic seeding. With this, you might miss some boundary cases, that are exposed with true random data, and it should work for true random data anyways

tests/nomad/test_parsing_opt.py
```diff
@@ -0,0 +1,200 @@
...
+        if np.issubdtype(data_type, np.unsignedinteger):
+            dat = prng.integers(
+                0, np.iinfo(data_type).max, size=n_values, dtype=data_type
+            ).reshape(-1, 50, 50)
+            mean = np.asarray(
+                np.sum(dat, dtype=np.float128) / np.float128(dat.size), dtype=data_type
+            ).item()
+        elif np.issubdtype(data_type, np.integer):
+            dat = prng.integers(
+                np.iinfo(data_type).min,
```
rettigl: Should this not also work for unsigneds?

tests/nomad/test_parsing_opt.py
```diff
@@ -0,0 +1,200 @@
...
+                dtype=data_type,
+            ).reshape(-1, 50, 50)
+            mean = np.asarray(
+                np.sum(dat, dtype=np.float128) / np.float128(dat.size), dtype=data_type
+            ).item()
+        elif np.issubdtype(data_type, np.floating):
+            dat = prng.random(size=n_values).astype(data_type).reshape(-1, 50, 50)
+            mean = np.asarray(
+                np.sum(dat, dtype=np.float128) / np.float128(dat.size), dtype=data_type
+            ).item()
```
rettigl: This can also be moved outside the if/elif

tests/nomad/test_parsing_opt.py
rettigl, line 1: I suggest to add tests for all statistics to verify proper function. Assessing the impact on memory consumption by automatic testing is probably not so simple.

rettigl (approved):
I did not go through everything again, nor did I test it personally, but looks now okay for me, apart from some comments to tests.

tests/nomad/test_parsing_opt.py
mkuehbach: I see value in both: Deterministic seeding on the positive side assures that we probe for each data_type the same sequence of differences just mapped onto the respective full range of each data_type. On the flipside, sure each seeding may or not hit specific edge cases that a non-deterministic seeding might catch. Either way, with testing numerically we likely never get all edge cases unless having a theoretical analysis of the code. Would decide what the majority thinks is useful.

Agree that it needs to work anyway for all cases irrespective if random data stream in or not.

tests/nomad/test_parsing_opt.py
mkuehbach: np.integer is the super-class including np.unsignedinteger

tests/nomad/test_parsing_opt.py
mkuehbach: Yes

tests/nomad/test_parsing_opt.py
rettigl: Sure, but: 
```
np.iinfo(np.uint16).min
0
```
So no need to the separation.

tests/nomad/test_parsing_opt.py
mkuehbach: On the first point, I agree that I would be good to add.

On the second point, it is possible but involved, the tricky part is to really catch peak memory consumption, sampling it needs here. I think adding this also in this PR is disproportionate. Agree though that is a useful professionalization topic to include in production code, maybe an aspect to consider in FAIRmat II.

tests/nomad/test_parsing_opt.py
mkuehbach: You right, np.iinfo(*).min is 0 for all unsigneds, yeah I should process these two together.

tests/nomad/test_parsing_opt.py
lukaspie: > Agree that it needs to work anyway for all cases irrespective if random data stream in or not.

If that is the case, I would say we should use a random seed.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -126,37 +126,167 @@ def _to_section(
...
+    # that is also not without debate it is questionable though whether this warrants
+    # to use then one of the most costly algorithms despite it being theoretically the
+    # most precise one
+
+    # the implementation here shows two approaches:
+    # i) classical Welford
+    # ii) vectorized implementation of the naive formula summarizing over chunks
+    # for computing mean and std using np.float64 in the accumulator though
+    # iii) using np.float128 precision would result in software emulation on the hardware
+    # making the computation substantially more costly than for np.float64
```
lukaspie: These are way to many inline comments. I am generally not a fan of large inline comments, mostly they can become outdated if the actual code changes. If this is something to be documented, it should go to the docs.

tests/nomad/test_parsing_opt.py
lukaspie: I think having the first test that Laurenz suggested would be needed for this PR.

lukaspie (approved):
I think it's fine, maybe another test should be added.

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Seems that was ok

src/pynxtools/nomad/schema_packages/schema.py
mkuehbach: Yes, we decided we will keep the mean but drop the stdev computation

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Seems acceptable as no further feedback streamed in

tests/nomad/test_parsing_opt.py
mkuehbach: I am against using a random seed.

Using a random seed makes it possible that one time the test fails but the next time it succeeds.
All depend on how strict the numerical proximity assertion np.isclose is implemented.
Because with a random seed "By default, with no seed provided, [default_rng](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng) will seed the RNG from nondeterministic data from the operating system and therefore generate different numbers each time".

Yes doing so over time will visit more cases but at the costs of non-deterministic tests.
Sure one could set the tolerance using rtol and say accept still if within +-1% about the ground truth but that is not tighter. I like to have it tight here.

tests/nomad/test_parsing_opt.py
mkuehbach: Left a comment why numpy does not like this and we should keep it this way for now

tests/nomad/test_parsing_opt.py
mkuehbach: Done

src/pynxtools/nomad/parsers/parser.py
mkuehbach: Summarized in the source code and moved longer part to docs

tests/nomad/test_parsing_opt.py
mkuehbach: Addressed as well

PR #763, opened 2026-03-26
testing why premature exiting

mkuehbach:


mkuehbach:
was used only for tests, closed in favor for #752 that has the full and working implementation

mkuehbach:
Closed, draft, used for a demo, local backup by @mkuehbach exists

PR #775, opened 2026-05-07
Introduce unified pynx CLI with distributed sub-command modules

### LINKED ISSUE #719: Clean up CLI namespace
Currently, `pynxtools` installs multiple standalone CLI scripts:
```
read_nexus → pynxtools.nexus.nexus:main
dataconverter → pynxtools.dataconverter.convert:main_cli
generate_eln → pynxtools.eln_mapper.eln_mapper:get_eln
validate_nexus → pynxtools.dataconverter.validate_file:validate_cli
```
`dataconverter` has another set of subcommands (`convert` (aliassed), `generate-template`, `get-readers`).

Each is installed as a separate top-level command. This polluts the global shell namespace and it makes it harder for users to discover functionality. Consistent documentation of usage is also hard. We **_could_** change this, i.e., make the whole CLI interface available through one namespace.

Proposal: **Consolidate under a single CLI namespace**
- Introduce a short, unique top-level CLI command, e.g., `pynx ...`
- Expose current functionality as subcommands:
  ```console
  pynx read             # read_nexus
  pynx convert          # dataconverter
  pynx generate-eln     # generate_eln
  pynx validate         # validate_nexus
  ```
- Pros: 
  - Single entry point for all functionality, 
  - Easier to document and discover commands
  - Reduces shell namespace pollution
  - Easier to maintain and extend in the future
- Cons:
  - Slightly longer commands (`pynx convert` instead of `dataconverter`)
  - (biggest issue) May break existent implementations, examples, documentation, etc. 
    - We could keep old commands as aliases to avoid breaking existing scripts, but: this may require extra effort to maintain backward-compatible aliases (or deprecating them at some point)
    - From semver POV, we are fine as we are (technically) still in beta.

Alternatively, we can keep the separate CLI commands which has the advantage of no breaking changes for existing users, but then we have all the disadvantage (multiple binaries cluttering the namespace, inconsistent naming, etc.).

So, in summary, for discussion:
- Do we want this breaking change?
- If so, what's the best CLI entry point? `pynx`? (short, recognizable, low risk of collision)
- Should old commands remain available (and for how long) as aliases for backward compatibility?
- Could this be an opportunity to clean up subcommands (e.g., remove `generate-template` from the `dataconverter namespace`?)

Opinions @mkuehbach @sherjeelshabih @RubelMozumder @rettigl?

lukaspie:
Introduces a unified `pynx` CLI entry point that consolidates all pynxtools command-line tools under a single coherent command group. Each sub-package owns its own `cli.py`; `src/pynxtools/cli.py` assembles them into the `pynx` group.

### Command structure

| Command | Status | Previous name |
|---|---|---|
| `pynx convert` | renamed | `dataconverter` |
| `pynx validate` | renamed | `validate_nexus` |
| `pynx read` | renamed | `read_nexus` |
| `pynx generate-eln` | renamed | `generate_eln` |
| `pynx convert generate-template` | renamed | `dataconverter generate-template` |
| `pynx convert get-readers` | **new** | — |
| `pynx convert reader-info` | **new** | — |
| `pynx inspect-appdef` | **new** | — |

The three new commands expose functionality that was previously only accessible via the Python API: listing installed readers, inspecting a reader's supported NXDLs and file extensions, and listing the required/optional fields of a NeXus application definition.

### Backward compatibility

All legacy entry points (`dataconverter`, `validate_nexus`, `read_nexus`, `generate_eln`) remain installed as deprecated aliases. They continue to work but emit a deprecation warning on stderr.

### Other changes

- `pynx convert` invoked without a sub-command defaults to the conversion run, preserving the existing `dataconverter` UX
- CLI modules are distributed (each sub-package owns its `cli.py`) rather than centralized, which keeps plugin extensibility intact
- Docs updated throughout: all examples use `pynx` commands, API reference (`cli-api.md`) covers all commands including the new ones
- CLI smoke tests added for `pynx read`, `pynx inspect-appdef`, `pynx convert get-readers`, and `pynx convert reader-info`

Fixse #719


docs/reference/cli-api.md
```diff
@@ -1,47 +1,67 @@
 # API for command line tools
 
-`pynxtools` supports a number of command line applications. This page provides documentation for their current API.
+`pynxtools` provides a unified `pynx` entry point with subcommands for all CLI operations. This page documents the current API.
```
mkuehbach, line 3: The name you chose for the entry point is okay, it differs sufficiently from another python package with a similar name https://pynx.esrf.fr/en/latest/tutorial/script-cdi.html

docs/reference/cli-api.md
```diff
@@ -1,47 +1,67 @@
...
+
+The legacy entry points (`dataconverter`, `validate_nexus`, `read_nexus`, `generate_eln`) are still installed as deprecated aliases — they continue to work but emit a deprecation warning.
```
mkuehbach, line 5: We should deprecate these with the `v1.0.0` release

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
...
+    "--ignore-undocumented",
+    is_flag=True,
+    default=False,
+    help="Ignore all undocumented concepts during validation.",
+)
+@click.option(
+    "--fail",
+    is_flag=True,
+    default=False,
+    help="Fail conversion and don't create an output file if the validation fails.",
```
mkuehbach, line 152: should be reformulated and "don't" always better "do not"

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
...
+@click.option(
+    "--skip-verify",
+    is_flag=True,
+    default=False,
+    help="Skips the verification routine during conversion.",
+)
+@click.option(
+    "--mapping",
+    help=(
+        "Takes a <name>.mapping.json file and converts data from given input files. "
```
mkuehbach, line 165:
```suggestion
        "Takes a <name>.mapping.json file"
```

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
...
+    if params_file:
+        try:
+            _convert(**parse_params_file(params_file))
+            return
+        except TypeError as exc:
+            sys.tracebacklimit = 0
+            raise click.UsageError(
+                "Please make sure you have the following entries in your "
+                "parameter file:\n\n# NeXusParser Parameter File - v0.0.1"
+                "\n\ndataconverter:\n\treader: value\n\tnxdl: value\n\tin"
```
mkuehbach, line 203: the term "dataconverter" we should then no longer use rather the new term

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
...
+                "put-file: value"
+            ) from exc
+    if nxdl is None:
+        raise click.UsageError("Missing option '--nxdl'")
+    if mapping:
+        logger.warning(
+            "The --mapping option is deprecated. Please use a config file with the -c argument instead."
```
mkuehbach, line 210: I see so config file is the new mapping file and then also params exists, I have to say I personally find `semantic-mapping` or `concept-mapping` better suited terms when at the same time there is a `params` file, I rather suggest to have a `config` file which is the `params` and a `concept-mapping` option for the here `config`.

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
...
+        kwargs["config_file"] = config_file
+
+    file_list = []
+    for file in files:
+        if os.path.isdir(file):
+            from pathlib import Path
+
+            p = Path(file)
+            for f in p.rglob("*"):
+                if f.is_file():
```
mkuehbach, line 223: Should warn if not throw if a file is not existent instead of silently dropping

src/pynxtools/dataconverter/cli.py
```diff
@@ -0,0 +1,354 @@
+#
```
mkuehbach, line 1: The --append click option should not be removed.

src/pynxtools/nexus/cli.py
```diff
@@ -0,0 +1,121 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""CLI commands for NeXus file inspection.
+
+Exposes two commands consumed by the top-level ``pynx`` group:
+
+``read``
+    Annotate and inspect a NeXus/HDF5 file (``pynx read NEXUS_FILE``).
```
mkuehbach: Which types of annotations?

src/pynxtools/nexus/cli.py
```diff
@@ -0,0 +1,121 @@
...
+    "--level",
+    type=click.Choice(["required", "recommended", "optional"]),
+    default="required",
+    show_default=True,
+    help="Minimum optionality level to include.",
+)
+def inspect_appdef(
+    nxdl: str, level: Literal["required", "recommended", "optional"] = "required"
+):
+    """List fields of a NeXus application definition.
```
mkuehbach: Lists fields with that `level`, right?

src/pynxtools/cli.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Top-level ``pynx`` CLI dispatcher.
+
+All pynxtools command-line tools are available under this single entry point::
+
+    pynx read NEXUS_FILE              # annotate and inspect a NeXus/HDF5 file
+    pynx convert [files...]           # convert data to NeXus
+    pynx convert generate-template    # generate a conversion template
```
mkuehbach:
```suggestion
    pynx convert generate-template    # generate and display a conversion template dictionary
```

src/pynxtools/cli.py
```diff
@@ -0,0 +1,54 @@
...
+    pynx convert get-readers          # list installed readers
```
mkuehbach:
```suggestion
    pynx convert get-readers          # lists all installed readers
```

mkuehbach: ?

src/pynxtools/cli.py
```diff
@@ -0,0 +1,54 @@
...
+    pynx convert reader-info          # show reader capabilities
+    pynx validate NEXUS_FILE          # validate a NeXus file against its appdef
```
mkuehbach:
```suggestion
    pynx validate NEXUS_FILE          # validate a NeXus file against its application definition
```

src/pynxtools/cli.py
```diff
@@ -0,0 +1,54 @@
...
+    pynx generate-eln                 # generate an ELN YAML scaffold
```
mkuehbach: we should mention a NOMAD ELN YAML
Thought that yields a working ELN instance, if so we should comment that this is the case instead of the less precise term `scaffold`

src/pynxtools/cli.py
```diff
@@ -0,0 +1,54 @@
...
+    pynx inspect-appdef NXDL          # list fields of an application definition
```
mkuehbach:
```suggestion
    pynx inspect-appdef NXDL          # list fields of an application definition with specific presence constraint
```

mkuehbach: ?

docs/how-tos/pynxtools/validate-nexus-files-other-tools.md
```diff
@@ -1378,7 +1378,7 @@ You can start the `punx` validation by running
     NeXus definitions version: main
     ```
 
-The output tables "findings" and "summary statistics" can be used to find error present in the NeXus file. As you can see, while the output is verbose and comprehensive, `punx` does not actually pick up on the issues that the [`pynxtools` validator finds](validate-nexus-files.md#validate_nexus).
+The output tables "findings" and "summary statistics" can be used to find error present in the NeXus file. As you can see, while the output is verbose and comprehensive, `punx` does not actually pick up on the issues that the [`pynxtools` validator finds](validate-nexus-files.md#pynx-validate).
```
mkuehbach:
```suggestion
The output tables "findings" and "summary statistics" can be used to find if errors are present in the NeXus file. While the output is verbose and comprehensive, you can observe that `punx` does not pick up on the issues that the [`pynxtools` validator](validate-nexus-files.md#pynx-validate) finds.
```

docs/how-tos/pynxtools/validate-nexus-files.md
```diff
@@ -34,18 +37,18 @@ In this how-to, we will learn how to use the `validate_nexus` and `read_nexus` c
 
 Note that you will need to have `pynxtools` installed in a Python environment. Learn more about the installation procedure in the [installation tutorial](../../tutorial/installation.md).
 
-## **`validate_nexus`**
+## **`pynx validate`**
 
-After installation, you can invoke the help call of the `validate_nexus` tool from the command line:
+After installation, you can invoke the help call of the `pynx validate` tool from the command line:
```
mkuehbach:
```suggestion
After installation, you can invoke the `--help` call of the `pynx validate` tool from the command line:
```

docs/how-tos/pynxtools/validate-nexus-files.md
```diff
@@ -60,70 +63,70 @@ Note that here we are passing the `--ignore-undocumented` flag to the validation
...
     ```bash exec="on" source="material-block" result="text"
-    read_nexus -f src/pynxtools/data/201805_WSe2_arpes.nxs 
+    pynx read src/pynxtools/data/201805_WSe2_arpes.nxs 
     ```
 
-In the output, several concepts are reported as "NOT IN SCHEMA". These are exactly those fields that we ignored with the `ignore-undocumented` flag about. NeXus allows to add additional groups/fields/attributes to NeXus files. However, such reports from the `validate_nexus`/`read_nexus` tools can also be indicators that a given part of the file is not compliant with the application definition as expected (e.g., because its name does not fit with the name of the intended NeXus concept).
+In the output, several concepts are reported as "NOT IN SCHEMA". These are exactly those fields that we ignored with the `ignore-undocumented` flag about. NeXus allows to add additional groups/fields/attributes to NeXus files. However, such reports from the `pynx validate`/`pynx read` tools can also be indicators that a given part of the file is not compliant with the application definition as expected (e.g., because its name does not fit with the name of the intended NeXus concept).
```
mkuehbach:
```suggestion
In the output, several concepts are reported as "NOT IN SCHEMA". These are exactly those fields that we ignored with the `ignore-undocumented` flag about. NeXus allows to add additional groups/fields/attributes to NeXus files. However, such reports from the `pynx validate` or `pynx read` tools can also be indicators that a given part of the file is not compliant with the application definition as expected (e.g., because its name does not fit with the name of the intended NeXus concept).
```

mkuehbach (changes requested):
Good further professionalization of pynxtools

docs/reference/cli-api.md
lukaspie: agreed

src/pynxtools/cli.py
lukaspie: correct

src/pynxtools/dataconverter/cli.py
lukaspie: Yes, that got lost when shuffling things around, I re-added it.

src/pynxtools/dataconverter/cli.py
lukaspie: done

src/pynxtools/dataconverter/cli.py
lukaspie: I changed it to "Takes a <name>.mapping.json file to map data from a given input files"

src/pynxtools/dataconverter/cli.py
lukaspie: No, we still need that in the params file. The dataconverter object still exists, but it is called differently in the CLI.

src/pynxtools/dataconverter/cli.py
lukaspie: This has been the case for a while. The config file is called config instead of mapping because it is the configuration of the reader. It provides more than just the mapping. See https://fairmat-nfdi.github.io/pynxtools/learn/pynxtools/multi-format-reader.html#special-rules

Anyway, this is out of the scope of what this PR addresses.

src/pynxtools/dataconverter/cli.py
lukaspie: Added. Also added an error raise if no valid input file is given.

src/pynxtools/nexus/cli.py
lukaspie: I changed the generic description here to be a bit more specific

src/pynxtools/nexus/cli.py
lukaspie: I changed this whole docstring to

    """List the concept paths defined in a NeXus application definition.

    Prints all paths at or above the requested optionality level (required,
    recommended, or optional). Useful for understanding what data must be
    provided before converting with ``pynx convert``, or as a lightweight
    alternative to ``pynx convert generate-template`` when you only need
    the field names rather than a fillable JSON skeleton.

    NXDL: application definition name, e.g. NXmpes
    """

src/pynxtools/cli.py
lukaspie: I changed this to "generate a reader or NOMAD ELN"

lukaspie:
One point to think about is examples, tutorials, etc., where we have used `dataconverter` for a while. We would probably want to slowly change those to `pynx converter`, but not urgent as `dataconverter` still persists (deprecated).

src/pynxtools/dataconverter/cli.py
mkuehbach: given input file
```suggestion
        "Takes a <name>.mapping.json file and converts data from given input file. "
```

lukaspie: No, that works with multiple input files.

src/pynxtools/dataconverter/cli.py
mkuehbach: okay with having this addressed in a different pr but current solution

mkuehbach (approved):
See two remaining comments

PR #776, opened 2026-05-08
Include the old convension to call nexsu_parser in the aliases list.

### LINKED ISSUE #773: [Bug]: Parser aliasing
### Contact Details

_No response_

### What happened?

In the recent version of the pynxtools package, the entry point locations have changed. This changes the identifiers of these entry points, leading to some subtle problems when processing older data.

E.g. the parser entry point seems to have changed from `pynxtools.nomad.entrypoints:nexus_parser` to `pynxtools.nomad.parsers:nexus_parser`. Any data parsed with the old entry point remembers this old parser name, and when attempting to reprocess the data, will not be able to find the matched parser. To fix this problem, you should provide a [parser alias](https://nomad-lab.eu/prod/v1/docs/reference/plugins.html#parserentrypoint) for the old name.

### Relevant log output

```shell
Traceback (most recent call last):
  File "opt/venv/lib/python3.12/site-packages/nomad/processing/data.py", line 1242, in full_entry_metadata
    self._apply_metadata_from_mongo(upload, entry_metadata)
  File "opt/venv/lib/python3.12/site-packages/nomad/processing/data.py", line 1166, in _apply_metadata_from_mongo
    parser = parser_dict[self.parser_name]
             ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'pynxtools.nomad.parsers:nexus_parser'
```

RubelMozumder:
closes #773 

In some previous version of `pynxtools`, we registered the nexus parser in a different way, `pynxtools.nomad.entrypoints:nexus_parser`. But later changing the entry point name to `pynxtools.nomad.parsers:nexus_parser` causes an error in nomad. This is fixed by adding the old convention in aliases in NexusParserEntryPoint.

I tested if the aliases are registered properly.

```ipython
In [1]: from nomad.parsing.parsers import match_parser, parser_dict, parsers

In [2]: for k, v in parser_dict.items():
   ...: 
   ...:     if 'pynxtools' in k:
   ...:         print(f"EntryPoint: {k}")
   ...:
EntryPoint: pynxtools.nomad.parsers:nexus_parser
EntryPoint: pynxtools.nomad.entrypoints:nexus_parser
```

### DIFF
--- a/src/pynxtools/nomad/parsers/__init__.py
+++ b/src/pynxtools/nomad/parsers/__init__.py
@@ -36,4 +36,7 @@ nexus_parser = NexusParserEntryPoint(
     description="A parser for nexus files.",
     mainfile_name_re=r".*\.nxs",
     mainfile_mime_re="application/x-hdf*",
+    aliases=[
+        "pynxtools.nomad.entrypoints:nexus_parser",
+    ],
 )


lukaspie:
It looks good, but we need to check it works. I guess we can only find out if we reprocess older data in production (like what was reported in the issue). @lauri-codes that probably solves the issue, right?

lukaspie (approved):

RubelMozumder:
> It looks good, but we need to check it works. I guess we can only find out if we reprocess older data in production (like what was reported in the issue). @lauri-codes that probably solves the issue, right?

Yeah, I logically checked if aliases refer to the correct entry point.

```ipython
In [4]: for k, v in parser_dict.items():
   ...: 
   ...:     if 'pynxtools' in k:
   ...:         print(f"Entry point :\n {k}: {v}")
   ...: 
Entry point :
 pynxtools.nomad.parsers:nexus_parser: pynxtools.nomad.parsers:nexus_parser
Entry point :
 pynxtools.nomad.entrypoints:nexus_parser: pynxtools.nomad.parsers:nexus_parser
 ```

lauri-codes:
Looks good, we can test it during the next release.

PR #778, opened 2026-05-11
Add `bytes` handler for json map reader

RubelMozumder:
closes #759 

This PR handles the JSON map reader when `hdf5` is one of the input data files.  

- [x] Checks dataset from hdf5 and converts from bytes to str (if needed)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,9 @@
     },
     "editor.defaultFormatter": "charliermarsh.ruff"
   },
-  "python.testing.pytestArgs": ["tests"],
+  "python.testing.pytestArgs": [
+    "."
+  ],
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,9 +11,7 @@
     },
     "editor.defaultFormatter": "charliermarsh.ruff"
   },
-  "python.testing.pytestArgs": [
-    "."
-  ],
+  "python.testing.pytestArgs": ["tests"],
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true
 }
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -1596,20 +1596,47 @@ def nested_dict_to_slash_separated_path(
             flattened_dict[path] = val
 
 
-def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
+# def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
+#     """
+#     Return the payload (e.g. attribute) as Any except bytes.
+
+#     - If `payload` is `bytes`, decode it using the given encoding.
+#     - Otherwise, return it unchanged.
+
+#     Args:
+#         payload: Any type.
+#         encoding: The character encoding to use when decoding bytes.
+
+#     Returns:
+#         The attribute as a string, or None if input was None.
+#     """
+#     if isinstance(payload, bytes):
+#         return payload.decode(encoding)
+#     return payload
+
+
+def decode_if_bytes(value: Any, encoding: str = "utf-8") -> Any:
+    """Recursively decode text-like bytes to Python strings.
+
+    Numeric arrays and non-text scalar values are returned unchanged.
     """
-    Return the payload (e.g. attribute) as Any except bytes.
+    if isinstance(value, bytes | np.bytes_):
+        return value.decode(encoding)
 
-    - If `payload` is `bytes`, decode it using the given encoding.
-    - Otherwise, return it unchanged.
+    elif isinstance(value, np.ndarray):
+        if value.dtype.kind == "S":
+            return np.char.decode(value, encoding)
+        elif value.dtype.kind == "O":
+            return np.vectorize(decode_if_bytes, otypes=[object])(value)
+        return value
 
-    Args:
-        payload: Any type.
-        encoding: The character encoding to use when decoding bytes.
+    elif isinstance(value, list):
+        return [decode_if_bytes(item, encoding) for item in value]
 
-    Returns:
-        The attribute as a string, or None if input was None.
-    """
-    if isinstance(payload, bytes):
-        return payload.decode(encoding)
-    return payload
+    elif isinstance(value, tuple):
+        return tuple(decode_if_bytes(item, encoding) for item in value)
+
+    elif isinstance(value, dict):
+        return {k: decode_if_bytes(v, encoding) for k, v in value.items()}
+
+    return value
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -1596,47 +1596,42 @@ def nested_dict_to_slash_separated_path(
             flattened_dict[path] = val
 
 
-# def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
-#     """
-#     Return the payload (e.g. attribute) as Any except bytes.
+def _decode_bytes_scalar(value: Any, encoding: str) -> Any:
+    """Decode scalar byte-like values and keep all other scalars unchanged."""
+    if isinstance(value, (bytes, np.bytes_)):
+        return value.decode(encoding)
+    return value
 
-#     - If `payload` is `bytes`, decode it using the given encoding.
-#     - Otherwise, return it unchanged.
 
-#     Args:
-#         payload: Any type.
-#         encoding: The character encoding to use when decoding bytes.
+def _decode_bytes_recursive(value: Any, encoding: str) -> Any:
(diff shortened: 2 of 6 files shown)


lukaspie (commented):
I think the `_decode_hdf_bytes` function is fairly similar to the existing [`decode_if_string` ](https://github.com/FAIRmat-NFDI/pynxtools/blob/618bed07b00c9f396b189d4771bd45b2395050cb/src/pynxtools/nexus/nexus.py#L34) function of the `nexus.py` module. They make very similar choices in how to decide bytes (not 100% the same), but have quite different purposes. Could you check if it makes sense to bring them together in one function? But only if it makes sense.

RubelMozumder:
> I think the `_decode_hdf_bytes` function is fairly similar to the existing [`decode_if_string` ](https://github.com/FAIRmat-NFDI/pynxtools/blob/618bed07b00c9f396b189d4771bd45b2395050cb/src/pynxtools/nexus/nexus.py#L34) function of the `nexus.py` module. They make very similar choices in how to decide bytes (not 100% the same), but have quite different purposes. Could you check if it makes sense to bring them together in one function? But only if it makes sense.

Yeah, both functions are not really for the same purpose. The function `decode_if_stting` always coerces the numpy array to a string (using `decoded_array.astype(str)`), even if the array contains is container of integers or floats. The scope of the function is to verify the nexus entity definition, which is pretty much always a string and makes sense to coerce to a string.

But the functions `_decode_hdf_bytes` focus on general conversion to their native data type and HDF5 data.

src/pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -55,6 +55,38 @@
 logger = logging.getLogger("pynxtools")
 
 
+def _decode_hdf_bytes(value: Any) -> Any:
```
lukaspie: @mkuehbach in `helpers.py`, there is this function:

```python
def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
    """
    Return the payload (e.g. attribute) as Any except bytes.

    - If `payload` is `bytes`, decode it using the given encoding.
    - Otherwise, return it unchanged.

    Args:
        payload: Any type.
        encoding: The character encoding to use when decoding bytes.

    Returns:
        The attribute as a string, or None if input was None.
    """
    if isinstance(payload, bytes):
        return payload.decode(encoding)
    return payload
```

Do you think the implementation here (which also unpacks numpy arrays) could be used instead of `decode_if_bytes` in the validator as well? I would like to avoid that we have several very similar functions in multiple places that essentially do the same thing.

.vscode/settings.json
```diff
@@ -11,7 +11,9 @@
     },
     "editor.defaultFormatter": "charliermarsh.ruff"
   },
-  "python.testing.pytestArgs": ["tests"],
+  "python.testing.pytestArgs": [
```
lukaspie: Why was this changed?

RubelMozumder: Removed!

src/pynxtools/dataconverter/readers/json_map/reader.py
mkuehbach: `decode_hdf_bytes` is useful and covers more cases than `decode_if_bytes` so yes, we should use `decode_hdf_bytes`.

The way this function though is currently written is not acceptable for performance reasons.
The issue is with all the list, tuple, as well as dict cases. Doing a recursion is fine but observe that for each recursive entry we run the same test and if cases for each element e.g.

This
```
    if isinstance(value, list):
        return [_decode_hdf_bytes(item) for item in value]
```

Will evaluate all five ifs for each item.

I recommend first to go with an if elif construct and further to pull the if case out of the function that does the actual recursion, so a recursion core function that just decodes.

Another issue with the function is that it assumes HDF5 content (given its name fragment `_hdf_`) but the argument is any, why not h5py.Dataset. If that does not work to avoid changes in other places in the code (which I am ok with accepting) we should at least rename the function to get rid of this hdf naming, as nothing in the function pins it to HDF5.

If the function were pinned for HDF5, we face anyway a more limited number of cases if we use h5py: regularly typed datasets, which will always be numpy arrays so the repetitive calls and ifs are completely unnecessary, or we face strings where they are also unnecessary for to always repeat, or we face byte structs (e.g. C++ structs dump into a binary data blob in HDF5) but these we should anyway not support and instead warn because these typically are (less and less so nowadays for creating flat dictionaries of concept, value, like
AccelerationVoltage, 15
BeamCurrent, 0.2)
this violates core design ideas of NOMAD and pynxtools so its fine to ignore them. In pynxtools-em for example we have this frequently for EDAX, we just unpack these arrays there and also not support it.

mkuehbach (changes requested):

RubelMozumder:
@mkuehbach comment is addressed!

mkuehbach:
My comment has not been addressed completely. The key point was that the if nest will still be evaluated if you are in the recursion which, yes is fine if your payload is trivial like a few kilobytes but not when it is MB or GB as you keep running unnecessary if clauses then millions of times. We should not unnecessary rush and introduce code that we have already a good understanding for that it is going to be slow in these cases.

On the flipside, there are also not yet that many of such cases, mind though that the decode_if_bytes function is planned to be used also in the validation and thus the effect is relevant.

On the other side, this PR is mainly addressing to reduce functionality duplication in the code base, good, and the case distinction made here adds value, good.

Taking in all these thoughts,  please add a TODO and comment about the possible performance penalty
in the e.g. above the line where you do the `isinstance(value, list)` for future updates and then I am fine with approving it.

Please also remove commented out code, its in the gh history anyway.

src/pynxtools/dataconverter/helpers.py
```diff
@@ -1596,20 +1596,47 @@ def nested_dict_to_slash_separated_path(
             flattened_dict[path] = val
 
 
-def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
-    """
-    Return the payload (e.g. attribute) as Any except bytes.
+# def decode_if_bytes(payload: Any, encoding: str = "utf-8") -> Any:
```
lukaspie: Should be removed if not used anymore

RubelMozumder:
> My comment has not been addressed completely. The key point was that the if nest will still be evaluated if you are in the recursion which, yes is fine if your payload is trivial like a few kilobytes but not when it is MB or GB as you keep running unnecessary if clauses then millions of times. We should not unnecessary rush and introduce code that we have already a good understanding for that it is going to be slow in these cases.
> 
> On the flipside, there are also not yet that many of such cases, mind though that the decode_if_bytes function is planned to be used also in the validation and thus the effect is relevant.
> 
> On the other side, this PR is mainly addressing to reduce functionality duplication in the code base, good, and the case distinction made here adds value, good.
> 
> Taking in all these thoughts, please add a TODO and comment about the possible performance penalty in the e.g. above the line where you do the `isinstance(value, list)` for future updates and then I am fine with approving it.
> 
> Please also remove commented out code, its in the gh history anyway.

I split the function to reduce the condition check in an if/elif clause. But, performance-wise, I think the last one was better than the current one. In Python, calling a function is more expensive than the condition check in an if clause. Whatever the payload size (GB/TB), for a numpy dataset, if the data type is homogeneous, the conversion to string happens at a lower level. But, if it is a numpy object, the data type can be heterogeneous, which needs a check for individual elements. The same individual datatype check is also required for elements and sub-elements of list, tuple, and dict.

src/pynxtools/dataconverter/helpers.py
RubelMozumder: Removed!

RubelMozumder:
> > My comment has not been addressed completely. The key point was that the if nest will still be evaluated if you are in the recursion which, yes is fine if your payload is trivial like a few kilobytes but not when it is MB or GB as you keep running unnecessary if clauses then millions of times. We should not unnecessary rush and introduce code that we have already a good understanding for that it is going to be slow in these cases.
> > On the flipside, there are also not yet that many of such cases, mind though that the decode_if_bytes function is planned to be used also in the validation and thus the effect is relevant.
> > On the other side, this PR is mainly addressing to reduce functionality duplication in the code base, good, and the case distinction made here adds value, good.
> > Taking in all these thoughts, please add a TODO and comment about the possible performance penalty in the e.g. above the line where you do the `isinstance(value, list)` for future updates and then I am fine with approving it.
> > Please also remove commented out code, its in the gh history anyway.
> 
> I split the function to reduce the condition check in an if/elif clause. But, performance-wise, I think the last one was better than the current one. In Python, calling a function is more expensive than the condition check in an if clause. Whatever the payload size (GB/TB), for a numpy dataset, if the data type is homogeneous, the conversion to string happens at a lower level. But, if it is a numpy object, the data type can be heterogeneous, which needs a check for individual elements. The same individual datatype check is also required for elements and sub-elements of list, tuple, and dict.


But in some cases, splitting the function into several specific functions reduces unnecessary `if condition` checks, especially when data is not numpy or a scalar.

lukaspie (approved):
I think that is okay now, with the clear TODO for a future optimization.

mkuehbach (approved):
OK

PR #779, opened 2026-05-12
Suggestion how to reduce unnecessary reporting of compression blosc settings to stdout

mkuehbach:


lukaspie (approved):
LGTM, I tested it locally; it does not log if compression is not used, but if compression _is_ used, I can see the log information

PR #786, opened 2026-05-19
Align the nexus_tree classes `nxdl.xsd` data model

lukaspie:
This remodels the `NexusNode` classes in `nexus_tree` to be closer connected to the definitions in `nxdl.xsd`. **This is a pre-requisite to be able to create the generator for NeXus Metainfo classes in Python.**

| Class| Function |
|-------|----------|
| `_NexusEntityBase(NexusNode)`| shared private base holding `dtype`, `items`, `open_enum`, `shape`, and `dim_symbols`. These represent XML attributes present on both `fieldType` and `attributeType` in the XSD. |
| NexusField(_NexusEntityBase)` |   models `fieldType` completely, adding `unit`, `long_name`, `interpretation`, `signal`, `axis`, `axes`, `primary`, `stride`, and `data_offset`. Field-only setters `_set_unit()` and `_set_field_attrs()` |
| `NexusAttribute(_NexusEntityBase)` |  models `attributeType`. No `unit`. |
| `NexusDefinition(NexusNode`) | replaces `NexusGroup(nx_class="NXroot")` as the root of `generate_tree_from`; carries `category`, `symbols` dict, and `ignoreExtra*` flags from the NXDL `definitionType` element. |
| Processing | `NexusVisitor` (ABC) |

Addiitional updates:
- `dim_symbols` added to `_NexusEntityBase`: parallel to shape, stores NXDL symbol names (e.g. "nP") for symbolic dimensions.
- Drop `dataconverter/nexus_tree.py` re-export shim; all imports now come from `pynxtools.nexus.nexus_tree` directly.
- Update all consumers: `validation.py`, `eln_mapper/{eln,reader_eln,schema_eln}.py`,  `dataconverter/cli.py`, `nexus/cli.py` — `NexusEntity` references replaced with `NexusField` or ``_NexusEntityBase` as appropriate; unit access guarded with `isinstance(node, NexusField)` checks.
- Update the documentation accordingly


src/pynxtools/nexus/nexus_tree.py
```diff
@@ -947,6 +947,66 @@ def __repr__(self) -> str:
         return f"{self.name} ({self.optionality[:3]})"
 
 
+class NexusDefinition(NexusNode):
+    """
+    Root node representing a complete NXDL definition file (``definitionType`` in ``nxdl.xsd``).
+
+    A definition is semantically distinct from a nested group (``NexusGroup``): it
```
RubelMozumder, line 954: IMO, having another class, NeXusBase e.g., NeXusBase, similar to the NexusDefinition would give a full and genuine representation of the NeXus Datamodel. But, if it is a very expensive work, it is ok to move ahead with this PR as it is.

RubelMozumder (approved):
PR description misleads a bit, probably came from some old text. 
`dtype, items, open_enum, shape, and dim_symbols` are not part of XML attributes in XSD in `fieldType` and `attributeType`. But, they are the active part of the NeXus tree.

Other than this, LGTM!

src/pynxtools/nexus/nexus_tree.py
lukaspie: Note that `NeXusDefinition` is _not_ meant for application definitions, but for the `definitionsType` in `nxdl.xsd` (which is used in **all** valid NXDL files).

So we don't really need a `NeXusBase` class, application definitions and base classes these are not defined in the nxdl.xsd per se.

RubelMozumder: Yeah, I also see NexusDefinition represents the all nxdl files as a full tree with a top-level node. Per implementation, it represents both `base and application` categories as well, which are part of the broader definition of NeXus. 

But, an attempt to open a base class is getting an error.
```
In [15]: 
    ...: from pynxtools.nexus.nexus_tree import NexusDefinition, generate_tree_from
    ...: 
    ...: root: NexusDefinition = generate_tree_from("NXdata")
    ...: print(root.category)   # "application"
    ...: print(root.symbols)    # {"nP": "...", ...}
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[15], line 3
      1 from pynxtools.nexus.nexus_tree import NexusDefinition, generate_tree_from
----> 3 root: NexusDefinition = generate_tree_from("NXdata")
```
Where it works for `NXspm` and other app def.

lukaspie:
> PR description misleads a bit, probably came from some old text. `dtype, items, open_enum, shape, and dim_symbols` are not part of XML attributes in XSD in `fieldType` and `attributeType`. But, they are the active part of the NeXus tree.
> 
> Other than this, LGTM!

Yes, you are right, these are not defined in XML, but rather represent the attributes defined in XML. I adjusted the PR description accordingly.

src/pynxtools/nexus/nexus_tree.py
RubelMozumder: I assumed it would also be able to represent base definition as well.

lukaspie: That's a limitation of `generate_tree_from`, not of `NexusDefinition`. Currently ´generate_tree_from` only works for application definition. I will make it work for base classes later when I create the metainfo generator.

Issue #292, opened 2024-03-21
[Bug]: Informative stdout gets printed to stderr


mkuehbach:
### Contact Details

markus.kuehbach@physik.hu-berlin.de

### What happened?

Running pynxtools-apm using pynxtools from apm-em-tracking branch I spotted that some informative messages of the are directed to stderr although they should be directed to stdout or some other non-error-type stream.
Here an example from an otherwise finely parsing use case monitored using command line calling dataconverter convert .... 1>STDOUT.txt 2>STDERR.txt

STDERR.txt contained thereafter:
Using apm reader to convert the given files:  
• [...]/production/R18_53705-v03.epos
• [...]/production/R18_53705-v03.rng
• eln_data_apm.yaml
• apm.oasis.specific.yaml
**The data entry corresponding to /ENTRY[entry]/definition/@version is required and hasn't been supplied by the reader.**
The output file generated: debug.aus_sydney_breen_1.nxs.

Only the line marked in bold I think warrants to be streamed to stderr as it is a warning everything else in stdout.

### Relevant log output

```shell
Here an example from an otherwise finely parsing use case monitored using command line calling dataconverter convert .... 1>STDOUT.txt 2>STDERR.txt

STDERR.txt contained thereafter:
Using apm reader to convert the given files:  
• [...]/production/R18_53705-v03.epos
• [...]/production/R18_53705-v03.rng
• eln_data_apm.yaml
• apm.oasis.specific.yaml
**The data entry corresponding to /ENTRY[entry]/definition/@version is required and hasn't been supplied by the reader.**
The output file generated: debug.aus_sydney_breen_1.nxs.

Only the line marked in bold I think warrants to be streamed to stderr as it is a warning everything else in stdout.
```

mkuehbach:
Inspected the situation further:
As of today's master branch state log level is INFO stream is None i.e. the default stream which is stderr.
So one can consider that info messages of the converter like above are indeed debug-type log messages for which we may use the interpretation that stderr is not only for errors so this is no bug, but:

In production runs now the only message indeed emitted by the converter unless there is at least a warning is alike to the above ones with always differing length. This avoids that one can check quickly just if the stderr content is 0 nothing so to skip to the next example instead of seeing some small byte payload that is always a similar message that (like the one above) that everything is right.

Two possible solutions:
[...]

PR #787, opened 2026-05-19
refactor: introduce ValidationVisitor implementing NexusVisitor

lukaspie:
## Summary

This PR implements the NeXus visitor architecture by migrating the HDF5 file validator to the `NexusVisitor` / `NexusFileHandler` pattern established by the annotator, and by consolidating the schema-resolution infrastructure that both tools share.

#### `dataconverter/validation.py` — `ValidationVisitor`

The free function `validate_hdf_group_against` (with its nested closures and `LRUCache`) is replaced by `ValidationVisitor`, a concrete `NexusVisitor` implementation driven by `NexusFileHandler`.

- File traversal, link resolution, and soft/external/broken-link detection are fully delegated to the handler.  `ValidationVisitor` only contains (limited) schema-resolution and problem-collection logic.
- `_find_node_for` now delegates to `NexusSchemaResolver.node_for_path` (see below), removing the bespoke `LRUCache`-backed closure. 
- `_variadic_node_exists_for` now calls `resolve_path` directly — deliberately bypassing the resolver cache to avoid polluting the shared HDF5-path cache with NXDL concept paths.
- The `best_namefit_of` local function inside `validate_dict_against` is replaced by `_select_best_namefit` (see below).
- `validate_hdf_group_against` is retained as a one-line backward-compatible wrapper.

#### `nexus/handler.py` — two optional hooks on `NexusVisitor`

Two optional hooks with default no-op implementations are added to the base class:

- `on_broken_link(hdf_path, link)` — called when a soft or external link cannot be resolved. `ValidationVisitor` overrides this to emit a `BrokenLink` problem.
- `on_external_link(hdf_path, link)` — called when an external link is first encountered, before the handler opens the external file.

`NexusFileHandler` now handles soft links, hard links (with cycle guard), and external links. Broken links dispatch `on_broken_link` and are otherwiseskipped. Cycle detection is extended to tolerate ancestor paths that do not
exist in the current root (e.g. when traversing an external file).

#### `nexus/nexus_tree.py` — `_select_best_namefit`

The scoreboard logic that was duplicated between `NexusNode.best_child_for` and `validate_dict_against.best_namefit_of` is extracted into a single module-level function `_select_best_namefit(name, nodes, hint)`.

- `best_child_for` is refactored to a two-pass approach: non-variadic exact match first, then `_select_best_namefit` on the variadic candidates.
- `required_fields_and_attrs_names` no longer adds `/@units` to the required entity list for fields with `unit="NX_UNITLESS"`

#### `nexus/schema_resolver.py` — `NexusSchemaResolver.node_for_path`

A new method `node_for_path(tree, path, node_type, nx_class, hint)` resolves a path within a known `NexusNode` tree without HDF5 file access.  This covers the `ValidationVisitor` use case (appdef and tree root are already known; NX_class
disambiguation is supplied explicitly) without duplicating the cache or the `best_child_for` call chain.

`resolve_path` gains support for `node_type="attribute"` in its signature so that callers can request attribute resolution at the last segment.

### What did not change

- The `validate_dict_against` entry point and its keyword arguments are unchanged.
- `validate_hdf_group_against` signature is unchanged (backward-compatible wrapper).
- All `ValidationProblem` codes and their log messages are unchanged.
- The `Annotator` and `NomadVisitor` implementations are not modified.

## Tests

- `tests/dataconverter/test_validation.py` — extended with link-traversal cases (soft links, external links, broken links within an entry).
- `tests/nexus/test_handler.py` — new file; covers `NexusFileHandler` linktraversal behaviour via a lightweight spy visitor, including cycle detection and broken external links.

### DIFF
--- a/docs/how-tos/pynxtools/implement-a-visitor.md
+++ b/docs/how-tos/pynxtools/implement-a-visitor.md
@@ -34,6 +34,18 @@ class MyVisitor(NexusVisitor):
     def on_complete(self, root: h5py.File) -> None: pass
 ```
 
+Two additional optional hooks have default no-op implementations and may be overridden:
+
+| Hook | When called | Default |
+|------|-------------|---------|
+| `on_broken_link(hdf_path, link)` | A soft or external link cannot be resolved. The broken node is then skipped. | no-op |
+| `on_external_link(hdf_path, link)` | An external link is first encountered, *before* the handler opens the external file. | no-op |
+
+The `link` argument is an `h5py.SoftLink` or `h5py.ExternalLink` — inspect its type to distinguish the two cases.  Nodes from a successfully resolved external file are visited with the same `hdf_path` prefix as the link itself; for example, a link at `"entry/ext"` whose target is a group containing `"value"` results in `on_field("entry/ext/value", ...)`.
+
+!!! note "Hard links"
+    Hard links are followed transparently. A built-in cycle guard prevents infinite recursion when a hard-linked descendant aliases an ancestor.
+
 ## Worked example: collecting all field paths and values
 
 This visitor collects every field path and its scalar value into a flat dictionary.
@@ -132,6 +144,56 @@ logging.basicConfig(level=logging.INFO)
 visitor = SchemaAwareVisitor("NXarpes")
 NexusFileHandler("path/to/arpes_file.nxs").process(visitor)
 ```
+
+## Handling broken and external links
+
+Override `on_broken_link` and `on_external_link` to react to links that cannot be resolved or to external files that are traversed inline.
+
+```python
+from __future__ import annotations
+
+import h5py
+
+from pynxtools.nexus.handler import NexusFileHandler, NexusVisitor
+
+
+class LinkAwareVisitor(NexusVisitor):
+    """Report broken links and track which external files were opened."""
+
+    def __init__(self) -> None:
+        self.broken: list[tuple[str, str]] = []   # (hdf_path, link_target)
+        self.external_files: set[str] = set()
+
+    def on_group(self, hdf_path: str, hdf_node: h5py.Group) -> None:
+        pass
+
+    def on_field(self, hdf_path: str, hdf_node: h5py.Dataset) -> None:
+        pass
+
+    def on_attribute(self, hdf_path: str, attr_name: str, attr_value, parent) -> None:
+        pass
+
+    def on_complete(self, root: h5py.File) -> None:
+        pass
+
+    def on_broken_link(self, hdf_path: str, link) -> None:
+        target = getattr(link, "path", str(link))
+        self.broken.append((hdf_path, target))
+        print(f"BROKEN LINK at {hdf_path!r} → {target!r}")
+
+    def on_external_link(self, hdf_path: str, link: h5py.ExternalLink) -> None:
+        self.external_files.add(link.filename)
+        print(f"External file opened: {link.filename!r} for node {hdf_path!r}")
+
+
+# Usage
+visitor = LinkAwareVisitor()
+NexusFileHandler("path/to/file.nxs").process(visitor)
+print(f"Broken links: {visitor.broken}")
+print(f"External files referenced: {visitor.external_files}")
+```
+
+`on_broken_link` is called for both broken soft links (`h5py.SoftLink`) and broken external links (`h5py.ExternalLink`); inspect `type(link)` to distinguish them.  `on_external_link` fires only for external links, immediately before the external file is opened — it does not fire again for nodes inside the external subtree.
 
 ## Resolving the application definition from the file
 
--- a/docs/how-tos/pynxtools/implement-a-visitor.md
+++ b/docs/how-tos/pynxtools/implement-a-visitor.md
@@ -41,7 +41,7 @@ Two additional optional hooks have default no-op implementations and may be over
 | `on_broken_link(hdf_path, link)` | A soft or external link cannot be resolved. The broken node is then skipped. | no-op |
 | `on_external_link(hdf_path, link)` | An external link is first encountered, *before* the handler opens the external file. | no-op |
 
-The `link` argument is an `h5py.SoftLink` or `h5py.ExternalLink` — inspect its type to distinguish the two cases.  Nodes from a successfully resolved external file are visited with the same `hdf_path` prefix as the link itself; for example, a link at `"entry/ext"` whose target is a group containing `"value"` results in `on_field("entry/ext/value", ...)`.
+The `link` argument is an `h5py.SoftLink` or `h5py.ExternalLink`. Inspect its type to distinguish the two cases. Nodes from a successfully resolved external file are visited with the same `hdf_path` prefix as the link itself; for example, a link at `"entry/ext"` whose target is a group containing `"value"` results in `on_field("entry/ext/value", ...)`.
 
 !!! note "Hard links"
     Hard links are followed transparently. A built-in cycle guard prevents infinite recursion when a hard-linked descendant aliases an ancestor.
@@ -193,7 +193,7 @@ print(f"Broken links: {visitor.broken}")
 print(f"External files referenced: {visitor.external_files}")
 ```
 
-`on_broken_link` is called for both broken soft links (`h5py.SoftLink`) and broken external links (`h5py.ExternalLink`); inspect `type(link)` to distinguish them.  `on_external_link` fires only for external links, immediately before the external file is opened — it does not fire again for nodes inside the external subtree.
+`on_broken_link` is called for both broken soft links (`h5py.SoftLink`) and broken external links (`h5py.ExternalLink`); inspect `type(link)` to distinguish them.  `on_external_link` fires only for external links, immediately before the external file is opened. It does not fire again for nodes inside the external subtree.
 
 ## Resolving the application definition from the file
 
--- a/src/pynxtools/dataconverter/validation.py
+++ b/src/pynxtools/dataconverter/validation.py
@@ -32,8 +32,6 @@ import logging
 import h5py
 import lxml.etree as ET
 import numpy as np
-from cachetools import LRUCache, cached
-from cachetools.keys import hashkey
 
 from pynxtools.dataconverter.helpers import (
     Collector,
@@ -58,8 +56,10 @@ from pynxtools.nexus.nexus_tree import (
     NexusField,
     NexusGroup,
     NexusNode,
+    _select_best_namefit,
     generate_tree_from,
 )
+from pynxtools.nexus.schema_resolver import NexusSchemaResolver, resolve_path
 from pynxtools.units import NXUnitSet, ureg
 
 logger = logging.getLogger("pynxtools")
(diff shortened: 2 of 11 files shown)


src/pynxtools/dataconverter/validation.py
```diff
@@ -800,141 +803,61 @@ def handle_attributes(
...
-    required_entities: set[str] = set()
-    update_required_concepts("", tree)
+        for req_concept in sorted(self._required_entities):
+            if any(req_concept.startswith(grp) for grp in self._required_groups):
+                continue
+            if "@" in req_concept:
+                if any(
+                    req_concept.rsplit("@", 1)[0].startswith(ent)
+                    for ent in self._required_entities
+                ):
```
RubelMozumder: StartsWith returns true for `sensor/@attr` and `sensorname/@attr`, which can be problematic.

src/pynxtools/nexus/schema_resolver.py
```diff
@@ -36,7 +36,7 @@
 def resolve_path(
     root: NexusNode,
     path: str,
-    node_type: Literal["group", "field"] | None = None,
+    node_type: Literal["group", "field", "attribute"] | None = None,
```
RubelMozumder, line 39: doc strings need to be updated and include the attribute.

RubelMozumder (approved):

RubelMozumder (approved):
LGTM!

src/pynxtools/nexus/schema_resolver.py
lukaspie: See the comment in the docstring

"""
Attributes are not resolved here because they always live on a parent node (group or field) and require knowing that parent's type.  Use :meth:`NexusSchemaResolver.attr_node_for` instead.
"""

So attribute are resolved in a different function.

src/pynxtools/dataconverter/validation.py
lukaspie: You are right, I replaced the `startswith` call with a path split and a check if a direct match exists.

docs/how-tos/pynxtools/implement-a-visitor.md
```diff
@@ -34,6 +34,18 @@ class MyVisitor(NexusVisitor):
...
 ```
 
+Two additional optional hooks have default no-op implementations and may be overridden:
+
+| Hook | When called | Default |
+|------|-------------|---------|
+| `on_broken_link(hdf_path, link)` | A soft or external link cannot be resolved. The broken node is then skipped. | no-op |
+| `on_external_link(hdf_path, link)` | An external link is first encountered, *before* the handler opens the external file. | no-op |
+
+The `link` argument is an `h5py.SoftLink` or `h5py.ExternalLink` — inspect its type to distinguish the two cases.  Nodes from a successfully resolved external file are visited with the same `hdf_path` prefix as the link itself; for example, a link at `"entry/ext"` whose target is a group containing `"value"` results in `on_field("entry/ext/value", ...)`.
```
mkuehbach:
```suggestion
The `link` argument is an `h5py.SoftLink` or `h5py.ExternalLink`. Inspect its type to distinguish the two cases. Nodes from a successfully resolved external file are visited with the same `hdf_path` prefix as the link itself; for example, a link at `"entry/ext"` whose target is a group containing `"value"` results in `on_field("entry/ext/value", ...)`.
```

docs/how-tos/pynxtools/implement-a-visitor.md
```diff
@@ -133,6 +145,56 @@ visitor = SchemaAwareVisitor("NXarpes")
...
+
+
+# Usage
+visitor = LinkAwareVisitor()
+NexusFileHandler("path/to/file.nxs").process(visitor)
+print(f"Broken links: {visitor.broken}")
+print(f"External files referenced: {visitor.external_files}")
+```
+
+`on_broken_link` is called for both broken soft links (`h5py.SoftLink`) and broken external links (`h5py.ExternalLink`); inspect `type(link)` to distinguish them.  `on_external_link` fires only for external links, immediately before the external file is opened — it does not fire again for nodes inside the external subtree.
```
mkuehbach:
```suggestion
`on_broken_link` is called for both broken soft links (`h5py.SoftLink`) and broken external links (`h5py.ExternalLink`); inspect `type(link)` to distinguish them.  `on_external_link` fires only for external links, immediately before the external file is opened. It does not fire again for nodes inside the external subtree.
```

src/pynxtools/dataconverter/validation.py
```diff
@@ -53,12 +51,15 @@
...
     NexusField,
     NexusGroup,
     NexusNode,
+    _select_best_namefit,
     generate_tree_from,
 )
+from pynxtools.nexus.schema_resolver import NexusSchemaResolver, resolve_path
 from pynxtools.units import NXUnitSet, ureg
 
 logger = logging.getLogger("pynxtools")
```
mkuehbach:
```suggestion
logger = logging.getLogger(__file__)
```

mkuehbach: ?

src/pynxtools/dataconverter/validation.py
```diff
@@ -170,179 +171,234 @@ def is_valid_unit_for_node(
...
+        parent: h5py.Group | h5py.Dataset,
+    ) -> None:
+        # Attributes are handled inside on_group / on_field via
+        # _handle_attributes so that the full attrs dict is available.
+        pass
 
-        hint_map: dict[str, str] = {"DATA": "signal", "AXISNAME": "axis"}
+    def on_complete(self, root: h5py.File) -> None:
+        """Emit validation errors for all required concepts not encountered."""
+        self._report_missing()
```
mkuehbach, line 276: _report(), not necessarily that always something is missing

src/pynxtools/nexus/handler.py
```diff
@@ -220,9 +305,27 @@ def _full_visit(
         for attr_name, attr_value in hdf_node.attrs.items():
             visitor.on_attribute(name, attr_name, attr_value, hdf_node)
 
-        # Recurse into children
+        # Recurse into children — inspect link types explicitly so that broken
+        # soft links and external links can be dispatched to the visitor rather
+        # than crashing (h5py returns None for a broken soft link in .items()).
         if isinstance(hdf_node, h5py.Group):
-            for ch_name, child in hdf_node.items():
+            for ch_name in hdf_node:
```
mkuehbach: child_name instead of ch_name

tests/nexus/test_handler.py
```diff
@@ -165,3 +165,113 @@ def on_complete(self, root) -> None:
...
+
+    assert len(visitor.broken_links) == 1
+    path, link = visitor.broken_links[0]
+    assert path == "entry/ext"
+    assert isinstance(link, h5py.ExternalLink)
+
+
+def test_handler_valid_external_link(tmp_path):
+    """A resolvable external link is traversed; on_external_link fires before traversal."""
+    ext_file = tmp_path / "ext.h5"
```
mkuehbach, line 258: remove the test files, yes they will be removed at the latest upon restarting the computer

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -149,6 +149,66 @@ def _xml_path_in_nxdl(elem: ET._Element) -> str:
...
+
+        if score > best_score:
+            if hint and hint_map.get(node.name) != hint:
+                continue
+            best_match = node
+            best_score = score
+
+        if score not in score_board:
+            score_board[score] = {"required": [], "recommended": [], "optional": []}
+        for constraint in ("optional", "required", "recommended"):
```
mkuehbach: would follow the order as in L191 (req, rec, opt) consistence also should result in the same dict state

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -149,6 +149,66 @@ def _xml_path_in_nxdl(elem: ET._Element) -> str:
     return "/" + "/".join(reversed(parts)) if parts else ""
 
 
+def _select_best_namefit(
+    name: str,
+    nodes: list["NexusNode"],
```
mkuehbach: TODO: list[NexusNode] when `pynxtools`will deprecate support for py3.10
https://devguide.python.org/versions/

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -149,6 +149,66 @@ def _xml_path_in_nxdl(elem: ET._Element) -> str:
...
+    hint: str | None = None,
+) -> Optional["NexusNode"]:
```
mkuehbach: see above

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -570,61 +634,19 @@ def best_child_for(
...
-                            f"{[children[idx].concept_path for idx in score_board[alternative_best_score][constraint]]} "
-                            f"constrained by {constraint}; indicates possible issues with nameTyping of "
-                            f"specific NeXus classes/concepts"
-                        )
-
-        return best_match
+        for node in children:
+            if node is not None and not node.variadic and name == node.name:
+                return node
+
```
mkuehbach, line 647: refactor `n`, `node`

mkuehbach (changes requested):

src/pynxtools/dataconverter/validation.py
lukaspie: Probably best practive, but the `pynxtools` logger has this defined

```
handler = logging.StreamHandler()
formatter = CustomFormatter("%(message)s")
handler.setFormatter(formatter)
logger.addHandler(handler)
logger.setLevel(logging.INFO)
```

So we would loose custom formatter. I suggest we address the logging in a separate PR. See also https://github.com/FAIRmat-NFDI/pynxtools/issues/292.

src/pynxtools/dataconverter/validation.py
lukaspie: Currently, `_report_missing` really only reports missing elements. So the function is named properly IMO.

We could refactor that the whole log only gets emitted `on_complete` in `_report`, but that is currently not the case

src/pynxtools/nexus/handler.py
lukaspie: changed

src/pynxtools/nexus/nexus_tree.py
lukaspie: added a TODO

src/pynxtools/nexus/nexus_tree.py
lukaspie: added a TODO

src/pynxtools/nexus/nexus_tree.py
lukaspie: Changed to use `node in node...` instead of `n for n`

tests/nexus/test_handler.py
lukaspie: Hm, not sure. I usually like to keep the artifacts in `tmp/` temporarily so that failed tests can be inspected more easily. These files are small anyway.

src/pynxtools/nexus/nexus_tree.py
lukaspie: okay, changed the order

src/pynxtools/dataconverter/validation.py
mkuehbach: ok

tests/nexus/test_handler.py
mkuehbach: maybe ok

mkuehbach (approved):
That a good first shot one should continue with, approve

PR #788, opened 2026-05-19
show `@target` attributes as `[NeXus link]` in annotator output

### LINKED ISSUE #697: `pynx read`: Recognize `@target` attributes for links
Currently, `read_nexus` does not recognize the `@target` attribute which is supposed to be written whenever a link is written. 
This leads to reports like this:
```
DEBUG - ===== ATTRS (//entry/reproducibility_indicators/cantilever_oscillator/phase_lock_loop/amplitude_excitation@target) (line: 7025)
DEBUG - value: /entry/instrument/cantilever_spm/cantilever_oscillator/phase_lock_loop/amplitude_excitation (line: 7026)
DEBUG - ===== GROUP (//entry/resolution_indicators [NXafm::/NXentry/NXcollection]): <HDF5 group "/entry/resolution_indicators" (2 members)> (line: 7174)
DEBUG - ===== ATTRS (//entry/resolution_indicators/amplitude_excitation@target) (line: 7255)
DEBUG - value: /entry/instrument/cantilever_spm/cantilever_oscillator/phase_lock_loop/amplitude_excitation (line: 7256)
DEBUG - @target - IS NOT IN SCHEMA (line: 7260)
```

It would be good if the tool can recognize links and in this case, report `@target` as documented (at least as long as https://github.com/nexusformat/definitions/pull/1558 is not resolved).

lukaspie:
Previously `@target` was silently suppressed (listed in `_SKIP_ATTRS` alongside `NX_class`). Now it is always emitted at DEBUG level as:
```
  @target = /canonical/path  [NeXus link]
```

The check runs before the in_schema lookup so the `[NeXus link]` label appears even if a schema node for 'target' is found. That is consistent with how the NIAC handles the `@target` attribute for links.

Fixed #697

### DIFF
--- a/src/pynxtools/annotator/annotator.py
+++ b/src/pynxtools/annotator/annotator.py
@@ -318,8 +318,8 @@ class Annotator(NexusVisitor):
             for src, values in node.get_inheritance_enums():
                 self._detail(det, "Enums", f"[{src}] {', '.join(values)}")
 
-    # Structural HDF5 attributes that carry no annotation value
-    _SKIP_ATTRS: frozenset = frozenset({"NX_class", "target"})
+    # Structural HDF5 attributes that carry no annotation value.
+    _SKIP_ATTRS: frozenset = frozenset({"NX_class"})
 
     def _annotate_attribute(
         self,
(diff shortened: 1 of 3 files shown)


RubelMozumder (approved):
LGTM!

Issue #677, opened 2025-07-30
Add different vscode configurations for debugging pynxtools, and its use with pynxtools-plugins


mkuehbach:
Suggestions for improving the code to yield more professionally configured debugging activities:
Came up in #675 suggest was to make https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/.vscode/settings.json?ref_type=heads#L61

PR #790, opened 2026-05-20
refactor: introduce NomadVisitor implementing NexusVisitor

lukaspie:
## Summary

Migrates the NOMAD archive population logic from `HandleNexus`-based callback traversal to the `NexusVisitor` / `NexusFileHandler` pattern.

#### `NomadVisitor(NexusVisitor)` — new class

`NexusParser._nexus_populate` (a callback that manually walked an HDF5 path string on every invocation) is replaced by `NomadVisitor`, a concrete `NexusVisitor` implementation driven by `NexusFileHandler`.

Key structural improvements:

- `on_group` eagerly navigates to (or creates) the NOMAD `MSection` for each group and stores it in `_sections: dict[hdf_path, MSection]`.  Subsequent `on_field` and `on_attribute` callbacks look up their parent section in O(1)instead of re-walking the full path.
- `sample_class_refs` tracking moves from `NexusParser` state into `NomadVisitor`, where it is naturally scoped to a single file traversal.
- Dead code removed: `get_sub_element_names`, `get_sub_elements`, `_clear_class_refs`, and the `MSubSectionList` import.
- `HandleNexus` dependency dropped from `NexusParser.parse`; replaced with a two-liner `NomadVisitor + NexusFileHandler`.

#### `NexusParser` — simplified

`parse()` now creates a `NomadVisitor` and hands it to `NexusFileHandler`. Post-traversal normalization (chemical formula, element set, metadata) is unchanged.

### What is NOT done (follow-up work)

The internal schema-resolution mechanism inside `NomadVisitor` is not yet aligned with the rest of the visitor codebase:

1. **`get_nxdl_doc` is still used** in `on_group`, `on_field`, and `on_attribute`. This function from `nexus.py` is slated for removal.
   It should be replaced by `NexusSchemaResolver` lookups — the same infrastructure used by `Annotator` and `ValidationVisitor`.

2. **`_populate_data` is still a unified callback** gated on an `attr` flag. Once `get_nxdl_doc` is replaced, this should split into two
   focused methods:
   - `_populate_field(node: NexusNode, hdf_node, current)` — field population
   - `_populate_attribute(node: NexusNode, attr_name, attr_value, current)` — attribute population

3. **`depth` / `nx_path` indexing** — `_populate_data` currently indexes into the XML element list returned by `get_nxdl_doc` via `nx_path[depth]`. Once replaced by `NexusNode`, the `depth` parameter and all `nx_path[-1].get("name")` / `nx_node.tag.endswith("group")` accesses disappear.

4. **`_to_section` takes an `ET.Element`** for the group node.  It should be updated to accept a `NexusNode` (`node.name` replaces
   `nx_node.attrib.get("name")` and `node.nx_class` replaces `nx_node.attrib["type"]`).

The latter will be the work of a immediate follow-up PR, to facilitate easier review.

### DIFF
--- a/src/pynxtools/nexus/nexus.py
+++ b/src/pynxtools/nexus/nexus.py
@@ -558,15 +558,18 @@ class HandleNexus:
         self.concept = concept
         self._handler = NexusFileHandler(nexus_file, is_open=is_open)
 
-    def process_nexus_master_file(self, parser):
-        """Process a nexus master file by processing all its nodes and their attributes."""
+    def process_nexus_master_file(self, parser=None):
+        """Process a nexus master file by processing all its nodes and their attributes.
+
+        The *parser* argument is ignored — the NOMAD parser now uses
+        ``NomadVisitor`` directly instead of the legacy callback interface.
+        """
         from pynxtools.annotator.annotator import Annotator
 
         visitor = Annotator(
             self.logger,
             documentation=self.documentation,
             concept=self.concept,
-            parser=parser,
         )
         self._handler.process(visitor)
 
--- a/src/pynxtools/nexus/nexus.py
+++ b/src/pynxtools/nexus/nexus.py
@@ -559,7 +559,7 @@ class HandleNexus:
         self._handler = NexusFileHandler(nexus_file, is_open=is_open)
 
     def process_nexus_master_file(self, parser=None):
-        """Process a nexus master file by processing all its nodes and their attributes.
+        """Process a NeXus/HDF5 master file by processing all its nodes and their attributes.
 
         The *parser* argument is ignored — the NOMAD parser now uses
         ``NomadVisitor`` directly instead of the legacy callback interface.
--- a/src/pynxtools/nomad/parsers/parser.py
+++ b/src/pynxtools/nomad/parsers/parser.py
@@ -61,6 +61,11 @@ def _to_group_name(nx_node: ET._Element):
     return grp_nm
 
 
+def _make_hdf_info(hdf_path: str, hdf_node: Any) -> dict[str, Any]:
+    """Create metadata describing an HDF node and its absolute path."""
+    return {"hdf_path": f"/{hdf_path}", "hdf_node": hdf_node}
+
+
 # noinspection SpellCheckingInspection
 def _to_section(
     hdf_name: str | None,
@@ -313,7 +318,7 @@ class NomadVisitor(NexusVisitor):
         # TODO: replace get_nxdl_doc with NexusSchemaResolver.node_for(hdf_path, hdf_node);
         # nxdef  → resolver.appdef_for(hdf_node)
         # nx_node (ET.Element) → NexusNode returned by the resolver
-        hdf_info = {"hdf_path": "/" + hdf_path, "hdf_node": hdf_node}
+        hdf_info = _make_hdf_info(hdf_path, hdf_node)
         _, nxdef, nxdl_path = get_nxdl_doc(hdf_info, doc=False)
         if nxdl_path is None or nxdl_path == "/":
             return
@@ -350,7 +355,7 @@ class NomadVisitor(NexusVisitor):
         """Populate the NOMAD Metainfo quantity for this HDF5 dataset."""
         # TODO: replace get_nxdl_doc with NexusSchemaResolver.node_for(hdf_path, hdf_node)
         # and call _populate_field(node, hdf_node, current) directly (no depth, no nx_path)
-        hdf_info = {"hdf_path": "/" + hdf_path, "hdf_node": hdf_node}
+        hdf_info = _make_hdf_info(hdf_path, hdf_node)
         _, nxdef, nxdl_path = get_nxdl_doc(hdf_info, doc=False)
         if nxdl_path is None or nxdl_path == "/":
             return
@@ -377,7 +382,7 @@ class NomadVisitor(NexusVisitor):
         if attr_name in self._SKIP_ATTRS:
             return
 
-        hdf_info = {"hdf_path": "/" + hdf_path, "hdf_node": parent}
+        hdf_info = _make_hdf_info(hdf_path, parent)
         # TODO: replace get_nxdl_doc with NexusSchemaResolver.attr_node_for(hdf_path, attr_name, parent)
         # and call _populate_attribute(node, attr_name, attr_value, current) directly
         req_str, nxdef, nxdl_path = get_nxdl_doc(hdf_info, doc=False, attr=attr_name)
@@ -527,6 +532,11 @@ class NomadVisitor(NexusVisitor):
                 return
 
             # Metainfo does not support precision higher than i8, u8, f8, c16
+            # TODO: is f2 supported ? maybe silently promote to f4 or f8, maybe downcast higher
+            # precision floating and complex to highest supported precision floating and complex respectively
+            # TODO: emit a warning in the case of hdf_node.dtype.kind in "fc" when hdf_node.dtype.itemsize < 4
+            # (e.g. when one is hitting the half-precision floats that were introduced with HDF5 2.0).
+            # TODO warn in case of facing arbitrary objects or structs"
             if hdf_node.dtype.kind in "iufc" and hdf_node.dtype.itemsize > 8:
                 self._logger.warning(
                     f"error while setting field {data_instance_name} in {current.m_def} precision {hdf_node.dtype.itemsize} too high for {field_name}"
@@ -540,6 +550,7 @@ class NomadVisitor(NexusVisitor):
                     if hdf_node.chunks is not None:  # iterate over hyperslabs (chunks)
                         field_stats = _get_field_stats_iuf_chunked(hdf_node)
                     else:  # load entire contiguous storage layout dataset at once
+                        # we suggest to use chunked storage to avoid these costly cases
                         field_stats = _get_field_stats_iuf_contiguous(hdf_node)
 
                     for suffix in FIELD_STATISTICS:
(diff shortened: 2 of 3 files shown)


src/pynxtools/nexus/nexus.py
```diff
@@ -558,15 +558,18 @@ def __init__(
         self.concept = concept
         self._handler = NexusFileHandler(nexus_file, is_open=is_open)
 
-    def process_nexus_master_file(self, parser):
-        """Process a nexus master file by processing all its nodes and their attributes."""
+    def process_nexus_master_file(self, parser=None):
+        """Process a nexus master file by processing all its nodes and their attributes.
```
mkuehbach:
```suggestion
        """Process a NeXus/HDF5 master file by processing all its nodes and their attributes.
```

src/pynxtools/nexus/nexus.py
```diff
@@ -558,15 +558,18 @@ def __init__(
...
+
+        The *parser* argument is ignored — the NOMAD parser now uses
```
mkuehbach, line 564: Remove the `parser` argument cuz no longer used.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -267,40 +270,164 @@ def _get_field_stats_iuf_contiguous(hdf_node) -> dict:
     return stats
 
 
-class NexusParser(MatchingParser):
+class NomadVisitor(NexusVisitor):
     """
-    NexusParser doc
+    NexusVisitor that populates a NOMAD archive from a NeXus/HDF5 file.
```
mkuehbach:
```suggestion
    Specialized NexusVisitor that populates a NOMAD archive from a NeXus/HDF5 file.
```

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -398,25 +527,19 @@ def _populate_data(
                 return
 
             # Metainfo does not support precision higher than i8, u8, f8, c16
-            # is f2 supported ? maybe silently promote to f4 or f8, maybe downcast higher
```
mkuehbach, line 401: Reformulate into TODO or extend L400 by the statement if f2 is supported or not, (Lauri should know for sure) but do not remove the comment completely

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -398,25 +527,19 @@ def _populate_data(
...
-            # precision floating and complex to highest supported precision floating and complex respectively
             if hdf_node.dtype.kind in "iufc" and hdf_node.dtype.itemsize > 8:
```
mkuehbach, line 542: Makes sense to add a # TODO here for emitting a warning in the case of hdf_node.dtype.kind in "fc" when hdf_node.dtype.itemsize < 4 (e.g. when one is hitting the half-precision floats that were introduced with HDF5 2.0).

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -398,25 +527,19 @@ def _populate_data(
...
                 self._logger.warning(
                     f"error while setting field {data_instance_name} in {current.m_def} precision {hdf_node.dtype.itemsize} too high for {field_name}"
                 )
                 return
 
-            # Metainfo also does not support unpacked arbitrary objects or struct
```
mkuehbach, line 409: do not remove this comment but rewrite it to e.g.
"# TODO warn in case of facing arbitrary objects or structs"
that comment should be moved to the other "Metainfo does not support block"

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -398,25 +527,19 @@ def _populate_data(
...
-            # also these should be caught
-
             field_stats: dict = {}  # stats built from finite iuf arrays
 
             if hdf_node.dtype.kind in "iuf":
                 if hdf_node.shape != ():  # non-scalar, compute stats
                     if hdf_node.chunks is not None:  # iterate over hyperslabs (chunks)
                         field_stats = _get_field_stats_iuf_chunked(hdf_node)
                     else:  # load entire contiguous storage layout dataset at once
-                        # we suggest to use chunked storage to avoid these costly cases
```
mkuehbach, line 419: the comment informs folks on the implications it has to inject contiguous data its essence should be kept but can be shortened

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -440,12 +559,9 @@ def _populate_data(
                             target_name=field_name + "[" + data_instance_name + "]",
                         )
                         return
-            # for all non iuf data use old approach, no stats for any of these
```
mkuehbach, line 443: shorten to "no stats for non-iuf data" but do not remove completely

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -688,28 +742,24 @@ def parse(
             import debugpy  # will connect to debugger if in debug mode
```
mkuehbach, line 758: move L742-744 on the right hand side to the top and commented out but keep essentially giving an advice how to debug the code for developers, including the comment on the breakpoint, either here or in the docs say in how-tos Hints on debugging and then this snippet with e.g. this parser.py as the example.

Both ideas are fine with me but not just removal.

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -267,40 +270,164 @@ def _get_field_stats_iuf_contiguous(hdf_node) -> dict:
...
+        section = _to_section(
+            group_name, nxdef_nomad, nx_node, parent_section, self._nx_root
+        )
+        self._collect_class(section)
+        section.m_set_section_attribute("m_nx_data_path", "/" + hdf_path)
+        section.m_set_section_attribute("m_nx_data_file", self._nxs_fname)
+        self._sections[hdf_path] = section
+
+    def on_field(self, hdf_path: str, hdf_node: h5py.Dataset) -> None:
+        """Populate the NOMAD quantity for this HDF5 dataset."""
```
mkuehbach:
```suggestion
        """Populate the NOMAD Metainfo quantity for this HDF5 dataset."""
```

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -267,40 +270,164 @@ def _get_field_stats_iuf_contiguous(hdf_node) -> dict:
...
+
+        nxdef_nomad = rename_nx_for_nomad(nxdef) if nxdef else None
+        parent_path = hdf_path.rsplit("/", 1)[0] if "/" in hdf_path else ""
+        current = self._sections.get(parent_path)
+        if current is None:
+            return
+
+        # depth after a full _nexus_populate walk of this field's path:
+        # one past the field name itself (matches post-loop depth in _nexus_populate)
+        depth = len(hdf_path.split("/")) + 1
```
mkuehbach, line 373: There should be a documentation about the expected naming convention of an hdf_path i.e. is there a preceding "/" or not

src/pynxtools/nomad/parsers/parser.py
```diff
@@ -267,40 +270,164 @@ def _get_field_stats_iuf_contiguous(hdf_node) -> dict:
...
+        hdf_path: str,
+        attr_name: str,
+        attr_value: Any,
+        parent: h5py.Group | h5py.Dataset,
+    ) -> None:
+        """Populate the NOMAD quantity attribute for this HDF5 attribute."""
+        if attr_name in self._SKIP_ATTRS:
+            return
+
+        hdf_info = {"hdf_path": "/" + hdf_path, "hdf_node": parent}
```
mkuehbach: that dict construction is used repetitively, move to a function?

mkuehbach (changes requested):
Reasonable

src/pynxtools/nexus/nexus.py
lukaspie: This is essentially a backwards compatiblity issue. With all my refactorings now, I am trying to keep it backwards compatible so that the next release still has the same public API. Probably overkill, but I want to make this as clean as possible. The whole class is deprecated now, so that will anyway go away soon.

src/pynxtools/nomad/parsers/parser.py
lukaspie: Sorry, I unintentionally removed that. I readded it as a TODO.

src/pynxtools/nomad/parsers/parser.py
lukaspie: Done, added `_make_hdf_info(hdf_path, hdf_node)`

src/pynxtools/nomad/parsers/parser.py
lukaspie: added TODO

src/pynxtools/nomad/parsers/parser.py
lukaspie: added TODO as suggested

src/pynxtools/nomad/parsers/parser.py
lukaspie: Added back in, not sure what happened with the comments

src/pynxtools/nomad/parsers/parser.py
lukaspie: readded + second line changed to TODO

src/pynxtools/nomad/parsers/parser.py
lukaspie: Let's just punt this, there is a whole issue on debugging, we can address that there #677. I readded the usage comment.

lukaspie (commented):
Thanks for the feedback @mkuehbach. I am not sure where I lost all the comments, but they are now back in.

src/pynxtools/nexus/nexus.py
mkuehbach: Ok for me as suggested, TODO if that might get lost

mkuehbach (approved):
Comments were addressed, two conversations remain open, you can address them and then resolve these.

src/pynxtools/nexus/nexus.py
lukaspie: The whole class has a TODO for removal, so I think that's fine.

src/pynxtools/nomad/parsers/parser.py
lukaspie: There's one when it is first introduced in line 297. I made that more explicit now.

lukaspie:
> Comments were addressed, two conversations remain open, you can address them and then resolve these.

Thank you, I addressed and resolved. Merging now.

PR #791, opened 2026-05-21
Support base classes in `generate_tree_from`, move NXDL helpers

lukaspie:
- `nexus_tree.generate_tree_from` is renamed from `appdef` parameter to `nxdl` and extended to handle base classes: for non-application definitions, children are traversed directly from the NXDL root element rather than looking for an `NXentry` group.

- `dataconverter.helpers.generate_template_from_nxdl` uses `nexus_tree.generate_tree_from` directly, implements a new walk to write the actual dataconverter template.

- Move NXDL XML helpers (`get_nxdl_root_and_path`, `get_all_parents_for`, `get_appdef_root`, `is_appdef, is_variadic`, `remove_namespace_from_tag`) and NeXus-to-Python type mappings (NEXUS_TO_PYTHON_DATA_TYPES, nx_char/int/…)
from dataconverter.helpers into nexus.utils, which is the correct layer.

- `dataconverter.helpers` re-exports all moved names for backward compatibility. nexus_tree.py now imports from nexus.utils directly, resolving the layering violation where a lower-level module imported from a higher-level one.

### DIFF
--- a/src/pynxtools/annotator/annotator.py
+++ b/src/pynxtools/annotator/annotator.py
@@ -384,9 +384,8 @@ class Annotator(NexusVisitor):
         attr = parts[1] if len(parts) > 1 else None
 
         # Bare class name (e.g. "NXbeam"): match groups by HDF5 NX_class attribute.
-        # generate_tree_from only supports application definitions, so full IS-A
-        # chain traversal for base classes (e.g. querying "NXobject" to find all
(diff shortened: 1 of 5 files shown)


RubelMozumder (approved):

RubelMozumder (approved):
LGTM! Functionality-wise, the code is working as expected. I also fixed the issue coming from spm.

PR #797, opened 2026-05-26
NeXus–NOMAD metainfo generator Phase 1

lukaspie:
## Summary

Introduces a **code generator** that reads the NeXus definitions through the `NexusNode` API and writes one importable Python file per NXDL base class. This is the first of five phases replacing the runtime XML→Section compilation in `schema.py` with a persistent, static, type-annotated Python schema. Everything is purely additive: the existing
`nexus_schema` entry point and `schema.py` are untouched.

---

## Motivation and long-term plan

The current `schema_packages/schema.py` compiles all 142 NXDL base classes and ~50 application definitions from raw XML at every NOMAD import. This means:

- No reusable Python schema classes for domain plugins to inherit from.
- Names leak into user-facing code (`start_time__field`, `ENTRY___default`).
- Generated schema cannot be inspected or extended by static analysis tools.
- No structured path from NeXus optionality (`required`/`recommended`/`optional`) to UI
  visibility or search indexing.
- Large monolithic schema package slows app-worker startup.

The full updated plan as well as Architecture decision records (ADRs) can be found in the `data-modeling` repo: https://github.com/FAIRmat-NFDI/data-modeling/tree/main/nexus-metainfo

### Planned phases

| Phase | Scope | Status |
|-------|-------|--------|
| **1** | Infrastructure + base class generation; `nexus_base_classes` entry point | **This PR** |
| 2 | Application + contributed definition generation; `nexus_applications` / `nexus_contributed` entry points | Planned |
| 3 | Bridge `schema.py` → thin shim delegating to `metainfo._package.build_package()` | Planned |
| 4 | Parser rewrite: annotation-based navigation replacing `section.more["nx_*"]` and `__field` names | Planned |
| 5 | Cleanup: remove XML-parsing code, `NexusBaseSection`, `NORMALIZER_MAP`, `BASESECTIONS_MAP` | Planned |
| N | `nomad-measurements` alignment: `XRayDiffraction(Xrd)`, `Transmission`, etc. | Parallel to 2–5 |

---

### Infrastructure (new files, no changes to existing code)

| File | Purpose |
|------|---------|
| `nomad/annotations.py` | Three `AnnotationModel` subclasses (`NeXusDefinition`, `NeXusGroup`, `NeXusQuantity`) registered at plugin load time; no NOMAD core change required |
| `nomad/converters/_naming.py` | Name conversion utilities, base section mapping, NX-type→Python-type mapping |
| `nomad/converters/nxdl_to_metainfo.py` | Generator: reads `NexusNode` tree, builds Jinja2 context, renders, formats with ruff, writes additive-only Python source |
| `nomad/converters/templates/base_class.py.j2` | Jinja2 template for generated base class files |
| `nomad/converters/cli.py` | `pynx nomad generate-metainfo` CLI (`--nx-class`, `--all`, `--dry-run`, `--force`, `--output-dir`) |
| `nomad/metainfo/__init__.py` | Public API (`build_package`, `all_sections`) + `NexusBaseClassesEntryPoint` |
| `nomad/metainfo/_package.py` | Assembles all generated modules into a single NOMAD `Package` |
| `nomad/metainfo/base_classes/{entry,object,sample,data,electronanalyzer}.py` | Five representative generated files; remaining 137 generated by `pynx nomad generate-metainfo --all` |

### `nexus_tree.py` (minor, backward-compatible)

- `deprecated` attribute added to all `NexusNode` subclasses.

---

### 1. NeXus inheritance chain preserved in Python

Each generated class extends its **NeXus parent** (e.g. `Process(Object)`) and, when the
parent chain doesn't already provide a NOMAD basesection, adds it as a secondary base:

```python
class Object(BaseSection): ...                       # NXobject — root
class Entry(Object, basesections.Measurement): ...
class Process(Object, basesections.ActivityStep): ...
class Data(Object, basesections.ActivityResult): ...
class Sample(Component, basesections.CompositeSystem): ...
```

The NOMAD basesection is imported as a **module** (`from nomad.datamodel.metainfo import
basesections`) to avoid name collisions when the generated NeXus class and NOMAD base
share the same Python name:

```python
# Both the generated NeXus parent and the NOMAD base are named Component here
class SampleComponent(Component, basesections.Component): ...
```

### 2. Three annotation classes replace `section.more["nx_*"]`

```python
class NeXusDefinition(AnnotationModel):   # on top-level class m_def
    nx_class: str; category: str; ignore_extra_*: bool; symbols: dict; deprecated: str

class NeXusGroup(AnnotationModel):        # on named concept class m_def OR cross-file SubSection
    nx_class: str; name: str; name_type: str; optionality: str; min/max_occurs: int

class NeXusQuantity(AnnotationModel):     # on every Quantity
    kind: str; name: str; type: str; units: str; optionality: str; enumeration: list; ...
```

Registered via `AnnotationModel.m_registry` at plugin load time — no NOMAD core PR needed.
The `optionality` field (`required`/`recommended`/`optional`) is the foundation for future
UI visibility filtering and search index control (ADR-002, pending NOMAD core mechanism).

### 3. Annotation placement rule

| Context | Annotation location |
|---------|-------------------|
| Top-level class | `a_nexus_definition` on `m_def` |
| Named concept class (group with own quantities, same file) | `a_nexus_group` on its `m_def`; parent `SubSection` is **clean** |
| Cross-file group reference | `a_nexus_group` on the `SubSection` |

### 4. Named concept classes only when the group adds something

A named concept class (e.g. `EntryUser(User)`) is generated only when the group defines
quantities that differ from the generic class — changed optionality, extra fields, or
different type/units/enumeration. Groups that exist purely for occurrence context point
their `SubSection` directly at the generic class. This also prevents circular imports
(all classes extend `Object`, so generating named concepts for every group would create
`object.py` → `log.py` → `object.py` cycles).

### 5. Conflict resolution: Quantity vs SubSection name collisions

When the inheritance chain has both a `SubSection` and a `Quantity` with the same Python
name, the group wins the unqualified name and the conflicting field gets a `_field` suffix:

```python
# NXguide.data collides with inherited Object.data (SubSection) → renamed
data_field = Quantity(...)
```

---

## Generating the remaining base classes

Only five representative classes are committed in this PR to keep the diff reviewable.
The remaining 137 are generated identically by:

```bash
cd packages/pynxtools
pynx nomad generate-metainfo --all
```

The generator is reproducible and additive-only — running it again never removes or renames
existing members, preserving any hand-authored `normalize()` implementations.

---

## What is NOT in this PR (deferred)

| Item | Phase |
|------|-------|
| Remaining 137 generated base class `.py` files | Follow-up commit after regeneration |
| Package equivalence test vs. existing `schema.py` | Phase 1 tests |
| Application definitions (`metainfo/applications/`) | Phase 2 |
| Contributed definitions (`metainfo/contributed/`) | Phase 2 |
| `schema.py` bridge (thin shim) | Phase 3 |
| Parser migration to annotation-based navigation | Phase 4 |
| `metainfo_to_nxdl.py` round-trip exporter | Phase 5 |
| UI visibility filtering via `optionality` | Requires NOMAD core mechanism (ADR-002) |

---


docs/learn/pynxtools/nomad-integration.md
```diff
@@ -89,15 +89,18 @@ The key quantities indexed by the NeXus normalizer depend on the application def
...
+bundled with it are compiled from NXDL XML at import time and registered as NOMAD
+`Section` classes. This compilation is handled by `nomad/schema_packages/schema.py`.
 
-This means:
+The compiled schema makes every NeXus concept searchable in NOMAD:
 
-- Every NeXus field becomes a Metainfo quantity with the correct type and unit.
-- Every NeXus base class becomes a NOMAD section.
+- Every NeXus field becomes a Metainfo `Quantity` with the correct Python type and unit.
+- Every NeXus group becomes a `SubSection`.
```
RubelMozumder: What's about attributes?

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -273,6 +274,11 @@ class NexusNode(NodeMixin):
         "optional": ("optional", "recommended", "required"),
     }
 
+    def _set_deprecated(self) -> None:
+        """Read the ``deprecated`` attribute from the first element in the inheritance chain."""
+        if self.inheritance:
+            self.deprecated = self.inheritance[0].attrib.get("deprecated")
```
RubelMozumder, line 280: Why inheritance[0], why not the other ancestors?

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,119 @@
...
+    return nxdl_name
+
+
+def nxdl_to_subsection_name(nxdl_name: str) -> str:
+    """Convert an NXDL group name to a safe Python subsection name.
+
+    Variadic groups (nameType=any/partial) use the lowercase NXDL class name
+    without the NX prefix as the subsection name.
+    """
+    if nxdl_name in _RESERVED_QUANTITY_NAMES:
```
RubelMozumder, line 89: Why this check for suction. Would it be an issue if a subsection and field have the same name?

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
```diff
@@ -0,0 +1,1159 @@
...
+# ---------------------------------------------------------------------------
+# Build subsection context from a NXTreeGroup node
+# ---------------------------------------------------------------------------
+
+
+def _build_subsection_from_node(
+    node: NXTreeGroup,
+    section_fqn: str,
+    is_named_concept: bool = False,
+) -> SubSectionContext:
```
RubelMozumder, line 404: I think this is for those groups, which have been extended or nameded but not extended, in a base class or application definition. I think _build_name_concept whould also useed for such in-place non extended group. That is reagardless a NeXus group in base class or application definition has been extended or not, the group may come as subclass in the same python file (as happning for extented group). I see it has a few benifits:
 - Those empty classes are extendable in python class
 - Building metainfo from NXDL, changes would happen always in the same place (pre generated python class) (if in future some groups in nexus schema got extended)
 - All groups follows the same conceptual architecture
 - Back to the nxdl would be easier from the same architectural benifit

RubelMozumder: We have already discussed it.

src/pynxtools/nomad/metainfo/base_classes/data.py
```diff
@@ -0,0 +1,769 @@
...
+            kind="attribute",
+            name="auxiliary_signals",
+            type="NX_CHAR",
+            name_type="specified",
+            optionality="optional",
+        ),
+    )
+    default_slice = Quantity(
+        type=np.float64,
+        description=(
```
RubelMozumder, line 288: It would through an validation error if the array is [2,nP]. And the description says that the quantity can be described with a char as well.

src/pynxtools/nomad/metainfo/base_classes/data.py
```diff
@@ -0,0 +1,769 @@
...
+        description=("Axis label"),
+        a_nexus_quantity=NeXusQuantity(
+            kind="attribute",
+            name="long_name",
+            type="NX_CHAR",
+            name_type="specified",
+            optionality="optional",
+            parent_field="AXISNAME",
+        ),
+    )
```
RubelMozumder, line 431: Some form of validation is needed for the attributes of the field or group when the quantity is variadic. This validation can be added in normalization in all sections. I have not come across a quantity property (something like `repeats=True`) that tells quantity is repeatable

src/pynxtools/nomad/metainfo/base_classes/electronanalyzer.py
```diff
@@ -0,0 +1,680 @@
...
+            symbols={
+                "nfa": "Number of fast axes (axes acquired simultaneously, without scanning a\n                physical quantity)",
+                "nsa": "Number of slow axes (axes acquired while scanning a physical quantity)",
+                "n_transmission_function": "Number of data points in the transmission function.",
+            },
+        ),
+    )
+
+    energy_resolution = SubSection(
+        section_def="pynxtools.nomad.metainfo.base_classes.electronanalyzer.ElectronanalyzerEnergyResolution",
```
RubelMozumder, line 78: Import path would be an issue while importing the `ElectronanalyzerEnergyResolution` due to constraints on the public API by `__all__`.

src/pynxtools/nomad/metainfo/base_classes/electronanalyzer.py
```diff
@@ -0,0 +1,680 @@
...
+    resolution = Quantity(
+        type=np.float64,
+        dimensionality="[angle]",
+        a_nexus_quantity=NeXusQuantity(
+            kind="field",
+            name="resolution",
+            type="NX_FLOAT",
+            name_type="specified",
+            optionality="optional",
+            units="NX_ANGLE",
```
RubelMozumder, line 567: I did not find the `dimensionality` property of quantity.

https://nomad-lab.eu/prod/v1/develop/docs/howto/plugins/types/schema_packages.html#quantities

src/pynxtools/nomad/metainfo/base_classes/data.py
```diff
@@ -0,0 +1,769 @@
...
+            "can be provided."
+        ),
+        a_nexus_quantity=NeXusQuantity(
+            kind="field",
+            name="AXISNAME",
+            type="NX_CHAR_OR_NUMBER",
+            name_type="any",
+            optionality="optional",
+        ),
+    )
```
RubelMozumder, line 417: Can we define a NOMAD custom type for such a situation so that would accept the types `int`, `float,` or `array of object` type?

```python
    ...: from typing import Union
    ...: import numpy as np
    ...: 
    ...: MyType = Union[int, float, str, np.ndarray]
In [13]: x: ArrayType = np.array([1, "one"], dtype=object)
In [14]: isinstance(x, MyType)
Out[14]: True
```

src/pynxtools/nomad/metainfo/base_classes/sample.py
```diff
@@ -0,0 +1,905 @@
...
+        ),
+    )
+    sample_component = SubSection(
+        section_def="pynxtools.nomad.metainfo.base_classes.sample_component.SampleComponent",
+        repeats=True,
+        variable=True,
+        description=(
+            "One group per sample component This is the preferred way of "
+            "recording per component information over the n_comp arrays"
+        ),
```
RubelMozumder, line 113: Two `sample_component` for quantity and subsection under the same namespace.

RubelMozumder (approved):
I think the most interesting things would happen in phase-2. There are no tests yet. I would write some tests for this.

docs/learn/pynxtools/nomad-integration.md
lukaspie: added a sentence on attributes -> quantities

src/pynxtools/nexus/nexus_tree.py
lukaspie: Because a concept itself is only ever deprecated itself. There's no instance of a deprecated field that is the parent of another non-deprecated field.

src/pynxtools/nomad/converters/_mapping.py
lukaspie: Yes, that's unallowed in NOMAD. With the implementation now, Subsections take precedent, i.e., if there are a group and a field of the same name, the field becomes a Quantity like `fieldname_quantity`.

src/pynxtools/nomad/metainfo/base_classes/data.py
lukaspie: This is unfortunately the route we have to take now, as NOMAD does not support mixed types. So we basically map to float for `NX_CHAR_OR_NUMBER`. But we always did that.

src/pynxtools/nomad/metainfo/base_classes/data.py
lukaspie: Technically, we could, but that would need to proliferate through all of nomad. NOMAD does not support mixed types at all.

src/pynxtools/nomad/metainfo/base_classes/data.py
lukaspie: We can do all kinds of normalization later, yes. But I would also try to be not too strict and rather document the intended use.

src/pynxtools/nomad/metainfo/base_classes/electronanalyzer.py
lukaspie: You can still import this; even if it is not in `__all__`, you can still do `from pynxtools.nomad.metainfo.base_classes.electronanalyzer import ElectronanalyzerEnergyResolution`.  `__all__` does define the public API, aka which classes we always guarantee to be there, but it doesn't restrict what can import

src/pynxtools/nomad/metainfo/base_classes/electronanalyzer.py
lukaspie: It is there: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/nomad/metainfo/metainfo.py#L3119

src/pynxtools/nomad/metainfo/base_classes/sample.py
lukaspie: That's indeed unfortunate, as there's sample_component (non-variadic field) and `NXsample_component` (variadic group) inside `NXsample`. I'll have to make the naming schema more rigorous to catch this violation as well.

lukaspie: I fixed it now, also if there's two concept of the same name with different type inside a class, the group takes precedent and the field goes to `FIELD_quantity`.

docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -0,0 +1,555 @@
...
+
+The existing pynxtools NOMAD integration compiles all NXDL base classes and
+application definitions from raw XML at import time inside `nomad/schema_packages/schema.py`.
+This approach works but tightly couples NXDL parsing to schema registration, makes the
+generated schema hard to inspect or extend, and produces no persistent Python artifacts
+that domain plugins can build on.
+
+Phase 1 of the NeXus–NOMAD Metainfo refactor introduces a **code generator** that reads
+the NeXus definitions through the `NexusNode` API and writes one Python file per NXDL
+base class.  Each generated file is a normal importable Python module containing a single
```
mkuehbach:
```suggestion
base class. Each generated file is a normal importable Python module containing a single
```

docs/learn/pynxtools/nomad-integration.md
```diff
@@ -89,15 +89,21 @@ The key quantities indexed by the NeXus normalizer depend on the application def
...
+`pynxtools` is loaded as a plugin, all NeXus base classes and application definitions
+bundled with it are compiled from NXDL XML at import time and registered as NOMAD
+`Section` classes. This compilation is handled by `nomad/schema_packages/schema.py`.
+
+The compiled schema makes every NeXus concept searchable in NOMAD:
+
+- Every NeXus group becomes a `SubSection`.
+- Every NeXus field becomes a Metainfo `Quantity` with the correct Python type and unit.
+- Every NeXus attributes becomes a Metainfo `Quantity`.
+  - It it is the attribute of a field, it is separated from its parent with two underscores (e.g. ``AXISNAME__long_name``).
```
mkuehbach: This and the next sentence unclear needs human revision

src/pynxtools/nexus/nexus_tree.py
```diff
@@ -1023,9 +1029,21 @@ def _set_definition_attrs(self) -> None:
...
+        definitions_url = get_definitions_url()
+        doc_base = NX_DOC_BASES.get(
+            definitions_url, "https://manual.nexusformat.org/classes"
+        )
+        nx_file = self.nxdl_base.split("/definitions/")[-1].split(".nxdl.xml")[0]
+        anchor = self.name.lower()
+        return f"{doc_base}/{nx_file}.html#{anchor}"
+
     def __init__(self, **data) -> None:
+        self.symbols = {}
```
mkuehbach, line 1043: type annotations

src/pynxtools/nomad/converters/templates/base_class.py.j2
```diff
@@ -0,0 +1,377 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+# This file is AUTO-GENERATED from the NeXus definitions (NXDL).
+# Run `pynx nomad generate-metainfo --nxdl {{ nx_name }}` to regenerate.
+# Additive-only: the generator will never remove or rename existing members.
```
mkuehbach: `members`

being slightly more detailed I would prefer

 `class members`,
`NeXus concepts`

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+# ---------------------------------------------------------------------------
+# Base section mapping
+# ---------------------------------------------------------------------------
+
+# TODO: this should be done in code and
+# there should be a check that this does not get overwritten
+# Maps NXDL top-level class name → (Python class name, dotted import path)
+BASESECTIONS_MAP: dict[str, tuple[str, str]] = {
+    "NXobject": ("BaseSection", "nomad.datamodel.metainfo.basesections"),
+    "NXentry": ("Measurement", "nomad.datamodel.metainfo.basesections"),
```
mkuehbach, line 103: Is a DFT simulation by Area C considered a `Measurement` or just mapping to a `BaseSection` ?
We have application definitions which model pure computer simulations, this collides with the here made assumption

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+    "NXprocess": ("ActivityStep", "nomad.datamodel.metainfo.basesections"),
```
mkuehbach, line 104: `ActivityStep` so including also activities by computers?

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+    "NXsample": ("CompositeSystem", "nomad.datamodel.metainfo.basesections"),
+    "NXsample_component": ("Component", "nomad.datamodel.metainfo.basesections"),
+    "NXfabrication": ("Instrument", "nomad.datamodel.metainfo.basesections"),
```
mkuehbach, line 107: I do not agree with this. `NXinstrument` should be mapping here like line 107 does but not `NXfabrication`.

We have a clear conceptual distinction between fabrication, component, and instrument, and we should not dilute these, no everything that has an NXfabrication instance is because of this an Instrument or should be blindly considered as an instrument.

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+    "NXdata": ("ActivityResult", "nomad.datamodel.metainfo.basesections"),
```
mkuehbach, line 108: So also activity by a computer right, like results from an analysis program?

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+"""
+Naming utilities for converting NXDL names to Python / NOMAD conventions.
+"""
+
+from __future__ import annotations
+
+# NOMAD BaseSection quantity names that conflict with NXDL field names.
+# Fields with these names get a `_quantity` suffix to avoid shadowing.
+_RESERVED_QUANTITY_NAMES: frozenset[str] = frozenset(
+    {"name", "datetime", "lab_id", "description"}
```
mkuehbach, line 27: I thought that this `lab_id` concept from a `BaseSection` was planned for removal or had already been removed for good to `id` ? Why coming up then again here and not `id` ?

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+    "NX_INT": "np.int64",
+    "NX_UINT": "np.int64",  # NOMAD has no unsigned int; annotation preserves intent
+    "NX_NUMBER": "np.float64",
+    "NX_POSINT": "np.int64",
+    "NX_COMPLEX": "np.complex128",
+    "NX_CHAR": "str",
+    "NX_BOOLEAN": "bool",
+    "NX_BINARY": "Bytes",  # nomad.metainfo.data_type.Bytes
+    "NX_DATE_TIME": "Datetime",  # from nomad.metainfo.data_type
+    "NX_CHAR_OR_NUMBER": "np.float64",
```
mkuehbach, line 144: would that work if one gives a string?

src/pynxtools/nomad/converters/_mapping.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+# ---------------------------------------------------------------------------
+# NX type → NOMAD type string (as used in generated source code)
+# ---------------------------------------------------------------------------
+
+# Values are the string expressions written into generated .py files.
+# numpy types are written as np.<name> so the template must import numpy as np.
+NX_TYPE_TO_SOURCE: dict[str, str] = {
+    "NX_FLOAT": "np.float64",
```
mkuehbach, line 134: I agree that you made reasonable defaults here, the subtlety is though that a constraint like NX_INT does not imply any constraint on precision. Like mentioned here also the concept of unsigned ints is not a part of NOMAD. 

Sure this promoting everything to the practically most well known fattest type gives some security but it is also a fundamental waste on resources. Think a good evolution for NOMAD would be to ask if key tools like `elasticsearch` and `NOMAD` could take advantage from if people don't need the full precision they also don't get the full precision, AI field is here ahead of us with the their half precision types and so on.

But yeah that's a macro NOMAD performance optimization question.

src/pynxtools/nomad/converters/templates/nexus.py.j2
```diff
@@ -0,0 +1,377 @@
+#
+# Copyright The NOMAD Authors.
```
mkuehbach, line 2: This template is used to generate a derivative source code product from LGPL-licensed and Apache-2.0 code, so we need to reflect this in the license statement.

github.com/FAIRmat-NFDI/nexus_definitions just copied the LGPL from nexusformat/definitions.git, which just seem to have "chosen" something, changed last time 14years ago.

Unfortunately, the license statements on the NIAC repo are also not consistently pulled through for each XML file and do not detail which kind of LGPL, this is an urgent shortcoming that the NIAC needs to fix ASAP so that we can merge it in a say `pynxtools=0.15.1`.

**They should just make a clean SPDX license statement in their repo**.

But questioning ChatGPT on the situation:
"Summarize how for both LGPL versions a taking of source code snippets and relicensing under Apache-2.0 is possible" leaves the situation as questionable:

(ChatGPT response):
```
The short answer is that it generally is not possible to take LGPL-licensed source code snippets, incorporate them into your own source code, and then relicense those copied snippets under Apache-2.0 alone.

Here's the distinction:

Scenario	LGPL-2.1	LGPL-3.0
Link to an LGPL library from Apache-2.0 code	✔ Allowed	✔ Allowed
Write new code that uses the library's API	✔ Apache-2.0	✔ Apache-2.0
Copy source code from the LGPL library into your project	✘ Copied code remains LGPL	✘ Copied code remains LGPL
Modify LGPL source files	✘ Modified files remain LGPL	✘ Modified files remain LGPL
Why?

The LGPL is a copyright license. When you copy copyrightable portions of LGPL code into your own project, you create a derivative work of that code. The LGPL grants permission to do this only if the copied portions continue to be licensed under the LGPL (or, in some circumstances, the GPL). It does not grant permission to convert the copied code to Apache-2.0.

What you can do

You can:

keep your original files under Apache-2.0,
keep copied or modified LGPL files under the LGPL,
distribute both together in the same repository,
clearly identify which files are under which license.

This is a common and accepted arrangement.

Small snippets

Very short, non-creative fragments (for example, a few trivial lines implementing an obvious algorithm) may not be protected by copyright at all. Such fragments can often be reused without triggering the LGPL. However, there is no fixed line-count rule; copyright depends on whether the copied expression is sufficiently original.

Clean-room reimplementation

If you study the behavior of an LGPL implementation and then write your own implementation from scratch, without copying protected expression, your implementation can be licensed under Apache-2.0. Copyright protects the expression of the code, not the underlying ideas, algorithms, interfaces, or functionality.

So, in summary:

Using an LGPL library from Apache-2.0 code is compatible.
Copying LGPL source code into Apache-2.0 files does not make the copied code Apache-2.0; those portions remain subject to the LGPL.
Reimplementing the same functionality independently allows you to license your implementation under Apache-2.0.
```

mkuehbach: Point is:
- NIAC needs to clarify the LGPL kind, i.e. best the SPDX in practice.
- For `pynxtools` we were good so far because we sub-moduled the stuff in and this assured that within the boundaries of the submodule the LGPL kicked overtaking the default of the pynxtools repo 
- What changes now is that we amalgamate LGPL and Apache code together into persistent artifacts and then immediately state that we relicense these. Ironically and I am pretty without knowing or caring about the whereabouts this is what the dynamic injection avoided, even though legally at the point anywhere in main memory the running code was never fully Apache-2.0.

If ChatGPT is right the NIAC would need to expose a library.
I think its a really serious issue with the NeXus definitions.

NIAC should work towards making the definitions more permissive.

To show good intent and assure future credibility we should solve these issues before
creating a pynxtools and NOMAD release which does it this way.

Already the dynamic injection before was questionable as at that point the same license proliferation should had been fiddled out.

mkuehbach (changes requested):
First round of review comments,
Implementation looks professional, great work without doubt.

Issue with unclear license on NeXus and our side needs resolution.

src/pynxtools/nomad/converters/templates/nexus.py.j2
mkuehbach: So setting the `base_classes` and `applications` directories under the same `LGPL` from the nexus_definitions and modifying the snippet in the generator might do the trick to keep us moving,
but permissive NOMAD-style is something different, so evolution on the NIAC is IMHO the way forward.

On another note, it is always a reality that note everything is always as permissively usable and composable as Apache-2.0 and CC-BY-4.0 alikes.

src/pynxtools/nexus/nexus_tree.py
lukaspie: This is just setting the value, it is [already annotated](https://github.com/FAIRmat-NFDI/pynxtools/blob/a8676f60b836f14bd3261dccb1aa212fd523abf5/src/pynxtools/nexus/nexus_tree.py#L1004) as `symbols: dict[str, str] = {}`

src/pynxtools/nomad/converters/_mapping.py
lukaspie: As of now, still called `lab_id` in the [basesections v2](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/nomad/datamodel/metainfo/basesections/v2.py?ref_type=heads#L123). Can be an discussion point in a later phase.

src/pynxtools/nomad/converters/_mapping.py
lukaspie: The main reason to have it be a `Measurement` is that `Measurement` has `instruments` (to be renamed to `instrument`) and `samples` (to be renamed to `sample`), which fits very well to `NXentry`. I would rather live with this ambiguity for the few application definitions which model pure computer simulations then letting `NXentry` just from `BaseSection`. This would be really against for we are trying to build here in terms of integrating with the basesections.

As these application definitions inherit from `NXentry` themselves, they can also not blindly inherit from a `Simulation` class as well. 

Area C uses the `Simulation` class, but that is still not in the base sections yet (only in `nomad-simulations`).

lukaspie: Alternative would be that `NXentry` inherits from `BaseSection` and then every application definitions either inherits from `Measurement` or `Simulation`.

Then again, if we move this to `nomad-measurements`, it doesn't make too much sense to have `Simulation` classes in there...

src/pynxtools/nomad/converters/_mapping.py
lukaspie: This will be changed later anyway. Activity and their steps are meant to cover an  `action that has a temporal extension and for some time depends on some entity`. So that would include computations of course.

src/pynxtools/nomad/converters/_mapping.py
lukaspie: This whole map will be changed anyway, together with changes in the base sections. So far, it just copies the map that we had used in the old schema.

`class Instrument` in base section was always meant to be the static instrument in the lab (so very close to `instrument.fabrication` in NeXus, where the NeXus instrument is the instrument while an experiment took place. So there is a mismatch there anyway.

**In general, we will probably completely remove this map and change the inheritance manually on the classes. But this can only be done once the Python schema is there and stable.**

lukaspie: Then the generator would not even touch the inherited classes anymore.

src/pynxtools/nomad/converters/_mapping.py
lukaspie: See above

src/pynxtools/nomad/converters/_mapping.py
lukaspie: Cannot be discussed here, that is a NOMAD discussion.

docs/learn/pynxtools/nomad-integration.md
lukaspie: Just mixed up `It` and `If`, fixed now.

src/pynxtools/nomad/converters/templates/base_class.py.j2
lukaspie: changed

src/pynxtools/nomad/converters/_mapping.py
lukaspie: No, unfortunately not. We cannot really map `NX_CHAR_OR_NUMBER` into the NOMAD system, so we always (also with the old dynamic schema) just mapped it to `np.float`. If the value is string, it gets dropped with a warning. Not ideal, but also no good way around it..

src/pynxtools/nomad/converters/templates/nexus.py.j2
lukaspie: As discussed, needs to be discussed with NIAC.

lukaspie:
I am closing here in favor of #801. Any open comments have been either reposted verbatim or at least referenced in #801.

PR #798, opened 2026-05-29
fix types for ET._Element, remove unused imports

lukaspie:
The changes introduced in #750 led to an issue related to typing, see https://github.com/DTU-Nanolab-materials-discovery/nomad-oasis-image/actions/runs/26627492430/job/78467724408?pr=74#step:5:5130.

The actual issue is that `ET.Element` is not an actual type, but a factory function. `ET._Element` is the actual type.

Also removes imports from the `parser.py` module that are unused.


mkuehbach (approved):
lgtm, thanks for the fix

PR #792, opened 2026-05-21
Pure `NXdata` detection logic

lukaspie:
Move pure `NXdata` detection logic out of `annotator/nxdata.py` into a new `nexus/nxdata.py` module with no logging dependency, so both the annotator and the validator can reuse it.

- `nexus/nxdata.py`: **`NXdataInfo`** dataclass; `classify_field` (v3→v2→v1 attribute detection); `find_default_nxentry`; `find_default_nxdata` (default-chain walk); `inspect_nxdata` (returns structured `NXdataInfo` implementing all three plottable conventions); axis/signal helpers re-exported for backward compatibility

- `annotator/nxdata.py` removed, functionality directly implemented in annotator

- `nexus/nexus.py: `_process_node` and `get_default_plottable` (both unused) removed

- `validation.py`: `_handle_nxdata` uses `inspect_nxdata` for structured signal/axes/aux_signals/AXISNAME_indices validation; v1 heuristic excluded (guard on `@signal` attr preserved to avoid false positives).

- `tests/nexus/test_nxdata.py`: unit tests covering `classify_field`, `find_default_nxdata`, `find_default_nxentry`, and `inspect_nxdata` across all three detection conventions.

Fixes #519

PR #799, opened 2026-05-29
apply fixes from remaining comments from #792

lukaspie:
`NXdata` detection:
- `inspect_nxdata` v1 convention now matches the NeXus spec: signal detected via `signal="1"` field attribute with `axis=N` on axis fields
- v2 is `signal="1"` with `@axes` on the signal field and `primary=1` ordering; the wrong single-dataset heuristic (no spec basis) is removed; `convention=None` when no spec-defined convention is satisfied
- docs: module docstring and NXdataInfo.convention updated to describe the three conventions accurately

Validation:
- added a validation info if outdated conventions `v2`/`v1` are used and a warning if no convention is there
- fix missing / before `@` in `NXdataAxisMismatch` path for `AXISNAME_indices`

tests:
- v1 test updated to use `signal="1"` + `axis=N` as per spec
- added test for single-dataset-no-signal → `convention=None`
- added `auxiliary_signals` order preservation test
- added `classify_field` counter-example for absent `@signal` attr


mkuehbach (approved):
Reasonable

PR #800, opened 2026-06-01
Add focused nxdl_to_metainfo unit tests for conflict and reserved-name handling

RubelMozumder:
## Summary

Phase 1 of the NeXus–NOMAD metainfo refactor: a code generator that reads NXDL base-class definitions through the \`NexusNode\` API and writes one importable Python file per base class, each containing a structured NOMAD \`Section\` with typed annotations.

| Component | Path |
|---|---|
| Generator | \`src/pynxtools/nomad/converters/nxdl_to_metainfo.py\` |
| Jinja2 template | \`src/pynxtools/nomad/converters/templates/base_class.py.j2\` |
| Annotations | \`src/pynxtools/nomad/annotations.py\` — \`NeXusDefinition\`, \`NeXusGroup\`, \`NeXusQuantity\` |
| CLI | \`pynx nomad generate-metainfo [--nx-class NX_CLASS]\` |
| Generated files | \`src/pynxtools/nomad/metainfo/base_classes/\` — NXobject, NXdata, NXentry, NXsample, NXelectronanalyzer |
| Package assembly | \`src/pynxtools/nomad/metainfo/_package.py\` |
| Test fixture | \`src/pynxtools/data/NXtestBase.nxdl.xml\` — controlled NXDL covering all dtype/structure cases |
| Docs | \`docs/learn/pynxtools/nexus-metainfo-generation.md\` |

The existing \`nexus_schema\` entry point is untouched — everything is purely additive.

## Why

The current NOMAD integration compiles all NXDL definitions from raw XML at import time, producing no persistent Python artifacts. This makes the registered schema hard to inspect, extend, or build on from domain plugins.

Phase 1 decouples generation from loading. All NXDL reading goes through the \`NexusNode\` API (inheritance resolution, dtype mapping, units already handled there). The resulting \`.py\` files are normal importable modules: reviewable in Git, stable across imports, and extendable by downstream plugins.

## Validation

21 tests in \`tests/nomad/test_nxdl_to_metainfo.py\`:

- **Helper units**: \`_build_quantity_from_node\` (dtype/shape/enum/unit mapping for fields and attributes), \`_build_subsection_from_node\` (all three \`name_type\` cases), \`_base_from_extends\` (real \`NexusDefinition\` via \`generate_tree_from\`)
- **Rename logic**: field suffixed when shadowed by an ancestor subsection, at both direct and named-concept level; \`_ensure_conflicts_precomputed\` records suffix set without mutating nodes
- **Reserved names**: \`name\`, \`lab_id\`, \`description\` → \`_field\` suffix (3 parametrized cases)
- **Full pipeline**: NXtestBase written to \`tmp_path\`, structural assertions, valid Python (\`compile()\`); byte-identical to committed golden at \`tests/data/nomad/converter/testbase.py\`
- **NXtest / NXtest_extended**: section-by-section structural comparison (class bases + \`{member: Quantity|SubSection}\`) against committed goldens — robust to formatting and docstring changes
- **Regression**: NXobject, NXdata, NXentry, NXsample generator output byte-identical to committed \`.py\` files
- \`mypy\`, \`ruff\`, \`cspell\` all pass


lukaspie (changes requested):
Thanks for starting to put this together. We really need proper test coverage for the converter tooland the scenarios you chose (conflict suffixing, reserved names, subsection flags, dry-run I/O) are the right ones.

The main thing I'd suggest to change before merging is the `FakeField` / `FakeAttribute` / `FakeGroup` mock classes. They already diverge from the real API and will continue to do so unless rigorously maintained. A concrete mismatch already present:

```python
# FakeField:
optionality: bool = False   # ← bool

# Real NexusNode:
optionality: str            # "required" | "recommended" | "optional"
```

`_build_quantity_from_node` reads `node.optionality` as a string and writes it into `NeXusQuantity`. A test that passes a bool will not catch a regression here. More fundamentally, the mocks bypass `generate_tree_from` entirely — if a real NXDL file produces a node where the interaction between `dtype`, `shape`, and enum handling breaks, none of these tests catch it.

I am not quite sure why we need the mock classes in the first place? I would rather suggest to use `NXtest` and `NXtest_extended`. They cover the dtypes you need (`NX_INT`, `NX_FLOAT`, `NX_CHAR`, `NX_BOOLEAN`), units, enumerations, and all three `nameType` variants (`specified`, `partial`, `any`). Use them directly:

```python
from pynxtools.nexus.nexus_tree import generate_tree_from
from pynxtools.nomad.converters.nxdl_to_metainfo import _build_quantity_from_node

def test_quantity_dtype_and_shape():
    root = generate_tree_from("NXtest")
    field = next(c for c in root.children if c.name == "char_value")
    qty = _build_quantity_from_node(field)
    assert qty.python_type == "str"
    assert qty.unit is None
```

For `NexusGroup` specifically, direct instantiation also works — all its `__init__` methods guard on `if not self.inheritance` and return early with an empty chain, so no XML is needed:

```python
node = NexusGroup(nx_class="NXdetector", name="ENTRY", nx_type="group", name_type="specified")
```

If the conflict/reserved-name scenarios aren't covered by the existing test NXDL files, I think it's cleaner to add the minimal fields there rather than faking Python objects. That allows walking the full `XML → NexusNode → QuantityContext` path.

---

The two I/O and orchestration tests are fine as they are:

- `test_write_base_class_dry_run_detects_content_change`
- `test_generate_all_base_classes_counts_only_changed`

src/pynxtools/data/NXtestBase.nxdl.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lukaspie, line 1: Genuine question: why is this needed; what does it contain that is not covered by the existing `NXtest.nxdl.xml`?

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+        ignore_extra_fields=False,
+        ignore_extra_attributes=False,
+        symbols=[],
+        inheritance=[SimpleNamespace(attrib={"extends": "NXancestor"})],
+        nxdl_base="NXentry.nxdl.xml",
+        children=[field, current_group],
+        get_docstring=lambda depth=1: {"en": "Doc."},
+    )
+
+    monkeypatch.setattr(converter, "_conflicts_precomputed", True)
```
lukaspie, line 286: I really don't understand all this monkeypatching. Why can't we work with actual NXDL test files? Then we don't actually have to make monkeypatched versions of the converter, but we can test it end-to-end?

lukaspie: If needed, we should add to `NXtest` or, as you already did, create a new NXDL test file that has such group/field conflict in it.

Especially for the parent/child case, we can just add to `NXtest_extended`.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+    concept_qty_names = [
+        quantity.python_name for quantity in context["named_concepts"][0].quantities
+    ]
+
+    assert names == ["conflict_concept_field", "conflict_concept_field__units"]
+    assert context["subsections"][0].python_name == "conflict_concept"
+    assert concept_qty_names == ["conflict_concept_field"]
+
+
+def test_ensure_conflicts_precomputed_marks_ancestor_field_for_group_conflict(
```
lukaspie, line 323: Same here, why this monkeypatching everywhere?

lukaspie (changes requested):

src/pynxtools/data/NXtestBase.nxdl.xml
RubelMozumder: This is for testing conversion for nxdl corresponding to the base classes. The NXtest and NXtest_extended only represent the application definition.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: The idea of testing is to make it two layers:

**layer-1**: Functionality test, e.g., method and functions. In this case, monkeypatch resets the module-level variable to its pre-initialized state and then runs the test. With monkeypatch pytest features, two good unittest examples are `test_build_quantity_from_field_maps_transformation_unit (to take care of the NXtransformation unit)`, `test_build_context_renames_field_when_parent_subsection_conflicts (to take care of name conflict in inheritance chain)`, to add such edge cases, we do not need to add an end-to-end test.

**layer-2**: End-to-end test where a complete Python file has been generated and tested by the nxdl file, e.g., NXtest (appdef) and NXtestBase (base class). As here it is an end-to-end test, we do not need any mondeypatch here.

PR #802, opened 2026-06-09
add guideline on AI contributions

lukaspie:


mkuehbach (changes requested):
Good draft, makes sense to me reading, decision should be with NOMAD management and executive board though, whether we as "The NOMAD Authors" here agree on this policy. I support the here suggested pragmatic approach, specifically the statement in L160 I find good, raising the particular and the non-delegatable awareness on responsibility of humans on AI usage.

The "should" statement in my opinion is too weak to assure provenance of AI usage is collected. Instead, we should rather formulate it strict for now and then professionalize
the wording in the `docs` throughout to follow e.g. the semantics of "should, must, needs" from a standard e.g. https://datatracker.ietf.org/doc/html/rfc2119

The following of the standard can be a future work.
Making the formulation on the "should" stricter already may help guarding against misunderstandings.

lukaspie:
> The "should" statement in my opinion is too weak to assure provenance of AI usage is collected. Instead, we should rather formulate it strict for now and then professionalize the wording in the `docs` throughout to follow e.g. the semantics of "should, must, needs" from a standard e.g. [datatracker.ietf.org/doc/html/rfc2119](https://datatracker.ietf.org/doc/html/rfc2119)
> 

I changed the two "should" statement to "must" now

mkuehbach:
Push in priority, seek decision with the management, to me an essential PR to merge with the first `>0.15.0` releases.

docs/tutorial/contributing.md
```diff
@@ -142,6 +142,26 @@ A number of examples exist which document how the tools can be used. For a stand
...
+For now, contributors are expected to follow these guidelines:
+
+**Declaration of AI usage.**
+If AI tools were used in preparing a contribution, this must be stated in the pull request description. The declaration must name the tool(s) used and briefly describe the scope of their involvement (e.g. "Used GitHub Copilot for initial design of X" or "Used Claude Code to assist with refactoring Y").
+
+**Human authorship and responsibility.**
+All AI-generated content must be reviewed, understood, and validated by the contributing author before submission. The contributor is fully responsible for the correctness, quality, and suitability of the code, regardless of how it was produced. Contributions that are evidently fully AI-produced with no human involvement will not be accepted.
+
+**Code review.**
+AI tools may be used as a supplementary aid during code review, but final review decisions must always be made by a human maintainer. Automated or AI-driven review does not substitute for human judgment on design, correctness, or scientific validity.
```
mkuehbach, line 163:
```suggestion
AI tools may be used as a supplementary aid during code review, but final review decisions must always be made by a human maintainer. Automated or AI-driven review does not substitute for human judgement on design, correctness, or scientific validity.
```

mkuehbach (approved):
Provisional approve, provided minor typo gets fix.
Provisional as FAIRmat EB has the final say.
Personally; I think the statement is balanced now,
it is still allowing now the wild west, except, fully AI generated, AI auto generated and merged
Most important though is the statement clarifies responsibilities.

docs/tutorial/contributing.md
lukaspie: Most of our documentation uses American English (though we mix it quite regularly, unfortunately), also in the spellcheck. So I think we should leave it at `judgment`.

mkuehbach: > Most of our documentation uses American English (though we mix it quite regularly, unfortunately), also in the spellcheck. So I think we should leave it at `judgment`.

Accepted

lukaspie:
This was discussed last week as a reasonable approach, at least for now. We'll go with this right now and then we can iterate in the future.

PR #805, opened 2026-06-10
use decode_if_bytes function, and flatten 1-element arrays

### LINKED ISSUE #804: [Bug]: Validation fails on certain nexus files
### Contact Details

rettig@fhi-berlin.mpg.de

### What happened?

When trying to validate a nexus file generated by IgorPro, the validation terminates with an AttributeError.
File for testing: https://zenodo.org/records/8238531/files/1.nxs?download=1

### Relevant log output

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 validate("1.nxs")

File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validate_file.py:81, in validate(file, ignore_undocumented)
     78 if not is_hdf5(file):
     79     raise click.FileError(file, hint=f'"{file}" is not a valid HDF5 file.')
---> 81 def_map = _get_def_map(file)
     83 if not def_map:
     84     logger.warning(f"Could not find any valid entry in file {file}")

File /mnt/pcshare/users/Laurenz/AreaB/transformations_testing/pynxtools/src/pynxtools/dataconverter/validate_file.py:55, in _get_def_map(file)
     49     for entry_name, dataset in h5file.items():
     50         if (
     51             helpers.decode_if_bytes(dataset.attrs.get("NX_class")) == "NXentry"
     52             and f"/{entry_name}/definition" in h5file
     53         ):
     54             def_map.update(
---> 55                 {entry_name: h5file[f"/{entry_name}/definition"][()].decode("utf8")}
     56             )
     58 return def_map

AttributeError: 'numpy.ndarray' object has no attribute 'decode'
```

rettigl:
Possible fix for #804

### DIFF
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -1597,10 +1597,9 @@ def decode_if_bytes(value: Any, encoding: str = "utf-8") -> Any:
         if value.dtype.kind == "O":
             if value.size == 1:
                 return _decode_bytes_recursive(value[0], encoding)
-            else:
-                return np.vectorize(_decode_bytes_recursive, otypes=[object])(
-                    value, encoding
-                )
+            return np.vectorize(_decode_bytes_recursive, otypes=[object])(
+                value, encoding
+            )
         return value
 
     return _decode_bytes_recursive(value, encoding)
--- a/src/pynxtools/dataconverter/helpers.py
+++ b/src/pynxtools/dataconverter/helpers.py
@@ -1595,9 +1595,12 @@ def decode_if_bytes(value: Any, encoding: str = "utf-8") -> Any:
         if value.dtype.kind == "S":
             return np.char.decode(value, encoding)
         if value.dtype.kind == "O":
-            return np.vectorize(_decode_bytes_recursive, otypes=[object])(
-                value, encoding
-            )
+            if value.size == 1:
+                return _decode_bytes_recursive(value[0], encoding)
+            else:
+                return np.vectorize(_decode_bytes_recursive, otypes=[object])(
+                    value, encoding
+                )
         return value
 
     return _decode_bytes_recursive(value, encoding)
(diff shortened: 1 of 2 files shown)


src/pynxtools/dataconverter/helpers.py
```diff
@@ -1595,9 +1595,12 @@ def decode_if_bytes(value: Any, encoding: str = "utf-8") -> Any:
...
         if value.dtype.kind == "O":
-            return np.vectorize(_decode_bytes_recursive, otypes=[object])(
-                value, encoding
-            )
+            if value.size == 1:
+                return _decode_bytes_recursive(value[0], encoding)
+            else:
+                return np.vectorize(_decode_bytes_recursive, otypes=[object])(
+                    value, encoding
+                )
```
lukaspie:
```suggestion
            if value.size == 1:
                return _decode_bytes_recursive(value[0], encoding)
            return np.vectorize(_decode_bytes_recursive, otypes=[object])(
                value, encoding
            )
```
We can just skip the else here.

lukaspie (approved):
I think this is reasonable and it should not break anything. I checked the validation on the supplied file in #804 and it indeed works.

Is there any reason though why the `definition` field is a 1D vector instead of a simple string? Is that just how Igor does it?

rettigl:
> Is there any reason though why the `definition` field is a 1D vector instead of a simple string? Is that just how Igor does it?

The way Igor was used here was to save 1-element waves containing the string, which ends up in the corresponding 1-element arrays. Not sure if there would be a better way to do it in Igor. If you think this does not hurt, I would suggest to merge this change. 
@mkuehbach Any idea about the failing pynxtools-apm test?

lukaspie:
> > Is there any reason though why the `definition` field is a 1D vector instead of a simple string? Is that just how Igor does it?
> 
> The way Igor was used here was to save 1-element waves containing the string, which ends up in the corresponding 1-element arrays. Not sure if there would be a better way to do it in Igor. If you think this does not hurt, I would suggest to merge this change. 

Ok, thanks for the explanation, makes perfect sense. Let's go ahead with the merge.

> @mkuehbach Any idea about the failing pynxtools-apm test?

This is a known issue and there will be a fix very soon, so you can just ignore it for now

PR #807, opened 2026-06-23
Json map copy mode

rettigl:
This adds the functionality to generate a template compatible with the multiformat/json-map reader from an existiting hdf5 file. intended for reorganizing data in a file to conform with a given Nexus Definition.
It also extends the json-map reader to read attributes from hdf5 files. 
Code is mostly AI-generated


src/pynxtools/dataconverter/cli.py
```diff
@@ -6,7 +6,7 @@
 
 """CLI commands for NeXus data conversion and validation.
 
-Exposes two top-level symbols consumed by the ``pynx`` group:
+Exposes three top-level symbols consumed by the ``pynx`` group:
```
lukaspie:
```suggestion
Exposes top-level symbols consumed by the ``pynx`` group:
```
Let's remove the number as this gets out of touch all the time.

src/pynxtools/dataconverter/cli.py
```diff
@@ -15,6 +15,10 @@
 
 ``validate``
     Standalone command to validate a NeXus HDF5 file (``pynx validate``).
+
+``list_keys``
```
lukaspie, line 19: Would it make sense to bundle this under `pynx convert` as it is mainly used in the restructuring? Or do you want to keep this outside?

src/pynxtools/dataconverter/cli.py
```diff
@@ -80,6 +84,22 @@ def convert():
         )
 
 
+@click.command("list_keys")
```
lukaspie: Other commands are called `my-command-name`, so I suggest to go for `list-keys` instead.

src/pynxtools/dataconverter/cli.py
```diff
@@ -80,6 +84,22 @@ def convert():
...
+@click.argument(
+    "file_path", type=click.Path(exists=True, dir_okay=False, readable=True)
+)
+def list_keys(file_path: str):
+    """Print all HDF5 dataset and attribute paths as JSON.
+
+    The output mirrors the mapping returned by
+    ``pynxtools.dataconverter.helpers.list_hdf5_paths``.
+    It can be used in conjunction with the json_map reader
+    to restructure data from hdf5/nxs files.
```
lukaspie:
```suggestion
    to restructure data from HDF5 NeXus files.
```

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
```diff
@@ -0,0 +1,83 @@
+# Restructure data inside hdf5/nexus files
+
+!!! info
+    This how-to shows a practical workflow for restructuring existing HDF5/NeXus data without rewriting the underlying file manually. The starting point is the list of available keys generated by `pynx list_keys`, followed by a second conversion pass using the `json_map` reader.
+
+Often, a file contains the right data, but in a different hierarchy or with different names than the target NeXus application definition expects. Instead of constructing a config file by hand to rewrite the data into a valid NeXus tree, pynxtools allows extracting the HDF5 tree structure from an existing hdf5/nxs file and create a mapping that allows you to reconstruct the original file using the `pynx convert` command. By modifying this template config file, the data can be placed into the desired concepts.
```
lukaspie:
```suggestion
Often, a file contains the right data, but in a different hierarchy or with different names than the target NeXus application definition expects. Instead of constructing a config file by hand to rewrite the data into a valid NeXus tree, pynxtools allows extracting the HDF5 tree structure from an existing `.hdf5`/`.nxs` file and create a mapping that allows you to reconstruct the original file using the `pynx convert` command. By modifying this template config file, the data can be placed into the desired concepts.
```

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
```diff
@@ -0,0 +1,83 @@
...
+```json
+{
+  "/ENTRY[entry]/SAMPLE[sample]/temperature": "@data:entry/measurement/temperature",
+  "/ENTRY[entry]/SAMPLE[sample]/temperature/@units": "@data:entry/measurement/temperature/@units",
+  "/ENTRY[entry]/INSTRUMENT[instrument]/detector/count_time": "@data:entry/measurement/count_time",
+  "/ENTRY[entry]/INSTRUMENT[instrument]/detector/count_time/@units": "ms"
+}
+```
+
+This tells the reader to put the selected data under different NeXus concepts. If the key is changed, the data is moved to a different location in the target file. If the value is changed, the same concept gets populated from a different source field or constant.
```
lukaspie:
```suggestion
This tells the dataconverter to put the selected data under different NeXus concepts. If the key is changed, the data is moved to a different location in the target file. If the value is changed, the same concept gets populated from a different source field or constant.
```
or `json_map` reader

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
```diff
@@ -0,0 +1,83 @@
+# Restructure data inside hdf5/nexus files
```
lukaspie: This files is not registered in `docs/index.md`. We should probably rethink if we want to include everything there on the main page, but before we do so, please add it there as well.

src/pynxtools/cli.py
```diff
@@ -71,4 +71,5 @@ def pynx():
 pynx.add_command(validate, name="validate")
 pynx.add_command(generate_eln, name="generate-eln")
 pynx.add_command(inspect_appdef, name="inspect-appdef")
+pynx.add_command(list_keys, name="list_keys")
```
lukaspie: See above, can/should this go under `convert`? Or possibly even under a new group `inspect`?

src/pynxtools/dataconverter/cli.py
lukaspie: Maybe even better would be to introduce a new CLI group `inspect`, which has `list-keys` as a subcommand? That would allow futher file inspection capbilities later.

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
```diff
@@ -0,0 +1,83 @@
...
+  "/ENTRY[entry]/SAMPLE[sample]/temperature/@units": "@data:entry/sample/temperature@units"
+}
+```
+
+The keys in this mapping correspond to the Nexus concepts and paths as defined in the config files for the [MultiFormatReader](../../learn/pynxtools/multi-format-reader.md). The values are the source paths inside the HDF5 file, using the `@data:` prefix that the JSON mapping reader provides.
+
+If you need to keep the output for editing, save it to a file:
+
+```bash
+pynx list_keys my_file.nxs > keys.json
```
lukaspie: Maybe this can be an option `--output`? Similar to how  `pynx convert generate-template` does it.

lukaspie (changes requested):
The core functionality seems useful and works as well. I added some comments with respect to maintnenance and to the CLI command.

I also a commit which adds some comments and a couple tests to clarify that root-level commands are not showing in the `list_keys` output. Note that the functionality for attribute extraction was also changed, in line with the changes made in #839.

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
rettigl: This is totally fine, I also added a function actually, to create a file from a jupyter notebook

src/pynxtools/dataconverter/cli.py
rettigl: This was a fist idea, we can certainly iterate that.

src/pynxtools/cli.py
rettigl: I would keep list-keys for now, and maybe restructure later

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
rettigl: done

docs/how-tos/pynxtools/restructure-data-inside-hdf5-nexus-files.md
rettigl: added

src/pynxtools/dataconverter/cli.py
rettigl: I kept it now there. I am not sure it makes sense to add subgroup commands?

rettigl (commented):
Thanks, I added the review suggestions. I could not actually build and test the docs, how is this done?
mkdocs build fails with error 

`mkdocs_click._exceptions.MkDocsClickException: Module 'pynxtools.cli' has no attribute 'nomad'`

even though pynxtools is installed with [docs,nomad,dev]

rettigl:
@lukaspie the writing of root-level attributes actually does not seem to work anyways. Somehow, they are filtered out here: https://github.com/FAIRmat-NFDI/pynxtools/blob/json_map_copy_mode/src/pynxtools/dataconverter/validation.py#L2147

The template:

`  "/@test":" test",`

generates:

```
WARNING: There were attributes set for the group or field , but the group or field does not exist.
WARNING: The attribute /@test will not be written.
```

src/pynxtools/cli.py
lukaspie: okay

src/pynxtools/dataconverter/cli.py
lukaspie: My idea was to occupy the `inspect` namespace already for future capabilities, but I guess we can always refactor later. Let's leave it as-is.

lukaspie:
> Thanks, I added the review suggestions. I could not actually build and test the docs, how is this done? mkdocs build fails with error
> 
> `mkdocs_click._exceptions.MkDocsClickException: Module 'pynxtools.cli' has no attribute 'nomad'`
> 
> even though pynxtools is installed with [docs,nomad,dev]

Indeed, that did not work. Not sure what was going on there, as the main docs deployment works. Anyway, I fixed it now. 

> @lukaspie the writing of root-level attributes actually does not seem to work anyways. Somehow, they are filtered out here: >https://github.com/FAIRmat-NFDI/pynxtools/blob/json_map_copy_mode/src/pynxtools/dataconverter/validation.py#L2147
>
>The template:
>
>`  "/@test":" test",`
>
>generates:
>
>```
>WARNING: There were attributes set for the group or field , but the group or field does not exist.
>WARNING: The attribute /@test will not be written.
>```

Yeah, that's right, as they have no parent, they are filtered here; never thought about that. I fixed this now and added a test.

For simplicity reasons, I did this all on this branch, hope that works for you.

lukaspie (approved):
LGTM, thanks for the implementation

You can merge with the failing `pynxtools-ellips` test, that is unrelated.

rettigl:
> LGTM, thanks for the implementation
> 
> You can merge with the failing `pynxtools-ellips` test, that is unrelated.

Great, thanks, will close and merge here.

PR #809, opened 2026-06-29
bump definitions, following minor name fix 

lukaspie:
See https://github.com/FAIRmat-NFDI/nexus_definitions/pull/429. This also has a large number of changed `nyaml` files, where we applied the keyword `\`-prefix (see https://github.com/FAIRmat-NFDI/nexus_definitions/pull/427)


RubelMozumder (approved):
LGTM!

PR #801, opened 2026-06-08
NeXus–NOMAD metainfo generator: Phases 1 + 2

lukaspie:
## NeXus–NOMAD metainfo generator: Phases 1 + 2

This PR replaces [#797](https://github.com/FAIRmat-NFDI/pynxtools/pull/797) (which is being closed) and now targets `main` directly. It combines what was previously split into Phase 1 (infrastructure + base classes) and Phase 2 (application + contributed definitions) into a single review, so everyone can review the whole generator and its output together rather than in two stacked, hard-to-review-independently pieces.

For the full plan, all phases, and the Architecture Decision Records, see the `data-modeling` repo: https://github.com/FAIRmat-NFDI/data-modeling/tree/main/nexus-metainfo

---

### Summary

Introduces a **code generator** that reads the NeXus definitions through the `NexusNode` API and writes one importable Python file per NXDL class — both base classes (`category="base"`) and application/contributed definitions (`category="application"`) — replacing the runtime XML→`Section` compilation in `schema.py` with a persistent, static, type-annotated Python schema.

Everything is purely additive: the existing `nexus_schema` entry point and `schema.py` are untouched. Two new entry points are added: `nexus_base_classes` (195 base classes) and `nexus_applications` (85 application/contributed definitions, discovered by NXDL `category`, not by file path — see the note on contributed definitions below).

---

### Motivation

The current `schema_packages/schema.py` compiles all NXDL base classes and application definitions from raw XML at every NOMAD import. This means:

- No reusable Python schema classes for domain plugins to inherit from.
- Names leak into user-facing code (`start_time__field`, `ENTRY___default`).
- Generated schema cannot be inspected or extended by static analysis tools.
- No structured path from NeXus optionality (`required`/`recommended`/`optional`) to UI visibility or search indexing.
- Large monolithic schema package slows app-worker startup.

### Where this fits in the overall plan

| Phase | Scope | Status |
|-------|-------|--------|
| **1 + 2** | Infrastructure, base class generation, application + contributed definition generation | **This PR** |
| 3 | New parser (`nexus_parser_v2`) + new NeXus search app (`nexus_app_v2`), additive alongside the existing ones | Separate PR, in progress |
| 4 | Migrate `pynxtools-*` plugin apps to the v2 schema/parser | Separate PR, in progress |
| 5 | Port `NORMALIZER_MAP` logic to `normalize()` methods on the generated classes | Planned |
[...]

PR #810, opened 2026-06-29
NeXus–NOMAD metainfo generator: Phase 3 (parser v2 + app v2)

lukaspie:
Builds on [#801](https://github.com/FAIRmat-NFDI/pynxtools/pull/801) (Phases 1 + 2). This branch is fully rebased onto main after merging #801. For the full plan, all phases, and the Architecture Decision Records, see the `data-modeling` repo: https://github.com/FAIRmat-NFDI/data-modeling/tree/main/nexus-metainfo (specifically [here](https://github.com/FAIRmat-NFDI/data-modeling/blob/main/nexus-metainfo/nexus-metainfo-integration-plan.md#phase-3--new-parser--new-app-v2-complete--superseded-the-original-bridge-schemapy-plan-below) for Phase 3 implementation).

---

### Summary

Phase 1+2 generated the static Python metainfo (base classes + applications). This PR adds a new parser (`nexus_parser_v2`) that reads .nxs files directly into the generated Python classes via their `a_nexus_*` annotations, and a new search app (`nexus_app_v2`) that mirrors the old app, but uses the new names in the schema. Both are currently purely additive, new entry points; nothing existing is touched or removed (for now).

### Testing this branch

To test with the new schema, install `pynxtools` from this branch and add these entry points to your `nomad.yaml` (in addition to or instead of the existing `nexus_schema`/`nexus_parser`/`nexus_app`):

```yaml
plugins:
  entry_points:
    include:
      - "pynxtools.nomad.metainfo:nexus_base_classes"
      - "pynxtools.nomad.metainfo:nexus_applications"
      - "pynxtools.nomad.parsers:nexus_parser_v2"
      - "pynxtools.nomad.apps.app_v2:nexus_app_v2"
```

Upload any NeXus (`.nxs`/HDF5) file as usual; `nexus_parser_v2` matches it and produces archives using the generated classes (`Entry`, `Xps`, `Arpes`, …). The parsed data is then browsable in the "NeXus v2" app.

---

#### 1. One archive per `NXentry`, plus one `Root` archive per file

A NeXus file can hold multiple `NXentry` groups. Each becomes its own archive: `archive.data` is the `Entry`/`Xps`/`Arpes`/… instance directly.  There is no wrapper section, unlike parser v1's `Root` container. A _separate_ **`Root(Experiment)`** archive is also produced per file, populated from the file's own NXroot attributes (`file_name`, `file_time`, `creator`, `NeXus_version`, …) and linking every `NXentry` archive in that file as an `ExperimentStep` (resolved lazily via `m_context` + `generate_entry_id`, since sibling entries aren't necessarily already persisted). `NXroot` is mapped to `basesections.Experiment` in `BASESECTIONS_MAP` for this. Entry names are disambiguated as `"{file stem} - {entry name}"` so that `entry`/`entry1`-named groups (the overwhelming common case) stay distinguishable across files; the Root archive itself is named `"{file stem} (NeXus file)"`.

#### 2. Annotation-driven matching replaces XML/NXDL resolution at parse time

`parser_v2.py` does not re-derive the schema from NXDL at parse time at all. Instead, it it builds a `_SectionIndex` per generated `Section` class once, from the `a_nexus_group`/`a_nexus_field` annotations already baked into the Phase 1+2 classes (`NeXusGroup.nx_class → SubSection`, field concept name → `Quantity`), and uses `NexusSchemaResolver` only for HDF5-side concept matching (variadic names, `NXdata` signal/axes hints). Due to the new schema, paths read more naturally (e.g. `data.instrument[0].energy_resolution.resolution`, not `data.ENTRY[0].INSTRUMENT[0].energy_resolution.resolution__field`).

#### 3. A new search app, `nexus_app_v2`

Separate from the existing NeXus app, but with the same layout. The columns/menu/dashboard are built against the v2 paths (no `__field`/`__group` suffixes). Locked to `section_defs.definition_qualified_name = pynxtools.nomad.metainfo.base_classes.Entry`, so it only ever shows v2-parsed entries (and the application definitions that inherit from `Entry`).

#### 4. Link quantities and conflict-renamed quantities are populated correctly at parse time

Two parser-level fixes needed for the typed-link resolution and BaseSection-collision renaming (both already generator-side in #801) to actually produce correct values: a `NeXusLink` quantity whose target type was resolved to something other than `str` is now populated like a regular field (h5py already dereferences the link transparently) instead of storing the raw HDF5 link path as a string. And a quantity renamed for a `BaseSection` collision (e.g. `name_quantity` for an NXDL field literally named `name`) now also mirrors its value into the NOMAD-native quantity it shadows (e.g. `name`) via a `_SectionIndex.shadow_map`, so both the NXDL-faithful path and the BaseSection-native path (used by `normalize()`, search, etc.) carry the value.

#### 5. `results.material` / chemical formula normalization wired up for v2 archives

`Entry`/`Root` instances now populate `results.material` (chemical formula, elements) during normalization, matching parser v1's behavior. This is needed for the new app's "Elements" menu (periodic table, formula filters) to return anything.

---

### What is NOT in this PR (deferred, see [#801](https://github.com/FAIRmat-NFDI/pynxtools/pull/801)'s phase table)

| Item | Phase |
|------|-------|
| Migrating `pynxtools-*` plugin apps to the v2 schema/parser | 4 |
| Field statistics | 4 |
| Porting `NORMALIZER_MAP` logic to `normalize()` methods, changes to the basesections | 5 |
| `metainfo_to_nxdl.py` round-trip exporter | 6 |
| Removing the old `schema.py`/`parser.py`/`NexusBaseSection`/`NORMALIZER_MAP` | 7 |


tests/nomad/test_parsing_opt_v2.py
mkuehbach, line 1: Note for myself, compare against v1, rm comment when done

tests/nomad/test_parsing_v2.py
```diff
@@ -0,0 +1,449 @@
...
+    assert source.nx_name == "source"
+    # good ENUM - x-ray
+    assert source.probe == "x-ray"
+    # wrong inherited ENUM - Burst (accepted for open enum)
+    assert source.mode == "Burst"
+    # wrong inherited ENUM for extended field - 'Free Electron Laser' (accepted for open enum)
+    assert source.type == "Free Electron Laser"
+
+    # Unit check
+    assert instrument.analyser.entrance_slit_size == ureg.Quantity("750 micrometer")
```
mkuehbach: if `entrance_slit_size` datatype is float, i.e. 750. check against tolerance, not ==

tests/nomad/test_parsing_v2.py
```diff
@@ -0,0 +1,449 @@
...
+
+    # Data
+    assert hasattr(arpes_entry, "data")
+    assert len(arpes_entry.data) >= 1
+    data = arpes_entry.data[0]
+    assert data.__dict__.get("nx_name") == "data"
+    assert data.nx_name == "data"
+
+    assert len(data.AXISNAME) == 3
+    # there is still a bug in the variadic name resolution, so skip these
```
mkuehbach: "these", which ?

src/pynxtools/nomad/parsers/parser_v2.py
```diff
@@ -0,0 +1,1166 @@
...
+                return  # Not our entry
+            section = self._create_entry_section()
+        else:
+            if not hdf_path.startswith(self._target_entry_name + "/"):
+                return  # Not under our entry
+            parent_section = self._sections.get(parent_path)
+            if parent_section is None:
+                return
+            nx_class_raw = hdf_node.attrs.get("NX_class", b"")
+            nx_class = (
```
mkuehbach: see this used at least one before, maybe L446-L454 utility function call

mkuehbach (changes requested):
Well documented, a reasonable rewrite, haven't run this yet though locally to test more

mkuehbach (approved):
As discussed at the DSUM 2026/08/18, technical implementation sound, more tests on production data possibly useful but should not block the FAIRmat team from moving forward with WPs in Phase 3, and beyond, can be tested as part of the more complete package in later stages given the report on sufficient initial feedback of many cases working already

tests/nomad/test_parsing_v2.py
lukaspie: Old comment, removed now

RubelMozumder:
SPM app verifies this PR via pynxtools-spm [PR-93](https://github.com/FAIRmat-NFDI/pynxtools-spm/pull/93)

src/pynxtools/nomad/parsers/parser_v2.py
lukaspie: extracted to function

tests/nomad/test_parsing_v2.py
lukaspie: done

lukaspie:
Merging as discussed

PR #812, opened 2026-07-15
Fix duplicated log output caused by root-logger propagation

### LINKED ISSUE #811: Duplicated log output caused by root-logger propagation
## Problem

Running any `pynx` CLI command (e.g. `pynx validate file.nxs`) prints every log message twice, with the second copy decorated with unrelated NOMAD context:
```
WARNING: The required attribute ... hasn't been supplied.
WARNING  pynxtools  2026-07-15T11:56:41Z WARNING: The required attribute ...

nomad.commit:
nomad.deployment: oasis
nomad.service: unknown nomad service
nomad.version: 1.4.4.dev26+g2dd2aba4d.d20260714
prefixed: True
taskName: None
```

## Root cause

The `"pynxtools"` logger (`pynxtools/__init__.py`) never disabled propagation, so every record it emits also bubbles up to the root logger. `pynx`'s top-level CLI (`cli.py`) unconditionally imports `pynxtools.nomad.cli`, which transitively imports `nomad.utils.structlogging`. That module calls `configure_logging()` inside `nomad-lab` during import, which installs a handler on the root logger and forces NOMAD's formatter onto it (hence the  `nomad.service`/`nomad.deployment`/`nomad.version`/`nomad.commit` context on every record).

As a result:, pynxtools's own handler prints the message once, and the NOMAD-formatted root handler prints it a second time with unrelated NOMAD metadata attached.

lukaspie:
Fixes #811 

Set `logger.propagate = False` on the `"pynxtools"` logger, so its records stay local to its own handler regardless of what else (re)configures the root logger.

This only addresses the pynxtools side. The underlying NOMAD-side issue (`nomad.utils.structlogging` mutating the global root logger as an import-time side effect (`configure_logging()` called at module scope)) is still present and can affect any other tool that imports `nomad-lab`. We could maybe discuss this with Area D at some point.

### DIFF
--- a/src/pynxtools/__init__.py
+++ b/src/pynxtools/__init__.py
@@ -40,6 +40,7 @@ formatter = CustomFormatter("%(message)s")
 handler.setFormatter(formatter)
 logger.addHandler(handler)
 logger.setLevel(logging.INFO)
+logger.propagate = False
 
 
 def _build_version(tag: str, distance: int, node: str, dirty: bool) -> str:


mkuehbach (dismissed):
Overwriting the default propagate parameter `True` makes sense to remain locally scoped, that's ok.

I am not able though to observe the issue that is reported in #811:
Tested with `pynxtools==0.15.0` and `nomad-lab==1.4.3` installed,
when running `pynx validate mynexusfile.nxs` I just get the pynxtools log messages just fine without nomad log messages.

The #811 mentions `pynxtools.nomad.cli` in `cli.py` but line is not part of the `v0.15.0` code base.

lukaspie:
> Overwriting the default propagate parameter `True` makes sense to remain locally scoped, that's ok.
> 
> I am not able though to observe the issue that is reported in #811: Tested with `pynxtools==0.15.0` and `nomad-lab==1.4.3` installed, when running `pynx validate mynexusfile.nxs` I just get the pynxtools log messages just fine without nomad log messages.
> 
> The #811 mentions `pynxtools.nomad.cli` in `cli.py` but line is not part of the `v0.15.0` code base.

Right, that's not on main, `pynxtools.nomad.cli` only exists in #801 and follow-ups. So you can see the issue when installing `pynxtools@nexus-metainfo-phase2` with `nomad-lab==1.4.3`.

In any way, what you said above keeping the logger locally scoped still holds.

RubelMozumder (approved):
LGTM!

PR #814, opened 2026-07-15
Fix false positive AXISNAME_indices required-attribute warning

### LINKED ISSUE #813: Validator raises false `required attribute AXISNAME_indices hasn't been supplied` warning
## Bug

Validating a valid NXem/NXdata file against its application definition produces a false-positive warning:

```
WARNING: The required attribute /entry/.../image_2d/@AXISNAME_indices hasn't been supplied.
```

even though the file correctly defines the concrete attribute (e.g. `@axis_i_indices`).

## Root cause

`AXISNAME_indices` is a group-level, variadic (`nameType="partial"`) attribute on `NXdata`. Three compounding bugs in `ValidationVisitor` (`src/pynxtools/dataconverter/validation.py`) prevent it from ever being matched off the required-entities set:

1. `_handle_nxdata`'s `AXISNAME_indices` check never calls `_remove_from_req_entities` for these attributes.
2. `_variadic_node_exists_for` calls `resolve_path(...)` with the `"@"` still attached to the last path segment (e.g. `".../@AXISNAME_indices"`), but `resolve_path`/`best_child_for` match against bare concept names (without `@`), so the schema lookup for the concept node silently fails.
3. Even when the concept node resolves, the instance-name comparison also keeps the `"@"` (`"@axis_i_indices"` vs `"AXISNAME_indices"`). Since this attribute's `nameType` is `"partial"` (not `"any"`), `get_nx_namefit` legitimately returns `-1` for the mismatch instead of masking it.

Interestingly, `tests/dataconverter/test_validation.py::test_validate_file_storage_layouts` already encoded this bug: one of its accepted outcomes in `warnings_storage_layouts_alternative` included this exact spurious warning line.

lukaspie:
Fixes #813

- [x] Fixes the three compounding bugs in `_variadic_node_exists_for` / `_handle_nxdata` (`validation.py`) reported in #813
- [x] Updates `test_validate_file_storage_layouts`'s `warnings_storage_layouts_alternative` expectations, which had encoded this exact spurious warning as one of two "acceptable" non-deterministic outcomes.
- [x] Adds a test on both validators: gave `NXtest`'s `NXODD_name` group a required `AXISNAME_indices` attribute and adds a ngeative case to `test_validate_data_dict` and `test_validate_nexus_file`. Also updated `NXtest_file.nxs` accordingly.
- [x] Verified against a real NXem file with `axis_i_indices`/`axis_j_indices` attributes — validator now reports the entry as valid


mkuehbach (approved):
Supported, thanks for catching that indeed false positive.

PR #817, opened 2026-07-22
bump definitions

lukaspie:
See https://github.com/FAIRmat-NFDI/nexus_definitions/pull/433

`get_local_name_from_xml` was removed from `dev_tools` a while ago, I replaced it here by a small inline helper. Anyway not super important as the `pynxtools` function that called `get_local_name_from_xml` is deprecated and not in use anymore.


RubelMozumder (approved):
LGTM!

PR #819, opened 2026-07-28
[fix] recursive parent application concept inheritance, default units

lukaspie:
Fixes two unrelated issues with the new schema:

- **inheritance for concepts defined in nested groups for application definitions extending other appdefs**
  Previously only top-level named groups inherited from the corresponding parent application concept. Nested named groups fell back to the generic base class because the updated parent application context was not passed to recursive calls, producing incorrect inheritance chains such as `MpesArpesInstrumentElectronanalyzerCollectioncolumn(MpesCollectioncolumn)` instead of
the accurate `MpesArpesInstrumentElectronanalyzerCollectioncolumn(MpesInstrumentElectronanalyzerCollectioncolumn)`. We fix this by extracting the parent application concept resolution into a helper and propagate the resolved parent application context through recursive `_build_named_concept` calls.

- **Default units did not match typical materials science units**
  Previously, we basically used SI units (e.g. `J` for `NX_ENERGY`) for the default units (and, crucially, `defaultDisplayUnit`). This led to data being harder to understand in the GUI. This fixes the default units for some of the unit categories. We probably will have to make manual changes later on (but only after the schema stabilizes), depending on the actual semantics and convention in a field. An example for this is ´NX_PRESSURE´ -> `mbar` or `pascal`.

## Follow-up: further cross-app inheritance gaps found by a corpus-wide audit

The fix above made `_parent_generates_concept_at`/`_declares_content_at` (the functions that decide whether a descendant application should inherit a generated concept from an intermediate ancestor application, rather than the fully generic base class) precise enough to catch real regressions — but precision surfaced several more cases where the check was either too permissive (false positives, referencing classes that are never actually emitted) or too strict (false negatives, silently skipping a real ancestor's content). None of these crash anything on their own — the chosen base is always some *existing* class — they just silently drop fields/subsections a descendant should have inherited. Found and fixed, in order:

- **Category-dependent optionality defaults.** NeXus defaults differ by category: base classes default to optional, application definitions default to required. An appdef that reuses a base class's group without any occurrence attributes at all always looked "required" by that silent default alone, with zero override intent — producing a phantom class reference (`SensorScanData`) that was never actually emitted.
- **`nxdl_base` attribution collapse.** `nxdl_base` tracks only whichever level most recently redeclared a child. A more-derived application redeclaring a child purely to nest one more thing underneath it (e.g. `NXsts` redeclaring `NXspm`'s own `bias_spectroscopy_environment` group just to add `z_controller`) hid the less-derived ancestor's real content from the check, since it only looked at children whose `nxdl_base` matched exactly.
- **Same-type, different-name content.** When a nested group's type already exists in the generic base class but under a different name (`NXxrd`'s own `raw_data` vs. generic `NXdetector`'s unnamed data child), the check found no name match and skipped straight to recursion instead of treating it as new content.
- **Requiredness on fields.** The field-comparison helper excluded requiredness entirely, but an ancestor can differ from generic *only* in requiredness (e.g. `NXxrot`'s own detector fields match generic dtype/unit exactly, differing only because an appdef's default is required vs. generic's optional) — real content that the standalone-emission logic already accounted for, but the cross-app check didn't.
- **`<link>` elements.** The check had no branch for links at all, so a group whose only own content is a link (`NXxbase`'s `data`, a bare link to the detector's data field) was entirely invisible to it.
- **`<choice>` elements**, for the same reason as links — closed proactively; no current application definition declares its own `<choice>`, so this isn't fixing a live bug today.

Verified via full corpus regeneration (all 85 applications, 281 files) after each fix, cross-checked against the raw NXDL for every changed class, `mypy` clean across the generated corpus, full test suite passing, and a dedicated audit confirming no generated class is selected as an inheritance base while contributing zero of its own quantities/subsections (the one exception, `Collection`, is empty by NeXus design).

One further case (`NXxrd`'s `DATA` vs. `NXmonopd`'s `<link>`s of the same name) is left unfixed — known, documented in the generator, not worth touching shared tree-merge code for something `NXxrd` is expected to outgrow once nomad-measurements integration lands.


RubelMozumder (changes requested):
I have found some inconsistencies in the inheritance chain of application definitions,
  1. `directtof.DirecttofInstrument` should be `tofraw.TofrawInstrument`
  2. `indirecttof.IndirecttofInstrument` should be `tofraw.TofrawInstrument`
  3. `iv_temp.IvTempSample` should be `sensor_scan.SensorScanSample`
  4. `iv_temp.IvTempInstrument` should be `sensor_scan.SensorScanInstrument`
  5. `iv_temp.IvTempInstrumentEnvironment` should be `sensor_scan.SensorScanInstrumentEnvironment` they needs to be

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
```diff
@@ -668,6 +668,39 @@ def _parent_generates_concept(child: NXTreeGroup, parent_app_file: str) -> bool:
...
+    inherit the parent app's concept
+    (``XpsInstrumentElectronanalyzer(MpesInstrumentElectronanalyzer)``), not the
+    generic base — otherwise parent-app additions (NXmpes's ``electron_detector``)
+    are dropped. Returns ``(base_class_override, parent_concept_file,
+    child_parent_app)``; the last is the ``parent_app`` for deeper recursion.
+    """
+    if parent_app is None:
+        return None, None, None
+    parent_app_file, parent_app_module, parent_naming_base = parent_app
+    if child.definition_file_at(1) != parent_app_file or not _parent_generates_concept(
```
RubelMozumder: The check assumes the parent application's declaration sits at position 1 of the child's
inheritance chain, which only holds when the derived app *also* declares the group. When
it doesn't, the parent's declaration is at index 0 and the test fails. Real chains:

```
NXsts   SAMPLE  → ['NXspm', 'NXsensor_scan', 'NXentry', 'NXsample', …]   parent at idx 0 → missed
NXsts   INSTRUMENT → ['NXsts', 'NXspm', 'NXsensor_scan', …]              parent at idx 1 → matched

**Review by RubelMozumder — COMMENTED** [review]
Continue!

File: src/pynxtools/nomad/converters/nxdl_to_metainfo.py (hunk 1 above)
Comment (lukaspie) [reply src/pynxtools/nomad/converters/nxdl_to_metainfo.py]: You're right, and it goes a bit deeper than the two examples above suggested.

The root cause turned out to be in the tree itself, not just the index math: `search_add_child_for` attributes a nested group's `nxdl_base` to whichever level last redeclares it. That's fine when a derived app only adds new siblings, but `NXiv_temp` **redeclares** `instrument` specifically to refine the `environment` child that already exists in `NXsensor_scan` (adding three named sensors to it). That refinement overwrites `environment`'s attribution to point at `NXiv_temp`, so `_parent_generates_concept` had no way left to see that `NXsensor_scan` also declares it.

`_parent_app_concept_override` now walks the whole ancestor-application chain instead of assuming the direct parent is always at a fixed spot, and for each ancestor it checks the raw XML at whatever position that ancestor actually sits at in `child.inheritance`, rather than trusting the merged tree's collapsed `nxdl_base`. If the nearest ancestor doesn't declare a group at all, or declares it but adds nothing new, it now falls through to the next one back, so a longer chain doesn't just give up and fall back to the generic base.

Regenerated everything to check for regressions: `SpmInstrument(SensorScanInstrument)` and `StsInstrument(SpmInstrument)`, which already worked, come out byte-identical. `iv_temp`'s three classes now inherit correctly, and it also caught `StsInstrumentBiasSpectroscopyEnvironment`, which had the same problem.

**Comment by lukaspie** [comment]
> I have found some inconsistencies in the inheritance chain of application definitions,
> 
> 1. `directtof.DirecttofInstrument` should be `tofraw.TofrawInstrument`
> 2. `indirecttof.IndirecttofInstrument` should be `tofraw.TofrawInstrument`
> 3. `iv_temp.IvTempSample` should be `sensor_scan.SensorScanSample`
> 4. `iv_temp.IvTempInstrument` should be `sensor_scan.SensorScanInstrument`
> 5. `iv_temp.IvTempInstrumentEnvironment` should be `sensor_scan.SensorScanInstrumentEnvironment` they needs to be

Indeed, the inheritance was not correctly resolved. See [this comment](https://github.com/FAIRmat-NFDI/pynxtools/pull/819#discussion_r3704695017), with the latest commit, three of these are fixed:

**Fixed** (`iv_temp`):
- `IvTempSample` → now inherits `SensorScanSample`
- `IvTempInstrument` → ment`now inherits `SensorScanInstrument`
- `IvTempInstrumentEnvironment` → now inherits `SensorScanInstrumentEnvironment`

However, the `directtof`/`indirecttof` inheritance is **deliberately** not changed. `NXdirecttof`/`NXindirecttof`'s `instrument` group has no `name` attribute (generic/variadic `NXinstrument`), while `NXtofraw`'s is `name="instrument"`. These aren't the same declared concept as far as NXDL is concerned: a generic "any instrument" group and "the one group specifically named `instrument`" are different constraints. This seems inconsistent in the NXDL itself rather than something for the generator to paper over. Could be a separate issue in the NeXus definitions if we want it resolved at the source.

**Comment by RubelMozumder** [comment]
> > I have found some inconsistencies in the inheritance chain of application definitions,
> > 
> > 1. `directtof.DirecttofInstrument` should be `tofraw.TofrawInstrument`
> > 2. `indirecttof.IndirecttofInstrument` should be `tofraw.TofrawInstrument`
> > 3. `iv_temp.IvTempSample` should be `sensor_scan.SensorScanSample`
> > 4. `iv_temp.IvTempInstrument` should be `sensor_scan.SensorScanInstrument`
> > 5. `iv_temp.IvTempInstrumentEnvironment` should be `sensor_scan.SensorScanInstrumentEnvironment` they needs to be
> 
> Indeed, the inheritance was not correctly resolved. See [this comment](https://github.com/FAIRmat-NFDI/pynxtools/pull/819#discussion_r3704695017), with the latest commit, three of these are fixed:
> 
> **Fixed** (`iv_temp`):
> 
> * `IvTempSample` → now inherits `SensorScanSample`
> * `IvTempInstrument` → ment`now inherits `SensorScanInstrument`
> * `IvTempInstrumentEnvironment` → now inherits `SensorScanInstrumentEnvironment`
> 
> However, the `directtof`/`indirecttof` inheritance is **deliberately** not changed. `NXdirecttof`/`NXindirecttof`'s `instrument` group has no `name` attribute (generic/variadic `NXinstrument`), while `NXtofraw`'s is `name="instrument"`. These aren't the same declared concept as far as NXDL is concerned: a generic "any instrument" group and "the one group specifically named `instrument`" are different constraints. This seems inconsistent in the NXDL itself rather than something for the generator to paper over. Could be a separate issue in the NeXus definitions if we want it resolved at the source.

Yeah, the chain rules are correct in `NXdirecttof`/`NXindirecttof`.

**Comment by RubelMozumder** [comment]
Still, there are some inconsistencies in the inheritance chain:
- `xeuler.XeulerInstrument(Instrument)` should be `XeulerInstrument(XbaseInstrument)` 
- `xeuler.XeulerInstrumentDetector(Detector)` should be `Xeuler.XeulerInstrumentDetector(XbaseInstrumentDetector)`

The same issue for `xlaue.XlaueInstrument(Instrument)` should be `xlaue.XlaueInstrument(XrotInstrument)`

**Review by RubelMozumder — APPROVED** [review]
All of the prvious issues are fixed. But, I see, now the inheritance chain breaks in different ways,

```
1. StsInstrumentBiasSpectroscopyEnvironment(SensorScanInstrumentEnvironment) --> StsInstrumentBiasSpectroscopyEnvironment(SpmInstrumentBiasSpectroscopyEnvironment)
2. XrdPanInstrumentDetector(Detector) --> XrdPanInstrumentDetector(XrdInstrumentDetector)
3. IvTempInstrumentEnvironment(Environment) --> IvTempInstrumentEnvironment(SensorScanInstrumentEnvironment)
4. IvTempInstrument(Instrument)--> IvTempInstrument(SensorScanInstrument) (I think this breaks because the changes are from `recommended` to `required` in the inheritance chain).
```

lukaspie:
> All of the prvious issues are fixed. But, I see, now the inheritance chain breaks in different ways,
> 
> ```
> 1. StsInstrumentBiasSpectroscopyEnvironment(SensorScanInstrumentEnvironment) --> StsInstrumentBiasSpectroscopyEnvironment(SpmInstrumentBiasSpectroscopyEnvironment)
> 2. XrdPanInstrumentDetector(Detector) --> XrdPanInstrumentDetector(XrdInstrumentDetector)
> 3. IvTempInstrumentEnvironment(Environment) --> IvTempInstrumentEnvironment(SensorScanInstrumentEnvironment)
> 4. IvTempInstrument(Instrument)--> IvTempInstrument(SensorScanInstrument) (I think this breaks because the changes are from ``recommended`` to ``required`` in the inheritance chain).
> ```

These are now also fixed, see details in the updated PR description. I think the inheritance is resolved much better now, but I am not 100% sure it is correct. I suggest we merge this now (it has substantially grown in scope already). The next step should be to have tests that cover the generator in a comprehensive manner, starting with #824.

PR #766, opened 2026-03-27
Deactivate the writing and cleaning of the NeXus schema in json file.

RubelMozumder:

PR #821, opened 2026-07-30
chore: remove code from deactivated schema caching

### LINKED ISSUE #765: Cleaning up code and house keeping associated with deactivating NeXus schema caching in NOMAD
An optimization attempt to load NeXus schemas faster in NOMAD was tested but turned out not substantially faster or having issues.

`v0.13.2` has still code snippets that refer to this dead code, as a hotfix in branch `DeactivateCaching` 
#766 adding to the v0.13.2 we deactivated its
use while prepping the deployment for NOMAD v1.4.2rc1, these changes should be cleaned up properly.

The issue and schema integration in NOMAD will anyway change in the refactoring and alignment sprints with Areas A, B, C, D soonish, then this issue here should be addressed as a side aspect.

lukaspie:
As we are now developing the new (much faster loading) schema, we can remove the old schema caching mechanism which was anyway not in use due to multiple issues, e.g., with normalizers not being properly attached.

Fixes #765

Reverts #613 and completes #760, #766, #767


RubelMozumder (approved):
LGTM!
Does `now developing the new schema` refers to schema generaton from with the jinja template (NeXus-NOMAD metainfor generation)?

lukaspie:
> LGTM! Does `now developing the new schema` refers to schema generaton from with the jinja template (NeXus-NOMAD metainfor generation)?

Yes, that will eventually replace the current schema in `schema.py`. That new schema loads much faster, so we don't have the need to cache the schema anymore.

PR #823, opened 2026-07-30
fix: pynx CLI requires nomad-lab even for non-NOMAD commands

### LINKED ISSUE #822: [Bug]: pynx CLI fails on all commands when nomad-lab isn't installed
### Contact Details

_No response_

### What happened?

Running any `pynx` subcommand (e.g. `pynx validate`, `pynx read`, `pynx convert`) fails with an error about a missing `nomad` package if the `nomad` extra (`nomad-lab`) is not installed, even though those commands have nothing to do with NOMAD.

## Root cause

`pynxtools/cli.py` unconditionally did:

    from pynxtools.nomad.cli import nomad

at module import time. This imports `pynxtools/nomad/__init__.py`, which unconditionally imports `nomad-lab` and raises `ImportError` if it's missing:
```
try:
    from nomad import config
    ...
except ImportError as exc:
    raise ImportError(
            "Could not import nomad package. Please install the package 'nomad-lab'."
    ) from exc
```

Since `nomad` is declared as an optional extra in `pyproject.toml`, this means the base `pynxtools` install is effectively broken for CLI use unless the extra is also installed.

## Expected behavior

Only `pynx nomad ...` sub-commands should require the `nomad` extra. All other `pynx` commands should work with a base `pynxtools` install, and `pynx nomad` should fail with a clear, actionable message when `nomad-lab` is missing.

### Relevant log output

```shell

```

lukaspie:
`pynx` eagerly imported `pynxtools.nomad.cli` at module load time, which unconditionally imports `nomad-lab` and raises `ImportError` if the optional `'nomad'` extra isn't installed. This broke every `pynx` command, not just `pynx nomad`.

Fix is to make the `nomad` sub-group lazily import `pynxtools.nomad.cli` only when `pynx nomad ...` is actually invoked, and rasise a clear `ClickException` pointing at `'pip install pynxtools[nomad]'` when it's missing.

Also adds a bunch of tests for the main CLI tool itselt, including a monkeypatch test to see what happens if `nomad-lab` is not installed.

Fixes #822

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          uv pip install ".[docs]"
+          uv pip install ".[docs,nomad]"
 
       - name: Build and Deploy
         run: |
--- a/src/pynxtools/cli.py
+++ b/src/pynxtools/cli.py
@@ -39,12 +39,39 @@ from pynxtools.annotator.cli import read
 from pynxtools.dataconverter.cli import convert, validate
 from pynxtools.eln_mapper.cli import generate_eln
 from pynxtools.nexus.cli import inspect_appdef
-from pynxtools.nomad.cli import nomad
+
+
+class _LazyNomadGroup(click.Group):
+    """``pynx nomad`` sub-group, loaded on first use.
+
+    ``pynxtools.nomad`` requires the ``nomad`` extra (``nomad-lab``). Importing
+    it eagerly here would make every ``pynx`` invocation, including commands
+    unrelated to NOMAD, fail for users who only installed the base package.
+    """
+
+    def __init__(self) -> None:
+        super().__init__(name="nomad", help="NOMAD integration tools.")
+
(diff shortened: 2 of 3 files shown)


RubelMozumder (approved):
I see a probable issue, 
 
- Issue in docs:
```
::: mkdocs-click
    :module: pynxtools.cli
    :command: nomad
```

That worked because `from pynxtools.nomad.cli import nomad` re-exported the symbol into `pynxtools.cli`. The docs environment needs the `nomad` extra as well — `uv pip install ".[docs,nomad]"` in build_docs.yml

lukaspie:
> I see a probable issue,
> 
> * Issue in docs:
> 
> ```
> ::: mkdocs-click
>     :module: pynxtools.cli
>     :command: nomad
> ```
> 
> That worked because `from pynxtools.nomad.cli import nomad` re-exported the symbol into `pynxtools.cli`. The docs environment needs the `nomad` extra as well — `uv pip install ".[docs,nomad]"` in build_docs.yml

Yes, makes sense, I added it there 👍

PR #824, opened 2026-08-05
Add unit tests for the NXDL → metainfo converter

RubelMozumder:
## Goal

Pin the transformation rules of `nxdl_to_metainfo`, the NXDL → NOMAD metainfo
generator, so a change to the converter or the Jinja2 template shows up as a
test failure rather than as a silently different generated schema.

## Flow

Two layers. Unit tests drive single converter functions one rule per test: the
definition root comes from a fixture NXDL, and the nodes under test are attached
to it directly rather than resolved from the NeXus tree, so each test states its
input in full. Golden tests then run the fixture NXDLs through the real path —
`generate_tree_from → build_context → render → write_class` — and compare
structurally (class bases and member names/kinds, via `ast`), so formatting and
docstring changes cannot cause a failure.

Fixture NXDLs are used instead of live definitions, so upstream NeXus changes
cannot break the goldens. Two unit tests deliberately do use the live `NXentry`
tree, to pin behavior against a real inheritance chain; the module docstring
says which and why.

## Where to start the review

1. The module docstring of `tests/nomad/test_nxdl_to_metainfo.py` — goal, flow,
   per-function coverage map, why private helpers are asserted on directly, and
   how to regenerate the golden files. It is the single source of truth; there is
   deliberately no second document to drift out of step with it.
2. `src/pynxtools/data/NXtestBase.nxdl.xml` and its one-line registration in
   `src/pynxtools/nexus/utils.py` — the only non-test change.
3. The rest of `tests/nomad/test_nxdl_to_metainfo.py` — each test documents what
   it pins.
4. `tests/data/nomad/converter/` — golden files, plus the `conftest.py` that
   keeps pytest from collecting them as test modules and records why their
   auto-generated "regenerate with" header does not apply to them.

## Known limitation

The converter memoizes NXDL lookups in module-level dicts
(`_NXDL_CATEGORY_CACHE`, `_base_group_nodes_cache`, `_base_class_qty_cache`).
Tests that stub `generate_tree_from` leave entries in those caches that were
computed under the stub, and `monkeypatch` restores the function but not the
dicts. Verified not to change any current assertion — `build_context("NXtest")`
is identical with and without the poisoned caches — but it does mean the suite
is not order-independent by construction. Left as is for this PR.


tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+
+Flow
+    Two layers, in file order.
+
+    1. Unit tests on single converter functions — ``_build_quantity_from_node``,
+       ``_build_subsection_from_node``, ``_base_from_extends``,
+       ``build_context``, ``write_base_class``, ``generate_all_base_classes``.
+       Input nodes are constructed directly rather than parsed.
+    2. Golden tests that render the fixture NXDLs end to end and compare the
+       result against stored output.
```
lukaspie: Tighten the language here, this is too detailed (especially naming the function names directly, these can always change and then this gets out of sync)

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+from pynxtools.nexus.nexus_tree import (
+    NexusAttribute,
+    NexusDefinition,
+    NexusField,
+    NexusGroup,
+    generate_tree_from,
+)
+from pynxtools.nomad.converters._mapping import _BASESECTION_RESERVED_NAMES
+
+_NX_NS = "http://definition.nexusformat.org/nxdl/3.1"
```
lukaspie: Import `namespaces = {"nx": "http://definition.nexusformat.org/nxdl/3.1"}` from `nexus_tree.py`, then reuse here

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+    assert section.variable is expected_variable
+    assert section.repeats is variadic
+
+
+def test_base_from_extends_for_direct_child_with_nomad_base():
+    """A direct NXobject child gets generated ``Object`` plus its NOMAD semantic bases.
+
+    Runs on the real ``NXentry`` tree, so the XML inheritance chain is exercised
+    unmocked and the expected NOMAD bases are read from ``BASESECTIONS_MAP``
+    rather than hard-coded.
```
lukaspie: This should be rephrased, very much LLM-speak

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+
+    ``build_context`` reaches outside the node tree in exactly two places a
+    synthetic definition cannot satisfy: ``_base_from_extends`` parses the real
+    NXDL named by ``extends``, and ``_all_ancestor_member_names`` walks the real
+    inheritance chain. Both are stubbed so the assertions isolate the
+    conflict-renaming rules; node traversal, quantity and subsection building
+    all still run for real.
+
+    ``ancestor_members`` is the ``(quantity_names, subsection_names)`` pair the
+    ancestor chain is taken to expose.
```
lukaspie: Rewrite, hard to understand

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+    )
+
+    context = converter.build_context("NXentry")
+    subsection_names = [s.python_name for s in context["subsections"]]
+
+    assert subsection_names == ["ancestor_quantity_name_group", "plain_group"]
+
+
+@pytest.mark.parametrize("reserved_name", ("name", "lab_id", "description"))
+def test_build_context_reserved_quantity_names_are_suffixed(monkeypatch, reserved_name):
```
lukaspie: This is not true anymore, we just use the same name as the quantities in the base section

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+    assert _class_bases(generated) == _class_bases(golden), (
+        "NXtestBase: class inheritance mismatch"
+    )
+    assert _class_members(generated) == _class_members(golden), (
+        "NXtestBase: class member mismatch (name or Quantity/SubSection kind changed)"
+    )
+
+
+# ---------------------------------------------------------------------------
+# NXtest / NXtest_extended: full-pipeline section-by-section golden tests
```
lukaspie:
```suggestion
# NXtest / NXtest_extended: full-pipeline section-by-section reference tests
```

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+            ("NXtest_extended", "test_extended.py"),
+        ):
+            src = c.render(c.build_context(nx))
+            # The fixtures resolve through _NXDL_SPECIAL_NAMES to absolute paths.
+            # Without this replace, the generating machine's layout ends up in
+            # the doc links stored in the golden.
+            pathlib.Path("tests/data/nomad/converter", fname).write_text(
+                src.replace(repo, ""), encoding="utf-8"
+            )
+        EOF
```
lukaspie: I suggest to make this a script in `scripts`?

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,586 @@
...
+
+    Review the resulting diff before committing: a structural change there is
+    exactly what the golden tests exist to make visible.
```
lukaspie: remove

tests/nomad/test_nxdl_to_metainfo.py
lukaspie: So this should actually fail now! Please check why it doesn't!

tests/data/nomad/converter/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+#
+
+"""Keep pytest from collecting the converter golden files as test modules.
+
+The goldens are named after the modules the converter generates (``test.py`` for
+NXtest, ``test_extended.py`` for NXtest_extended), and ``test_extended.py``
+matches pytest's default ``test_*.py`` discovery pattern. They are reference
+data, not tests: importing them would require the generated metainfo packages to
+be importable, which is not what this suite checks. ``collect_ignore_glob`` is
+pytest's supported hook for excluding files from collection in a directory.
```
lukaspie:
```suggestion
""Prevent pytest from collecting golden reference files as tests.

The converter's golden files include Python modules (e.g. ``test.py`` and
``test_extended.py``) whose names match pytest's default discovery patterns.
They are reference data for comparison, not executable tests.
"""
```
just simplifying

.cspell/custom-dictionary.txt
```diff
@@ -287,3 +287,5 @@ ylabel
 zaxis
 zenodo
 zstd
+nxtestbase
+testbase
```
lukaspie: This file is generally in alphabetical order. Revert, then use ´scripts/generate_custom_dict.sh` instead to regenerate.

src/pynxtools/data/NXtestBase.nxdl.xml
```diff
@@ -0,0 +1,44 @@
...
+    xmlns="http://definition.nexusformat.org/nxdl/3.1"
+ name="NXtestBase"
+ extends="NXobject"
+ type="group"
+ category="base"
+>
+    <doc>
+        Minimal base class used as a controlled fixture in converter unit tests.
+        Not part of the NeXus standard definitions. Covers: NX_CHAR, NX_FLOAT with
+        units, NX_INT, NX_BOOLEAN, closed enumeration, group reference, and
```
lukaspie: This is good, but not yet comprehensive. Things this should additionally cover:
- a non-named, variadic subgroup (`<group type="instrument"`)
- a named, partial, variadic subgroup (e.g. `<group type="NXsample" name="sampleID"  nameType="partial" >`)
- a field whose Metainfo name will clash with the name of an unnamed variadic group in `NXobject` (e.g. `<field name="collection" type="NX_CHAR"`) -> to check for name clashes and renaming
- a named, partial, variadic field (`<field name="labelID" nameType="partial" type="NX_CHAR">`
- dimensions -> to check conversion to shape
- open enums
- `units="NX_ANY"` -> `flexible_units` in the schema
- (possibly)  transformations  -> with respect to units
- (possibly) links

tests/data/nomad/converter/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+Note on the goldens' own headers: each carries the generator's standard banner,
+including ``Run `pynx nomad generate-metainfo --nxdl <NXclass>` to regenerate``.
+That command does not regenerate these files — it writes into
+``src/pynxtools/nomad/metainfo/``. The banner is template boilerplate that comes
+along with the rendered source. To refresh the goldens, use the snippet in the
+module docstring of ``tests/nomad/test_nxdl_to_metainfo.py``.
```
lukaspie: I think this information is not needed, espcially not here.

As stated below, the snippe should go to `scripts` instead

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Tests for the NXDL -> NOMAD metainfo converter.
+
+Goal
+    Pin the converter's transformation rules: how an NXDL node becomes a
+    ``Quantity`` or ``SubSection``, how names are resolved when they collide,
+    and how a whole NXDL renders to Python source.
```
lukaspie: Remove, this is clear

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+Flow
+    Two layers, in file order.
+
+    1. Unit tests on single converter functions — ``_build_quantity_from_node``,
+       ``_build_subsection_from_node``, ``_base_from_extends``,
+       ``build_context``, ``write_base_class``, ``generate_all_base_classes``.
+       The definition root comes from a fixture NXDL, and the nodes under test
+       are attached to it directly rather than resolved from the NeXus tree.
+    2. Golden tests that render the fixture NXDLs end to end and compare the
+       result against stored output.
```
lukaspie: Tighten the language here, this is too detailed (especially naming the function names directly, these can always change and then this gets out of sync)

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+    - ``_base_from_extends`` — primary Python base plus NOMAD semantic bases.
+    - ``build_context`` — the three name-conflict rules: field vs. SubSection,
+      group vs. ancestor Quantity, reserved ``BaseSection`` names.
+    - ``write_base_class`` — a dry run reports the difference and leaves the file
+      on disk unwritten; a full run puts the module in the folder the NXDL's
+      category selects, and the rendered source compiles.
+    - ``generate_all_base_classes`` — the returned count includes only changed
+      classes.
+    - ``render`` (via the golden files) — class bases and member names/kinds for
+      every fixture NXDL.
```
lukaspie: Again, this is way too detailed. It will just get out of touch when the function names change.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+      Two unit tests deliberately do not: ``_base_from_extends`` is asserted
+      against the real ``NXentry`` tree, and the ``write_base_class`` dry-run
+      resolves ``NXentry``'s category. Both would notice an upstream change to
+      ``NXentry``, which is the point — they pin behavior against a real
+      inheritance chain rather than a fixture.
+    - Golden files in ``tests/data/nomad/converter/``. Its ``conftest.py`` sets
+      ``collect_ignore_glob`` because ``test.py`` and ``test_extended.py`` match
+      pytest's discovery pattern but are reference data, not tests.
+    - No test depends on another. Each builds its own input and asserts on its
+      own output, so they can run in any order or alone.
```
lukaspie: Again, too detailed. Please rewrite this whole module dosctring to focus on the important aspects.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+            ("NXtest_extended", "test_extended.py"),
+        ):
+            src = c.render(c.build_context(nx))
+            # The fixtures resolve through _NXDL_SPECIAL_NAMES to absolute paths.
+            # Without this replace, the generating machine's layout ends up in
+            # the doc links stored in the golden.
+            pathlib.Path("tests/data/nomad/converter", fname).write_text(
+                src.replace(repo, ""), encoding="utf-8"
+            )
+        EOF
```
lukaspie: I suggest to make this a script in `scripts`?

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+
+    Review the resulting diff before committing: a structural change there is
+    exactly what the golden tests exist to make visible.
```
lukaspie: remove

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+    NexusAttribute,
+    NexusDefinition,
+    NexusField,
+    NexusGroup,
+    generate_tree_from,
+)
+from pynxtools.nexus.utils import get_nxdl_root_and_path
+
+
+def _fixture_definition(nx_name: str) -> NexusDefinition:
```
lukaspie: Make this an actual fixture (using a factory)

```python
@pytest.fixture
def definition_factory():
    def factory(nx_name: str) -> NexusDefinition:
        """..."""
        elem, path = get_nxdl_root_and_path(nx_name)
        return NexusDefinition(name=nx_name, nxdl_base=path, inheritance=[elem])

    return factory
```

Then use it like this
```
def test_something(definition_factory):
    root = definition_factory("NXtestBase")
```

This is more idiomatic for `pytest`.

lukaspie: Also the docstring is again way too long and over-detailed.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+    a root's children with those of its ``NXentry`` group, and the
+    ``generate_tree_from`` stub below would make ``_nxdl_category()`` report
+    ``category="application"`` for every class it is asked about, which routes
+    group lookups to the wrong output package and drops them.
+    """
+    elem, path = get_nxdl_root_and_path(nx_name)
+    return NexusDefinition(name=nx_name, nxdl_base=path, inheritance=[elem])
+
+
+def _class_members(source: str) -> dict[str, dict[str, str]]:
```
lukaspie: Check if this `_existing_member_names` from the generator can be reused here.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+                                elif isinstance(fn, ast.Attribute):
+                                    kind = fn.attr
+                                else:
+                                    continue
+                                members[tgt.id] = kind
+            result[node.name] = members
+    return result
+
+
+def _class_bases(source: str) -> dict[str, list[str]]:
```
lukaspie: Also here, can functionality from the converter be reused?

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+                    base.value, ast.Name
+                ):
+                    bases.append(f"{base.value.id}.{base.attr}")
+            result[node.name] = bases
+    return result
+
+
+# Golden files generated from the fixture NXDLs (NXtestBase, NXtest,
+# NXtest_extended) — independent of the live NeXus definitions repo.
+_CONVERTER_GOLDEN_DIR = Path(__file__).parent.parent / "data" / "nomad" / "converter"
```
lukaspie: While probably technically correct, "golden files" just sounds weird and LLM-generated. Suggestion to rename it across the file to "reference files" (and hence `REFERENCE_OUTPUT_DIR`, etc.). This is also more in line with the rest of the tests in the repo.

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+def test_build_quantity_from_field_maps_transformation_unit(monkeypatch):
+    """Field nodes carry unit, shape and enum semantics onto a QuantityContext.
+
+    ``NX_TRANSFORMATION`` collapses to ``NX_ANY`` and a closed enum becomes
+    ``MEnum``; both reach the template verbatim, so a wrong mapping corrupts
+    every generated Quantity silently.
+    """
+    # NXUnitSet.get_dimensionality requires the full NOMAD unit database at runtime.
+    # The lambda isolates this test from that external dependency.
+    monkeypatch.setattr(converter.NXUnitSet, "get_dimensionality", lambda unit: "[]")
```
lukaspie: Shouldn't this work without monkeypatching?

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+
+    ``build_context`` reaches outside the node tree in exactly two places that a
+    root carrying only the nodes under test cannot satisfy:
+    ``_base_from_extends`` parses the real NXDL named by ``extends``, and
+    ``_all_ancestor_member_names`` walks the real inheritance chain. Both are
+    stubbed so the assertions isolate the conflict-renaming rules; node
+    traversal, quantity and subsection building all still run for real.
+
+    ``ancestor_members`` is the ``(quantity_names, subsection_names)`` pair the
+    ancestor chain is taken to expose.
```
lukaspie: rephrase, this is very hard to understand.

Focus on what it tests, not how it does it

tests/nomad/test_nxdl_to_metainfo.py
```diff
@@ -0,0 +1,584 @@
...
+
+    monkeypatch.setattr(converter, "write_class", fake_write)
+
+    written = converter.generate_all_base_classes(output_dir=tmp_path)
+
+    assert written == 1
+
+
+# ---------------------------------------------------------------------------
+# Full-pipeline tests: fixture NXDL → Python source on disk
```
lukaspie: I would put these full end to end tests into a separate test file

lukaspie (changes requested):
Overall, this is good. I left some comments suggesting improvements. I think we would also benefit from not just using LLM-generated function docstrings, but reviewing these and making them less verbose / more focussed.

tests/nomad/test_nxdl_to_metainfo.py
lukaspie: Sorry, this must have moved from an earlier review time. But you get what I mean.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Moved to a separate file.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Still, it is true if the `name` quantity has `has_shape=True`. And in the test, there is always a shape. But I am considering the other cases as well.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: All docs are restructured and simplified.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Yeah, with the latest code, without monkeypatching, it's working.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: No such function exists.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: The `_existing_member_names` can be used, but `_class_members` is more robust because it also considers the kind of the class.

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: A script is added!

RubelMozumder (commented):
Docs are thoroughly reviewd and simplified.

src/pynxtools/data/NXtestBase.nxdl.xml
RubelMozumder: Included!

.cspell/custom-dictionary.txt
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: Addressed!

tests/nomad/test_nxdl_to_metainfo.py
RubelMozumder: It is not used anymore.

src/pynxtools/data/NXtestBase.nxdl.xml
```diff
@@ -0,0 +1,85 @@
...
+    <field name="sampled_values" type="NX_FLOAT" units="NX_UNITLESS">
+        <doc>A one-dimensional field (tests dimensions → shape conversion).</doc>
+        <dimensions rank="1"><dim index="1" value="n"/></dimensions>
+    </field>
+    <field name="scale_factor" type="NX_FLOAT" units="NX_ANY">
+        <doc>A value in any unit (tests NX_ANY → flexible_unit in the schema).</doc>
+    </field>
+    <field name="offset" type="NX_FLOAT" units="NX_TRANSFORMATION">
+        <doc>A transformation value (tests NX_TRANSFORMATION → NX_ANY unit mapping).</doc>
+    </field>
```
lukaspie, line 60: This is only valid inside `NXtransformations` class normally, but it is fine for the test

tests/nomad/test_nxdl_to_metainfo.py
lukaspie: ok

lukaspie (approved):
Looks good now. I am approving, pending that the tests run through.

Issue #825, opened 2026-08-05
validate_dict_against() silently corrupts link/VDS template entries before Writer runs


lukaspie:
## Summary

`validate_dict_against()` silently overwrites `{"link": ...}` template entries with their fully unpacked value, permanently destroying the link (and any VDS `"shape"` slice) before the `Writer` ever runs. This turns a correctly sliced virtual dataset (or any plain internal/external link) into a full, unsliced, non-virtual copy in the output HDF5 file.

I discovered this while I was investigating why w`pynxtools-ellips`'s ellipsometry example produces an `NXellipsometry` file whose default-plot signal (`Psi_50deg`, requested as a 1D slice of a 3D `measured_data` array via `{"link": "/entry/data_collection/measured_data", "shape": np.index_exp[0, 0, :]}`) ended up as a full, non-virtual `(N_angles, N_observables, N_spectrum)` dataset instead: This causes anNXdata axis-length mismatch and breaks h5web's default plot rendering (`h5grove`/`jupyterlab_h5web` 500 error), see https://github.com/FAIRmat-NFDI/pynxtools-ellips/issues/40.

Can be reprouced by running

```bash
pynx convert run examples/eln_data.yaml examples/test-data.dat --reader ellips --nxdl NXellipsometry --output out.nxs
```

```python
>>> f['/entry/data_collection/Psi_50deg'].is_virtual
False
>>> f['/entry/data_collection/Psi_50deg'].shape
(3, 2, 1088)   # expected (1088,) — a 1D slice matching /entry/data_collection/wavelength
```

The root cause is a write-back in `handle_field` in `validation.py` that overwrites the original `{"link": ...}` template entry with its resolved value even though resolution is only meant for local type-checking. Two related issues: a false-positive on list-form multi-source VDS links, and an `if`/`if`/`else` (should be `elif`) that double-processes NXdata groups. Fix implemented on a branch; PR #826

PR #827, opened 2026-08-06
tests: cross-application inheritance + default-unit regression tests

lukaspie:
Adds tests for the generated metainfo schema inside `src/pynxtools/nomad/metainfo`.

- Content checks for the generated NOMAD metainfo classes (here we check directly against expected content from NXDL)
- Cross-application named-concept inheritance: for groups appearing in multiple application definitions in an `extends` chain, verifies the generated concept inherits from the nearest ancestor with real content, not a more generic one. Covers content checks on representative generated classes. See #819 for the corresponding implementation.

Additionally, adds a test for another change introduced in #819: the materials-science-friendly unit defaults (e.g. `mbar` for `NX_PRESSURE`) in `NXUnitSet.get_default_unit`.

Complementary to #824 (converter mechanics via synthetic fixtures), no overlap.


tests/nomad/test_generated_metainfo_content.py
```diff
@@ -0,0 +1,383 @@
...
+        for child in elem
+        if child.tag.split("}")[-1] == tag and "name" in child.attrib
+    }
+
+
+def _own_child_group_names(elem) -> set[str]:
+    """Return the subsection names of an XML element's direct ``<group>`` children.
+
+    Named groups use their explicit name; anonymous groups use the lowercased
+    NXDL type name with any ``NX`` prefix removed.
```
RubelMozumder: Jus wondering how it handles the anonymous group `NXanyclass` and named group `anyclass`. Finally, both will evolve exactly the same entity.

tests/nomad/test_generated_metainfo_content.py
```diff
@@ -0,0 +1,383 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("field_name", ["beam_center_x", "beam_center_y"])
+def test_base_class_quantity_unit_matches_nxdl(field_name):
+    """Verify that Detector field units and shapes match the NXDL definition."""
+    node = _top_level_field("NXdetector", field_name)
+    quantity = Detector.m_def.all_quantities[field_name]
+    assert quantity.unit == _expected_unit(node)
+    assert quantity.shape == _expected_shape(node)
```
RubelMozumder: We can make this test a bit generic,

```python
def _unit_fields(nx_class):
    return [f.name for f in generate_tree_from(nx_class).children
            if isinstance(f, NexusField) and f.unit]

@pytest.mark.parametrize("nx_class, cls", [("NXdetector", Detector), ("NXbeam_stop", BeamStop), ...])
def test_base_class_quantity_unit_matches_nxdl(nx_class, cls):
    for name in _unit_fields(nx_class):
        assert cls.m_def.all_quantities[name].unit == _expected_unit(...)
```
So, we can add a few more classes.

tests/nomad/test_generated_metainfo_content.py
```diff
@@ -0,0 +1,383 @@
...
+    convention: no <dimensions> at all -> [], each dimension -> "*" (this
+    corpus only uses symbolic/unbounded dims, never fixed integers).
+    """
+    if node.shape is None:
+        return []
+    assert all(d is None for d in node.shape), (
+        "helper assumes symbolic/unbounded dims; extend if a fixed-size "
+        "dimension shows up in a case this test covers"
+    )
+    return ["*"] * len(node.shape)
```
RubelMozumder: The test is intended to fail for the fixed-size dimensions now. Or does the converter make the fixed-size dimension open?

tests/nomad/test_generated_metainfo_content.py
```diff
@@ -0,0 +1,383 @@
...
+    assert issubclass(Xbase, basesections.Measurement)
+
+    entry = generate_tree_from("NXxbase")
+    entry_group = next(
+        c
+        for c in entry.children
+        if isinstance(c, NexusGroup) and c.nx_class == "NXentry"
+    )
+    expected_group_names = {
+        c.name.lower() if c.name_type == "any" else c.name
```
RubelMozumder: The `_own_child_group_names` function can be utilised here.

RubelMozumder (approved):
I have a few comments, other than that, LGTM!

tests/nomad/test_generated_metainfo_content.py
lukaspie: Yes, they get the same name, but that is intentional.

tests/nomad/test_generated_metainfo_content.py
lukaspie: I have gone with your suggestion, but also added a shape check

tests/nomad/test_generated_metainfo_content.py
lukaspie: done

tests/nomad/test_generated_metainfo_content.py
lukaspie: I have refactored this to 
```
return [d if d is not None else "*" for d in node.shape]
```
so that explicitly takes fixed shapes into account

tests/nomad/test_generated_metainfo_content.py
lukaspie: I adjusted the docstring slightly to make this clearer.

lukaspie:
`pynxtools-ellips` tests are failing because of #825, ignoring here

PR #828, opened 2026-08-07
fix additive-write check for generator

lukaspie:
If `force=False` in the generator, any changes to the `normalize()` (or any other) method should not be overwritten. However, the existing additive-only guard only checked for differences for _names_ of the class members, not in their body. This PR fixes tht by comparing each shared member's source text, not just its name. member present in both files but with different content now counts as user-modified, exactly like a member absent from the template counts as user-added, and blocks the overwrite the same way.

Also adds more comprehensive tests for this overwrite functionality.


RubelMozumder (approved):
LGTM!

Issue #830, opened 2026-08-10
JOSS review Software paper - Software Design


Grant-Xue:
The Software Design section provides a clear description of the major architectural decisions behind `pynxtools`, including the machine-actionable NeXus schema approach, plugin-based architecture, the Template object, and the validation workflow.

Minor comment:
The authors could strengthen this section by briefly discussing the trade-offs associated with their architectural choices. For example, the plugin-based architecture improves extensibility and allows independent development of new instrument-specific converters, but it may introduce additional complexity in plugin management, dependency handling, and maintaining consistency across different extensions. Similarly, treating NeXus definitions as machine-actionable schemas improves validation and interoperability, but may introduce additional computational overhead and implementation complexity compared with simpler conversion approaches.

A brief discussion of how these trade-offs were considered and balanced would better demonstrate the design rationale and align with the JOSS Software Design criteria.

lukaspie:
@sherjeelshabih FYI.

See https://github.com/openjournals/joss-reviews/issues/10691#issuecomment-5236244430

sherjeelshabih:
Thank you for the feedback. We have addressed this comment by adding a paragraph to the Software Design section discussing the trade-offs associated with the plugin-based architecture and schema-driven validation. In particular, we now describe the additional complexity and processing overhead introduced by these design choices, and explain why we consider these costs justified by the benefits in terms of extensibility, validation, interoperability, and maintainability as the number of supported techniques and file formats grows.

The corresponding changes have been included in the merged [PR 831](#831).

Issue #829, opened 2026-08-10
JOSS Review: Software Paper - Summary Section


Grant-Xue:
The Summary section provides a clear description of the purpose and high-level functionality of `pynxtools`.

The authors clearly introduce the motivation of the software by describing the challenges associated with heterogeneous experimental data formats and the need for FAIR data practices. The functionality of `pynxtools`, including data conversion to the NeXus format, modular plugin architecture, and automated validation, is appropriately summarized. The authors also provide evidence of practical impact through integration with the NOMAD platform and usage in experimental workflows.

Minor correction:
Some domain-specific terminology, such as "NeXus application definitions", may not be immediately clear to readers outside the NeXus community. A brief clarification of these terms would improve accessibility for a broader audience.

lukaspie:
@sherjeelshabih FYI.

See https://github.com/openjournals/joss-reviews/issues/10691#issuecomment-5236244430

sherjeelshabih:
Thank you for the feedback again. We added a brief clarification of the term “NeXus application definitions” in the Summary section. We now describe them as standardized schemas that specify how data and metadata for a particular type of experiment are organized within NeXus, making the terminology more accessible to readers outside the NeXus community.

The corresponding change has been included in the merged [PR 831](#831).

PR #831, opened 2026-08-10
Adds detail about costs and the decisions in design with these costs …

sherjeelshabih:
Addresses #829 and #830.


lukaspie (approved):
Looks good, thanks

PR #833, opened 2026-08-19
Add HDF5Reference and field statistics for shape-ful NeXus quantities

lukaspie:
This is the implementation for Phase 4 of the NeXus schema 

## Summary

- Use `HDF5Reference` typing for shape-ful (array-like) NeXus quantities generated from NXDL: instead of parsing the full array, these quantities store a reference into the originating HDF5/NeXus file, resolved and validated in `parser_v2.py`. Both the new and old GUI can resolve these now
- REe-adds per-field `FIELD_STATISTICS` (`__min`/`__max`/`__size`/`__ndim`) computed at parse time, but **only** for `Data`-derived fields (so fields in `NXdata` and inheriting classes), since the raw array itself is no longer stored directly. We expcility drop field statistics for non-NXdata-fields in order to keep the schema size manageable.
- Suppresses ELN annotations (`a_eln`, `a_display`) on `HDF5Reference`-typed quantities: their value is a reference string, not something a user edits, so no ELN component ever applies.
- Regenerates the `tests/data/nomad/converter/` reference fixtures to match current generator output (they predated both the HDF5Reference and field-statistics features).

## Performed tests

- [x] `pytest tests/nomad/`, all passing
- [x] Verified end-to-end against a real parsed XPS archive in both the current GUI and the new GUI (`nomad-gui`, branch [`417-add-support-for-hdf5references`](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-gui/-/merge_requests/580)): `HDF5Reference` quantities render correctly.

### DIFF
--- a/docs/learn/pynxtools/nexus-metainfo-generation.md
+++ b/docs/learn/pynxtools/nexus-metainfo-generation.md
@@ -547,14 +547,10 @@ exactly the `NX_ANY` case.
 
 ---
 
-## FIELD_STATISTICS
+## Field statistics
 
-Numeric array quantities inside an `NXdata`-derived class (`node.nx_class ==
-"NXdata"` at generation time — the generated class's `m_def` carries
-`a_nexus_group(nx_class="NXdata", ...)` or, for the top-level class, its own
-`a_nexus_definition(nx_class="NXdata")`) get four parallel scalar quantities
-generated automatically, so the array's summary statistics are searchable
-even though the array itself (JSON-stored in the archive) is not:
+Numeric array quantities inside an `NXdata`-derived class get four parallel
+scalar quantities generated automatically, so the array's summary statistics are searchable even though the array itself is not:
 
 ```python
 # NXDL: <field name="energy" type="NX_FLOAT" units="NX_ENERGY"> inside an NXdata group
@@ -576,45 +572,9 @@ energy__ndim = Quantity(
     description="Number of dimensions of energy in the HDF5 file.",
 )
 ```
-
-`__ndim` uses `np.int8`, not `np.uint8` — NOMAD has no real unsigned-integer
-type and silently downcasts `np.uint8` to signed `m_int8` in its type
-registry, so the generator uses the signed type directly rather than relying
-on that implicit downcast. No `{name}__mean` quantity is generated: v1's own
-mean-into-bare-field write already fails silently for the same class of
-field, so there is no working precedent to preserve — the main quantity's own
-value is still populated with the mean when it has no declared `shape`
-(matching existing behavior for signal/axis fields), and is only skipped when
-it has a real declared shape that can't hold a reduced scalar (e.g. a 3D
-`NXdata` signal).
-
-**Scope — any field inside an `NXdata`-derived class, not just variadic
-signal/axis fields.** Every field inside an `NXdata` group is either the
-signal or an axis, both array-valued by NeXus convention, whether the NXDL
-gives it a variadic name (`DATA`/`AXISNAME`, `name_type="any"`/`"partial"`)
-or a concrete one (e.g. NXmpes's `energy`/`kx`/`ky` inside its `DATA` group,
-`name_type="specified"`). An earlier, narrower version of this check only
-counted the variadic case as array-valued, missing every concretely-named
-axis field — this is why, for example, `energy__min`/`energy__max` only
-appeared on the `AXISNAME`-pattern fields at first and not on NXmpes's own
-`energy`/`kx`/`ky`/`delay`, until the check was widened to cover the whole
-`NXdata` group uniformly.
-
-This "is the field effectively an array" decision is intentionally kept
-separate from the quantity's own declared `shape`: `NXdata` signals/axes
-routinely have no `<dimensions>` element in NXDL at all (`node.shape is
-None`), so `has_statistics` can't be driven by `shape` alone — but declaring
-`shape=["*"]` on the quantity itself to work around that would change how the
-parser populates it, and previously broke real-file scalar-style population
-for fields like `AXISNAME`. So the generator computes "is this array-like for
-statistics purposes" independently of what `shape` gets declared on the
-Quantity.
-
-The parser (`parser_v2.py`) populates the four fields via
-`get_field_stats_iuf_chunked()`/`get_field_stats_iuf_contiguous()`
-(`nomad/parsers/_field_io.py`), wrapping each stat with `MQuantity.wrap()`
-when the underlying field is variadic so distinctly-named instances (e.g.
-`intensity__min` vs. `errors__min`) don't collide.
+Note that this applies only to fields in an ``NXdata`` group. Statistics are
+generated exclusively for data arrays, not for arbitrary quantities elsewhere
+in the schema.  Restricting statistics to data arrays keeps the generated schema manageable while covering the primary data intended for searching.
 
 ---
 
--- a/docs/learn/pynxtools/nexus-metainfo-generation.md
+++ b/docs/learn/pynxtools/nexus-metainfo-generation.md
@@ -547,6 +547,77 @@ exactly the `NX_ANY` case.
 
 ---
 
+## FIELD_STATISTICS
+
+Numeric array quantities inside an `NXdata`-derived class (`node.nx_class ==
+"NXdata"` at generation time — the generated class's `m_def` carries
+`a_nexus_group(nx_class="NXdata", ...)` or, for the top-level class, its own
+`a_nexus_definition(nx_class="NXdata")`) get four parallel scalar quantities
+generated automatically, so the array's summary statistics are searchable
+even though the array itself (JSON-stored in the archive) is not:
+
+```python
+# NXDL: <field name="energy" type="NX_FLOAT" units="NX_ENERGY"> inside an NXdata group
+energy = Quantity(type=np.float64, ...)
+energy__min = Quantity(
+    type=np.float64,
+    description="Minimum of energy, computed over the full array at parse time.",
+)
+energy__max = Quantity(
+    type=np.float64,
+    description="Maximum of energy, computed over the full array at parse time.",
+)
+energy__size = Quantity(
+    type=np.int64,
+    description="Number of elements of energy in the HDF5 file.",
+)
+energy__ndim = Quantity(
+    type=np.int8,
+    description="Number of dimensions of energy in the HDF5 file.",
+)
+```
+
+`__ndim` uses `np.int8`, not `np.uint8` — NOMAD has no real unsigned-integer
+type and silently downcasts `np.uint8` to signed `m_int8` in its type
+registry, so the generator uses the signed type directly rather than relying
+on that implicit downcast. No `{name}__mean` quantity is generated: v1's own
+mean-into-bare-field write already fails silently for the same class of
+field, so there is no working precedent to preserve — the main quantity's own
+value is still populated with the mean when it has no declared `shape`
+(matching existing behavior for signal/axis fields), and is only skipped when
+it has a real declared shape that can't hold a reduced scalar (e.g. a 3D
+`NXdata` signal).
+
+**Scope — any field inside an `NXdata`-derived class, not just variadic
+signal/axis fields.** Every field inside an `NXdata` group is either the
+signal or an axis, both array-valued by NeXus convention, whether the NXDL
+gives it a variadic name (`DATA`/`AXISNAME`, `name_type="any"`/`"partial"`)
+or a concrete one (e.g. NXmpes's `energy`/`kx`/`ky` inside its `DATA` group,
+`name_type="specified"`). An earlier, narrower version of this check only
+counted the variadic case as array-valued, missing every concretely-named
+axis field — this is why, for example, `energy__min`/`energy__max` only
+appeared on the `AXISNAME`-pattern fields at first and not on NXmpes's own
+`energy`/`kx`/`ky`/`delay`, until the check was widened to cover the whole
+`NXdata` group uniformly.
+
+This "is the field effectively an array" decision is intentionally kept
+separate from the quantity's own declared `shape`: `NXdata` signals/axes
+routinely have no `<dimensions>` element in NXDL at all (`node.shape is
+None`), so `has_statistics` can't be driven by `shape` alone — but declaring
+`shape=["*"]` on the quantity itself to work around that would change how the
+parser populates it, and previously broke real-file scalar-style population
+for fields like `AXISNAME`. So the generator computes "is this array-like for
+statistics purposes" independently of what `shape` gets declared on the
+Quantity.
+
+The parser (`parser_v2.py`) populates the four fields via
+`get_field_stats_iuf_chunked()`/`get_field_stats_iuf_contiguous()`
+(`nomad/parsers/_field_io.py`), wrapping each stat with `MQuantity.wrap()`
+when the underlying field is variadic so distinctly-named instances (e.g.
+`intensity__min` vs. `errors__min`) don't collide.
+
+---
+
 ## Additive-only generation
 
 The generator never removes hand-written additions. When a file already exists:
--- a/docs/learn/pynxtools/nexus-metainfo-generation.md
+++ b/docs/learn/pynxtools/nexus-metainfo-generation.md
@@ -489,20 +489,36 @@ treated as unbounded.
 
 ## Shape representation
 
-NXDL `<dimensions>` elements map to `shape` on the generated `Quantity`. Concrete
-integer sizes are preserved as-is. Unbounded or symbolically-named dimensions (e.g.
-`nP`, `nz`) become the wildcard `"*"`. NOMAD does not interpret NeXus symbol names.
-Symbol definitions from the NXDL `<symbols>` block are preserved in
-`NeXusDefinition.symbols` on the class `m_def`.
+NXDL `<dimensions>` elements map to `shape` on the generated `Quantity` —
+for **attributes** only. An attribute is not an addressable HDF5 object, so
+it keeps its array value directly:
 
 ```python
-# NXDL: <dimensions rank="2"><dim index="1" value="nP"/><dim index="2" value="3"/></dimensions>
-data = Quantity(
+# NXDL <attribute> with <dimensions rank="1"><dim index="1" value="n"/></dimensions>
+some_attr = Quantity(
     type=np.float64,
-    shape=["*", 3],
+    shape=["*"],
     ...
 )
 ```
+
+**Fields** become `HDF5Reference`-typed instead, whenever they are array-valued —
+either because the NXDL declares `<dimensions>`, or, inside an `NXdata`-derived class, because rank is only known at parse time (e.g. `NXdata`'s own `DATA`/`AXISNAME` fields, which declare no `<dimensions>` at all). The array is never copied into the archive. The `Quantity` instead stores a reference string to the dataset in the source HDF5/NeXus file (`<file>#/<hdf5_path>`), resolved at parse time in `pynxtools/nomad/parsers/parser_v2.py`. `shape`, `unit`, `dimensionality`, and `flexible_unit` don't apply to a reference and are omitted. No ELN component is attached — there is nothing to edit.
+
+```python
+# NXDL: <field name="data" type="NX_FLOAT" units="NX_ANY"> inside an NXdata group
+data = Quantity(
+    type=HDF5Reference,
+    links=[...],
+    ...
+)
+```
+
+Concrete integer dimension sizes are preserved as-is on attributes; unbounded
+or symbolically-named dimensions (e.g. `nP`, `nz`) become the wildcard `"*"`.
+NOMAD does not interpret NeXus symbol names. Symbol definitions from the NXDL
+`<symbols>` block are preserved in `NeXusDefinition.symbols` on the class
+`m_def`.
 
 ---
 
--- a/src/pynxtools/nomad/converters/nxdl_to_metainfo.py
+++ b/src/pynxtools/nomad/converters/nxdl_to_metainfo.py
@@ -568,20 +568,22 @@ def _build_quantity_from_node(
         python_type, shape, node.name_type, scalar_items, field_name=node.name or ""
     )
 
-    # NXDL leaves some NXdata signals (DATA/AXISNAME) with no <dimensions>
-    # element at all — node.shape is None, even though they are, by NeXus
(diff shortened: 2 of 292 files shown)


docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -489,21 +489,37 @@ treated as unbounded.
...
 
 ```python
-# NXDL: <dimensions rank="2"><dim index="1" value="nP"/><dim index="2" value="3"/></dimensions>
-data = Quantity(
+# NXDL <attribute> with <dimensions rank="1"><dim index="1" value="n"/></dimensions>
+some_attr = Quantity(
     type=np.float64,
-    shape=["*", 3],
+    shape=["*"],
     ...
```
RubelMozumder, line 501: The previous example was more reasonable as it showed two dimensions. I would prefer to have both multidimensional and single-dimensional.

docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -489,21 +489,37 @@ treated as unbounded.
...
 )
 ```
 
+**Fields** become `HDF5Reference`-typed instead, whenever they are array-valued —
+either because the NXDL declares `<dimensions>`, or, inside an `NXdata`-derived class, because rank is only known at parse time (e.g. `NXdata`'s own `DATA`/`AXISNAME` fields, which declare no `<dimensions>` at all). The array is never copied into the archive. The `Quantity` instead stores a reference string to the dataset in the source HDF5/NeXus file (`<file>#/<hdf5_path>`), resolved at parse time in `pynxtools/nomad/parsers/parser_v2.py`. `shape`, `unit`, `dimensionality`, and `flexible_unit` don't apply to a reference and are omitted. No ELN component is attached — there is nothing to edit.
```
RubelMozumder, line 506: What is the benifit to use python file as a reference to docs? It fits better in the reference section, and use the URL to that section in the description. It does not bring any extra benefit.

docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -489,21 +489,37 @@ treated as unbounded.
...
+    links=[...],
+    ...
+)
+```
+
+Concrete integer dimension sizes are preserved as-is on attributes; unbounded
+or symbolically-named dimensions (e.g. `nP`, `nz`) become the wildcard `"*"`.
+NOMAD does not interpret NeXus symbol names. Symbol definitions from the NXDL
+`<symbols>` block are preserved in `NeXusDefinition.symbols` on the class
+`m_def`.
```
RubelMozumder, line 521: If the symbol or specific dimension tag is not handled by Nomad, then how could a Nomad plot be generated? Though, there is a question of whether the NOMAD plot would be supported if anyone uses nexus file or HDF file as a data source.

docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -547,6 +563,37 @@ exactly the `NX_ANY` case.
...
+    description="Number of elements of energy in the HDF5 file.",
+)
+energy__ndim = Quantity(
+    type=np.int8,
+    description="Number of dimensions of energy in the HDF5 file.",
+)
+```
+Note that this applies only to fields in an ``NXdata`` group. Statistics are
+generated exclusively for data arrays, not for arbitrary quantities elsewhere
+in the schema.  Restricting statistics to data arrays keeps the generated schema manageable while covering the primary data intended for searching.
```
RubelMozumder, line 593: HDF5Reference is used for NXdata and any field where data is described by `dimension`. But field statistics are only available for NXdata/<any-field>, why is there this duality?

src/pynxtools/nomad/metainfo/applications/apm.py
```diff
@@ -4084,7 +4058,7 @@ class ApmMeasurementStandingVoltageTime(Data):
         ),
     )
     title = Quantity(
-        type=str,
+        type=HDF5Reference,
         links=[
             "https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXapm.html#nxapm-entry-measurement-standing-voltage-time-title-field"
         ],
```
RubelMozumder, line 4064: This is a title `str` typed; why is it an HDF5reference?

src/pynxtools/nomad/metainfo/applications/apm.py
```diff
@@ -4264,7 +4261,7 @@ class ApmMeasurementPulseFrequencyTime(Data):
         ),
     )
     title = Quantity(
-        type=str,
+        type=HDF5Reference,
         links=[
             "https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXapm.html#nxapm-entry-measurement-pulse-frequency-time-title-field"
         ],
```
RubelMozumder, line 4267: HDF5Reference?

src/pynxtools/nomad/metainfo/applications/apm.py
```diff
@@ -4866,7 +4942,7 @@ class ApmMeasurementPressureTime(Data):
         ),
     )
     pressure = Quantity(
-        type=np.float64,
+        type=HDF5Reference,
         links=[
             "https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXapm.html#nxapm-entry-measurement-pressure-time-pressure-field"
         ],
```
RubelMozumder, line 4948: Here is the same reference issue.

src/pynxtools/nomad/metainfo/applications/apm.py
```diff
@@ -4829,9 +4892,22 @@ class ApmMeasurementPressureTime(Data):
...
+        type=np.float64,
+        description="Maximum of elapsed_time, computed over the full array at parse time.",
+    )
+    elapsed_time__size = Quantity(
+        type=np.int64,
+        description="Number of elements of elapsed_time in the HDF5 file.",
+    )
+    elapsed_time__ndim = Quantity(
+        type=np.int8,
+        description="Number of dimensions of elapsed_time in the HDF5 file.",
```
RubelMozumder, line 4910: The same Reference issue is here.

RubelMozumder (changes requested):
These are a few comments. There are some other comments comming.

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: Because having field statistics for all shapeful quantities would blow up the schema to much. So this is a compromise between usability and performance.

The plan is anyway to make the statistics something that _any_ `HDF5Reference` can provide (so nomad core side). Then all of these statistics can go away again. But this is in the future.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: This one should be correct, as it is a float.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: Good question, need to check

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: What do you mean by NOMAD plot? The only plot we are doing at the moment is anyway through the NeXus HDF5web view.

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: You mean "remove the mention of  `pynxtools/nomad/parsers/parser_v2.py`"? That can of course be done.

In general, the `nexus-metainfo-generation` docs would benefit from a rewrite, currently it is a WIP version where a bunch of information is dropped, but this should be more user-facing, I agree.

docs/learn/pynxtools/nexus-metainfo-generation.md
RubelMozumder: HDF5Reference is for array data only, and the statistics are only for thouse HDF5Reference.  I do not think the extra 4 float numbers for array data would cause any trouble regarding the cache and space. 

If there are no statistics, that is completely ok; otherwise, this duality would be seen as a bug by users who never know our implicit rule.

src/pynxtools/nomad/metainfo/applications/apm.py
RubelMozumder: It is not an array, and HDF5Reference only for array data with dimensions, isn't it?

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
```diff
@@ -558,8 +575,31 @@ def _build_quantity_from_node(
     else:
         python_type = nx_type_to_source(node.dtype)
 
+    # Only NXdata fields receive statistics. Since signals and axes are always
+    # array-valued by convention, treat missing <dimensions> as "array of unknown
+    # rank" for this decision only; the Quantity.shape remains the explicit NXDL shape.
+    is_effectively_array = bool(shape) or (
+        is_nxdata_class and isinstance(node, NXTreeField)
```
RubelMozumder, line 582: Would it convert the NXdata/title field to an HDF5Reference array?

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: The issue is that adding 4 more quantities per quantity just blows up the schema, which makes it slow to start the app. We discussed this extensively already, we cannot just keep adding more quantities if we want the startup time to be reasonably fast.

lukaspie: This is not about data size, but schema size

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
```diff
@@ -558,8 +575,31 @@ def _build_quantity_from_node(
...
+        scalar_items,
+        field_name=node.name or "",
+        is_hdf5_reference=is_hdf5_reference,
+    )
+
+    has_statistics = (
+        is_nxdata_class
+        and isinstance(node, NXTreeField)
+        and is_effectively_array
+        and python_type in ("np.float64", "np.int64", "np.complex128")
```
RubelMozumder, line 602: has_statistics accepts np.complex128, but the parser computes stats only for dtype kind iuf, so complex NXdata fields get four permanently empty quantities.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: I do not understand this comment, these are the field statistics.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: Float values can be arrays, so we are using HDF5Reference all the time for these.

src/pynxtools/nomad/parsers/parser_v2.py
```diff
@@ -800,6 +887,78 @@ def _populate_field(
...
+                    # Each element of the array is independently enumerated
+                    # (e.g. NXsample's sample_component: one label per entry).
+                    bad_values = [
+                        v for v in value.flat if v not in field_ann.enumeration
+                    ]
+                else:
+                    bad_values = [] if value in field_ann.enumeration else [value]
+            except Exception as e:
+                self._logger.debug("Error checking enum for %s: %s", hdf_field_name, e)
+                return
```
RubelMozumder, line 952: The enumeration check reads the full dataset via hdf_node[()], which is exactly the cost HDF5Reference avoids. What about hdf_node.size?

src/pynxtools/nomad/converters/templates/nexus.py.j2
```diff
@@ -213,22 +222,60 @@ class {{ class_name }}({{ base_class }}{% if base_is_generated %}{% for mod, cls
         {% endif %}
         {% endif %}
     )
+    {% if qty.has_statistics %}
+    {{ qty.python_name }}__min = Quantity(
+        type={{ qty.python_type }},
+        {% if qty.node.variadic %}
+        variable=True,
+        {% endif %}
+        description="Minimum of {{ qty.python_name }}, computed over the full array at parse time.",
```
RubelMozumder, line 231: Stat names share the {field}__{suffix} namespace with field-attribute quantities, and NXparameters already declares attributes named `min` and `max` that will raise a conflict.

src/pynxtools/nomad/converters/templates/nexus.py.j2
```diff
@@ -33,7 +33,8 @@ import numpy as np
 
 {% for mod, cls in nomad_extra_bases %}{% if mod != "nomad.datamodel.metainfo.basesections" %}from {{ mod }} import {{ cls }}
 {% endif %}{% endfor %}{% if needs_basesections %}from nomad.datamodel.metainfo import basesections
-{% endif %}from nomad.datamodel.metainfo.annotations import ELNAnnotation, ELNComponentEnum, SchemaAnnotation
+{% endif %}from nomad.datamodel.hdf5 import HDF5Reference
```
RubelMozumder, line 36: HDF5Reference is imported unconditionally, so many generated modules (e.g. applications/afm.py, base_classes/amplifier.py) never use it.

tests/nomad/test_parsing_v2.py
```diff
@@ -147,18 +147,17 @@ def test_arpes_example(arpes_archive):
...
     assert data.energies is not None
-    assert data.energies.check("eV")
+    assert data.energies.endswith("#/entry/data/energies")
     # manual name resolution
     assert data.AXISNAME["angles"] is not None
     # TODO: reimplement with field statistics
     # assert data.AXISNAME__max["angles__max"].value == 2.168025463513032
-    assert (1 * data.AXISNAME["angles"].unit).check("1/Å")
-    assert (1 * data.AXISNAME["delays"].unit).check("fs")
+    assert data.AXISNAME["angles"].value.endswith("#/entry/data/angles")
```
RubelMozumder, line 159: The commented-out assertion can be enabled now: I get AXISNAME__max["angles__max"] == 2.168025463513032. Without it, no test covers the variadic MQuantity-wrapped stats path.

tests/nomad/test_parsing_opt_v2.py
```diff
@@ -14,8 +14,8 @@
 check, which required a literal "eventID"/"imageID" prefix — fixed to use the
 canonical ``get_nx_namefit()`` matcher instead.
 
-The FIELD_STATISTICS value assertions (mean/min/max/size/ndim) are commented out —
-see the TODO block below for why.
+Asserts the FIELD_STATISTICS parallel quantities (`{name}__min/__max/__size/__ndim`) that
```
RubelMozumder: The comment says real stays unset and keeps its 3D shape, but it is now an HDF5Reference with shape=[] and is populated. Therefore, it can be asserted in a test with an array.

docs/learn/pynxtools/nexus-metainfo-generation.md
```diff
@@ -547,6 +563,37 @@ exactly the `NX_ANY` case.
...
 ---
 
+## Field statistics
+
+Numeric array quantities inside an `NXdata`-derived class get four parallel
+scalar quantities generated automatically, so the array's summary statistics are searchable even though the array itself is not:
+
+```python
+# NXDL: <field name="energy" type="NX_FLOAT" units="NX_ENERGY"> inside an NXdata group
+energy = Quantity(type=np.float64, ...)
```
RubelMozumder, line 573: Shouldn't be type=HDF5Reference.

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
```diff
@@ -103,6 +103,17 @@ class QuantityContext:
     eln_component: str | None
     # Default value for the ELN annotation; set for single-value MEnum only.
     eln_default: str | None
+    # True for numeric array fields inside NXdata-derived classes — the generator
+    # emits parallel {name}__mean/__min/__max/__size/__ndim scalar quantities.
```
RubelMozumder, line 107: No __mean quantity is generated. The same issue in the PR description.

RubelMozumder (changes requested):
The __min/__max carry no unit, e.g., angles__max is 2.168 without `deg`. With this final comment I am done.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: This was indeed a bug, it is now fixed. strs/bools/enums are regular quantities, not using `HDF5Reference`.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: see above

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: I remove the reference of the file.

RubelMozumder: > You mean "remove the mention of `pynxtools/nomad/parsers/parser_v2.py`"? That can of course be done.
> 
> In general, the `nexus-metainfo-generation` docs would benefit from a rewrite, currently it is a WIP version where a bunch of information is dropped, but this should be more user-facing, I agree.

Yeah, removing the code reference from the docs.

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: The thing is, if there is `shape=["*", 3]`, this now gets mapped to an `HDF5Reference`, so we cannot keep this.

src/pynxtools/nomad/metainfo/applications/apm.py
lukaspie: Sorry, I make a mistake here earlier: the reason this is a `HDF5Reference` is that this is a field inside `NXdata` , where we expect float fields to be arrays (axes or signals). Outside of NXdata, this would be a regular Quantity with type=np.float64`.

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
lukaspie: It converts those NXdata fields to an HDF5Reference that are floats/integers and/or have dimensions. So no, title would not be an HDF5Reference.

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
lukaspie: You are right. This is fixed now by using `_STATISTICS_TYPES = ("np.float64", "np.int64")` and only building the field statistics for these.

src/pynxtools/nomad/parsers/parser_v2.py
lukaspie: One wants to check the actual value right? So there's no way around opening the array. Given that this only comes up for enumerated fields where the actual value is small  (something like https://fairmat-nfdi.github.io/nexus_definitions/classes/applications/NXxps.html#nxxps-entry-xps-coordinate-system-x-field), I think it's okay to open the dataset here.

`hdf_node.size` does not tell you anything about the actual value.

src/pynxtools/nomad/converters/templates/nexus.py.j2
lukaspie: `PARAMETER__min` is indeed quite annoying. I don't think it would really raise any issue, it is more that this value gets overwritten by the field statistics and of course it carries a different semantic meaning. Need to think about it.

src/pynxtools/nomad/converters/templates/nexus.py.j2
lukaspie: This is fixed now, it is only imported if needed

tests/nomad/test_parsing_v2.py
lukaspie: yup, enabled now

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: Yes, because of the dimensionality. I removed that, then Quantity is fine again.

docs/learn/pynxtools/nexus-metainfo-generation.md
RubelMozumder: Yeah, right. In this case, the quantity type should be `HDF5Reference`, not np.float64

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
lukaspie: This is cleaned up now, indeed no `__mean` exists

docs/learn/pynxtools/nexus-metainfo-generation.md
RubelMozumder: NOMAD also allows customised plots with `PlotSection`. If only way to represent the plot using HDF5Reference that's all good.

src/pynxtools/nomad/metainfo/applications/apm.py
RubelMozumder: Sorry for making the comment short and ambiguous. Ahh, there are multiple Elapsed_time fields, both scalars and arrays (fields in NXdata)

src/pynxtools/nomad/converters/nxdl_to_metainfo.py
RubelMozumder: Yeah, it is fixed. A title in NXapm is now of type str, which was `HDF5Reference` in the previous version. The issue was coming from other place, which is fixed now.

RubelMozumder (approved):
LGTM!

docs/learn/pynxtools/nexus-metainfo-generation.md
lukaspie: If you want to use a custrom plot (not from an existing NeXus file), you could for example use `H5WebAnnotation`

lukaspie:
> The __min/__max carry no unit, e.g., angles__max is 2.168 without `deg`. With this final comment I am done.

They contain units now as well :+1

src/pynxtools/nomad/converters/templates/nexus.py.j2
lukaspie: Let's table this for now. Eventually, these field statistics should just come from the quantity itself, so then it will not be an issue anymore.

PR #839, opened 2026-09-09
bump definitions, prepare pynxtools-xas

lukaspie:
- bump definitions
- add `reference` to `UNALLOWED_GROUP_NAME` for the NOMAD schema. `NXxas_trans` and others declare `<group type="NXsubentry" name="reference">` inside `NXentry`. In the old schema, `NXentry` maps to the `NexusActivityStep` base section (via `BASESECTIONS_MAP`), which already defines a `reference` *Quantity*. So, to avoid a NOMAD Metainfo conflict, the group becomes a `SubSection` named `reference_group` instead.
- Small addition to the `json_map` HDF5 mapper to allow HDF5 attributes to be mapped as well. Needed for parsing in pynxtools-xas.

Note that this does not change the generated Metainfo Python schema, as we want to have separate commits to cherry-pick for a patch release. I will make a follow-up PR to update the Python schema as well.


src/pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -248,6 +249,32 @@ def _handle_hdf5_file(self, file_path: str) -> dict[str, Any]:
...
+        try:
+            with h5py.File(file_path, "r") as h5f:
+                _insert_attrs_from_h5(h5f, self.data)
+        except Exception:
+            logger.warning(
+                "Could not read HDF5 attributes from %s; @data:...@attr "
+                "config tokens for this file will not resolve.",
+                file_path,
+            )
+
```
rettigl, line 281: I have added something very similar here: https://github.com/FAIRmat-NFDI/pynxtools/pull/807

src/pynxtools/dataconverter/readers/json_map/reader.py
```diff
@@ -248,6 +249,32 @@ def _handle_hdf5_file(self, file_path: str) -> dict[str, Any]:
             file_path, lazy=False, unpacker=unpack_hdf_dataset_for_json_map
         )
         merge(self.data, hdf)
+
+        # hdfdict.load() only round-trips pynxtools' own TYPE marker, not
```
mkuehbach: Round-tripping the file twice is more costly than doing its once, the PR is okay for building a demo.

mkuehbach (approved):

src/pynxtools/dataconverter/readers/json_map/reader.py
lukaspie: Indeed, they are quite similar, thanks for the hint.

After comparing them I tried to build a single combined version for the traversal:
- Walk the attributes in a module-level `_insert_attrs_from_h5` helper fucntion, matching the other module-level helpers in the file. This is similar to what you had in #8007.
- The function now only creates dicts for *groups* and leaves dataset values loaded by hdfdict untouched. So it no longer initializes an entry for every item. Since hdfdict has already populated everything for datasets, we don't need to init empty dicts.
- Added capture of root-level file attributes (these are attributes on the `h5py.File` object itself), which neither original version handled.
- Dropped the `try`/`except` block because by the time the attribute walk runs, `hdfdict.load()` has already opened and read the same file successfully, so a second open can't fail in a new way.

Can you check the current version and if it would make sense for #807 as well?

src/pynxtools/dataconverter/readers/json_map/reader.py
lukaspie: Alternative would be to do it in `hdfdict.load` which I think is a bit problematic due to licensing (you would need to use `decode_if_bytes` in there, mixing Apache import into MIT code. But I can have a look if that can be separated. In any case, we are only iterating the file metadat (`.attrs`) and not the actual values, so this should not meaningfully slow down the reading.

src/pynxtools/dataconverter/readers/json_map/reader.py
lukaspie: I mean merging this now here because I need a quick patch release. I will rebase #807 and then @rettigl and I can see if that works as well.

PR #840, opened 2026-09-10
regenerate metainfo python schema after definitions update

lukaspie:
Follow-up to #839, updates the Python metainfo files. Kept separate to allow patch release with cherry-pick.


src/pynxtools/nomad/metainfo/applications/xas_herfd.py
```diff
@@ -0,0 +1,3060 @@
...
+        ),
+        a_eln=ELNAnnotation(
+            component=ELNComponentEnum.NumberEditQuantity,
+        ),
+        a_display={"unit": "eV"},
+    )
+    is_experimental = Quantity(
+        type=bool,
+        links=[
+            "https://fairmat-nfdi.github.io/nexus_definitions/classes/contributed_definitions/NXxas.html#nxxas-entry-is-experimental-field"
```
RubelMozumder, line 200: This is not part of the xas_herfd but xas, though it is not wrong, but I am wondering if it is intentional.

RubelMozumder (approved):
LGTM!

src/pynxtools/nomad/metainfo/applications/xas_herfd.py
lukaspie: Thanks for checking, this was indeed incorrect. I found a small bug in the converter that carried these inherited quantities to the "lowest-level" application definition. This is now fixed with https://github.com/FAIRmat-NFDI/pynxtools/commit/0ad868ed48e2cbf9eeb0ca97d1f6e8c502eeb543 and the classes are updated.

PR #842, opened 2026-09-15
metainfo schema: provenance tracking for hand-written additions

lukaspie:
Regenerating the NeXus metainfo currently overwrites hand-written content (e.g. new quantities or a custom `normalize()` method) and there is no way to tell it apart from generated code. This adds provenance tracking: a sidecar file (`generated_manifest.json`) records which members/imports are coming from the generator (together with hashes for them), so regeneration can refresh generated content while preserving hand edits.

It also adds a `--fix` flag for fixing the generated Python files. This is relevant for fixing bugs in the generator code or if there were some NXDL changes on the NIAC side we need to integrate.

The provenance sidecar file was seeded once, explicitly, via `bootstrap_manifest()`, and from now on, it gets auto-updated.

In the CLI, I added the `--fix` and ensured that the flags that are mutually exclusive are tagged as such.

Also added a lightweight test for the provenance tracker.


RubelMozumder (approved):
LGTM!
Over the updates of the schema it might need a bit manual attention. But, I am wondering how it handles the following questions.

Regarding a case where a quantity can be refered by multiple names, and if someone renames a quantity to different canonical name, regenegation of the schema will fetch the same quantity from NXDL. The same situation, if docs is written or extended, duplication of the quantity also appear.

Another, open question, how would it transform back to the NXDL with such a multi-occurring situation?

lukaspie:
> Regarding a case where a quantity can be refered by multiple names, and if someone renames a quantity to different canonical name, regenegation of the schema will fetch the same quantity from NXDL. The same situation, if docs is written or extended, duplication of the quantity also appear.
>

Yes, this will be a situation where we cannot avoid manual changes. But we can try to limit them, that's what I am trying here. But anyway, renaming quantities can only happen for non--approved definitions (so everything in contributed).
 
> Another, open question, how would it transform back to the NXDL with such a multi-occurring situation?

This is an even more complicated question. We have to explore this a bit more in depth when we build the backwards converter. But we will anyway only do this sparringly: _if_ we want to bring it back to NIAC, but not for every single class we develop. I hope that this provenance feature will help us there too though. We can distinguish between _changed_ and _addded_ concepts.

